PR #21, opened 2015-11-12
Fix typo in README (exits->exists)

mika:
Simple typo, but relevant since it's a keyword for the configuration file :)


aelsabbahy:
Thanks, I really should look for a way to just autogenerate the doc.

PR #47, opened 2016-03-11
Add TAP output support

mika:
Implementing the TAP specification (https://testanything.org/tap-specification.html) this allows usage of goss e.g. with Jenkins' TAP Plugin.

Output example:

````
% release/goss-linux-amd64 v --format tap --no-color
1..4
ok 1 - Package: zsh: installed: matches expectation: [true]
not ok 2 - Package: zsh: version: doesn't match, expect: [["5.0.7-6"]] found: [["5.0.7-5"]]
ok 3 - Package: vim: installed: matches expectation: [true]
ok 4 - Package: vim: version: matches expectation: [["2:7.4.488-7"]]
````

Signed-off-by: Michael Prokop <mprokop@sipwise.com>

### DIFF
--- a/outputs/tap.go
+++ b/outputs/tap.go
@@ -1,0 +1,47 @@
+package outputs
+
+import (
+	"fmt"
+	"strconv"
+	"time"
+
+	"github.com/aelsabbahy/goss/resource"
+)
+
+type Tap struct{}
+
+func (r Tap) Output(results <-chan []resource.TestResult, startTime time.Time) (exitCode int) {
+	testCount := 0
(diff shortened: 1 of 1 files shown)


aelsabbahy:
Thanks for the contribution, can you remove `junit.go` from this commit.

Should the return code be 1 or 2 on failure? I only used 2 in Nagios since it has special meaning there.

mika:
Oh sorry, the junit.go inclusion was by mistake, fixed.

Good point about the return code, I just made it return 1 on failure.

PR #48, opened 2016-03-11
Add junit output support

mika:
Implementing the junit specification (http://help.catchsoftware.com/display/ET/JUnit+Format) this allows usage of goss e.g. with Jenkins' JUnit Plugin.

Output example:

````
<testsuite name="goss" errors="0" tests="4" failures="1" time="0.071" timestamp="2016-03-11T14:17:57Z">
<testcase name="Package vim installed" time="0.056">
<system-out>Package: vim: installed: matches expectation: [true]</system-out>
</testcase>
<testcase name="Package vim version" time="0.000">
<system-out>Package: vim: version: matches expectation: [["2:7.4.488-7"]]</system-out>
</testcase>
<testcase name="Package zsh installed" time="0.061">
<system-out>Package: zsh: installed: matches expectation: [true]</system-out>
</testcase>
<testcase name="Package zsh version" time="0.000">
<system-err>Package: zsh: version: doesn't match, expect: [["5.0.7-6"]] found: [["5.0.7-5"]]</system-err>
<failure>Package: zsh: version: doesn't match, expect: [["5.0.7-6"]] found: [["5.0.7-5"]]</failure>
</testcase>
</testsuite>
````
Signed-off-by: Michael Prokop <mprokop@sipwise.com>

### DIFF
--- a/outputs/junit.go
+++ b/outputs/junit.go
@@ -1,0 +1,75 @@
+package outputs
+
+import (
+	"fmt"
+	"strconv"
+	"time"
+
+	"github.com/aelsabbahy/goss/resource"
+)
+
+type JUnit struct{}
+
+func (r JUnit) Output(results <-chan []resource.TestResult, startTime time.Time) (exitCode int) {
+	testCount := 0
+	failed := 0
+
+	// ISO8601 timeformat
+	location, _ := time.LoadLocation("Etc/UTC")
+	timestamp := time.Now().In(location).Format(time.RFC3339)
+
+	var summary map[int]string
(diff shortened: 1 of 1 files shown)


aelsabbahy:
Should the return code be 1 or 2 on failure? I only used 2 in Nagios since it has special meaning there.

mika:
Ah sure, makes sense - just adjusted it to fail with return code 1 (and rebased the PR).

aelsabbahy:
Thanks for the contribution!

PR #35, opened 2016-01-19
add support for Aline Linux

ryancox:

PR #60, opened 2016-03-28
Support Pacman package manager

### LINKED ISSUE #57: Support querying Pacman package manager
The package resource could easily add support for the `pacman` package manager used in Arch Linux. If we add support, I'll switch distributive's package check to use goss's code. Would you merge a PR to this effect?

siddharthist:
Fixes #57 

Tests added and passing. Thanks for reviewing!

### DIFF
--- a/integration-tests/test.sh
+++ b/integration-tests/test.sh
@@ -14,7 +14,11 @@ for arch in amd64 386;do
     if docker ps -a | grep goss_int_test_$os;then
       docker rm -vf goss_int_test_$os
     fi
-    id=$(docker run -v "$PWD/goss:/tmp/goss"  -d --name "goss_int_test_$os" "aelsabbahy/goss_$os" /sbin/init)
+    if [[ $os == "arch" ]]; then
+      id=$(docker run -v /run:/run:ro -v /sys/fs/cgroup:/sys/fs/cgroup:ro -v "$PWD/goss:/tmp/goss"  -d --name "goss_int_test_$os" "aelsabbahy/goss_$os" /sbin/init)
+    else
+      id=$(docker run -v "$PWD/goss:/tmp/goss"  -d --name "goss_int_test_$os" "aelsabbahy/goss_$os" /sbin/init)
+    fi
     ip=$(docker inspect --format '{{ .NetworkSettings.IPAddress }}' "$id")
     trap "rv=\$?; docker rm -vf $id; exit \$rv" INT TERM EXIT
     # Give httpd time to start up
@@ -25,7 +29,7 @@ for arch in amd64 386;do
   out=$(docker exec goss_int_test_$os bash -c "time /tmp/goss/$os/goss-linux-$arch -g /tmp/goss/$os/goss.json validate")
   echo "$out"
 
-  grep -q 'Count: 39, Failed: 0' <<<"$out"
+  egrep -q 'Count: [0-9]{2}, Failed: 0' <<<"$out"
 
   docker exec goss_int_test_$os bash -c "bash -x /tmp/goss/generate_goss.sh $os $arch"
 
--- a/integration-tests/test.sh
+++ b/integration-tests/test.sh
@@ -31,15 +31,17 @@ for arch in amd64 386;do
 
   egrep -q 'Count: [0-9]{2}, Failed: 0' <<<"$out"
 
-  docker exec goss_int_test_$os bash -c "bash -x /tmp/goss/generate_goss.sh $os $arch"
+  if [[ ! $os == "arch" ]]; then
+    docker exec goss_int_test_$os bash -c "bash -x /tmp/goss/generate_goss.sh $os $arch"
 
-  docker exec goss_int_test_$os bash -c "diff -wu /tmp/goss/${os}/goss-expected.json /tmp/goss/${os}/goss-generated.json"
+    docker exec goss_int_test_$os bash -c "diff -wu /tmp/goss/${os}/goss-expected.json /tmp/goss/${os}/goss-generated.json"
 
-  docker exec goss_int_test_$os bash -c "diff -wu /tmp/goss/${os}/goss-aa-expected.json /tmp/goss/${os}/goss-aa-generated.json"
+    docker exec goss_int_test_$os bash -c "diff -wu /tmp/goss/${os}/goss-aa-expected.json /tmp/goss/${os}/goss-aa-generated.json"
 
-  docker exec goss_int_test_$os bash -c "bash -x /tmp/goss/generate_goss.sh $os $arch -q"
+    docker exec goss_int_test_$os bash -c "bash -x /tmp/goss/generate_goss.sh $os $arch -q"
 
-  docker exec goss_int_test_$os bash -c "diff -wu /tmp/goss/${os}/goss-expected-q.json /tmp/goss/${os}/goss-generated.json"
+    docker exec goss_int_test_$os bash -c "diff -wu /tmp/goss/${os}/goss-expected-q.json /tmp/goss/${os}/goss-generated.json"
+  fi
 
   #docker rm -vf goss_int_test_$os
 done
--- a/system/package_pacman.go
+++ b/system/package_pacman.go
@@ -1,0 +1,55 @@
+package system
+
+import (
+	"errors"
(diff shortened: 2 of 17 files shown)


aelsabbahy:
Tests aren't running for arch, add them integration-test task: https://github.com/aelsabbahy/goss/pull/60/files#diff-b67911656ef5d18c4ae36cb6741b7965R43

siddharthist:
I'm not convinced that the error is related. If you search "exit 1" in the full log, the error is one of service detection.

aelsabbahy:
The travis results are hard to read because they're run in parallel, just run `make arch` locally and you'll see the errors.

It might be that `service` is not handling arch correctly, I haven't had time to dig into it to understand why.

siddharthist:
@aelsabbahy It looks to me like that's the case: http://pastebin.com/raw/nnx4fh8f

aelsabbahy:
It's part of goss-shared.json. `service` is returning running: true in arch for a non existent service.

siddharthist:
@aelsabbahy What would you like to do about that? It doesn't seem like a problem related to this PR.

aelsabbahy:
Don't think we should have half-baked arch-linux support. It would be very easy for someone to assume that Arch is supported when it's not fully functional.

Feel free to create another PR for service or bundle the two up.

Take a look at #35 for reference, this was a PR to add alpine Linux support from another contributor.

siddharthist:
@aelsabbahy This doesn't strike me as an Arch Linux specific problem, Arch uses systemd for service management just like many other distributions.

aelsabbahy:
If you have an Archlinux virtual or physical machine, I would test it there. If it works, then it might be the Dockerfile or docker run calls that need to be fixed to support systemd properly.

siddharthist:
@aelsabbahy Good thinking! Looks like it works properly on my machine, I'll try to get a better Dockerfile.
```
$ mv integration-tests/goss/goss-shared.json goss.yaml
$ goss validate
...................

Total Duration: 0.502s
Count: 19, Failed: 0

```

siddharthist:
I think the problem is likely that [goss ignores errors when querying services](https://github.com/aelsabbahy/goss/blob/d28f3cc6d708fb012ea614acf712eb56712a7de3/resource/service.go#L35-L36). If this were not the case, we'd likely see an error here, instead of a false positive.

siddharthist:
It turns out that `dbus` is not running in the Arch docker container, which makes a lot of sense (there's no need for services in a single-process container). I can't find an easy way to get it running, either. I think we should just remove the service check from the shared goss tests, what do you think @aelsabbahy?

aelsabbahy:
#64 Fixes the silent error when system.go is misconfigured to init for a system that wasn't using init, which is the case here. Rebasing on master will show you a clearer error message.

system.go needs to be changed to mark archlinux as NewServiceDbus rather than using the default NewServiceInit.

I would look into doing something like: https://github.com/maci0/docker-systemd-unpriv for running systemd inside docker.

siddharthist:
@aelsabbahy I fixed the error message problem more broadly in 2f668ad and added arch with NewServiceDbus in ce31e45. I'll take a look at that example and let you know if I can come up with anything!

siddharthist:
@aelsabbahy Would you be willing to help get DBus running in Docker? I'm not sure I know how.

aelsabbahy:
I don't mind looking into this, but my schedule has been pretty busy, so it might take me a while to get around to it.

FYI, 2f668ad addresses the goss `add` and `autoadd` case, I don't believe that's used when doing a `validate`

siddharthist:
@aelsabbahy Thanks for all your feedback, it's much appreciated! I got dbus running in docker, but it requires some mounts on the host (and the host has to be running systemd). This conforms to the example you provided and runs just fine locally, but doesn't seem to like to run in Travis. What would you like to do? 

One option would be to skip `goss-shared.json` or just the service `foobar` check when `$os == arch` and `$CI == true`.

aelsabbahy:
Wasn't able to get a successful run locally, can you change:
`make -j10`
to:
`make`

In the travis.yml file https://github.com/aelsabbahy/goss/blob/master/.travis.yml#L25

This will make it easier to see why it's failing in travis.

siddharthist:
@aelsabbahy Are you testing locally on a distro running systemd?

siddharthist:
Here's the error: 
```
Failures:

foobar: enabled: Error: No such interface '\''org.freedesktop.DBus.Properties'\'' on object at path /org/freedesktop/systemd1/unit/foobar_2eservice
foobar: running: Error: No such interface '\''org.freedesktop.DBus.Properties'\'' on object at path /org/freedesktop/systemd1/unit/foobar_2eservice
```

aelsabbahy:
I was able to get it working with centos7 using the instructions from: https://hub.docker.com/_/centos/

I think that rules out travis, I wonder if there's a way to get systemd in Arch to behave the same way as it does in redhat/centOS.

I would try to see if you can get it working in a similar fashion as centos7 one more time.. If you can't I guess we can move the service test out of shared, and into a 2nd shared file that's imported by all the other integration tests.

Here's centos7 succeeding and arch failing for reference:
* https://travis-ci.org/aelsabbahy/goss/builds/120183315
* 8a3bf5f

siddharthist:
@aelsabbahy How would I remove it from goss-shared.json and still have it included in testing? Sorry, I'm trying to read through the testing scripts, but I'm having a hard time understanding them.

aelsabbahy:
I would create a 2nd shared json file with the service entry in it and import that in all the relevant goss.json test fixtures. Example: https://github.com/aelsabbahy/goss/blob/master/integration-tests/goss/centos6/goss.json#L66

siddharthist:
@aelsabbahy Oh, I see. Thank you!

siddharthist:
@aelsabbahy Finally got passing tests :smile:

system/package_pacman.go
```diff
@@ -0,0 +1,55 @@
...
+	p.loaded = true
+	// TODO: extract versions
+	cmd := util.NewCommand("pacman", "-Q", "--color", "never", "--noconfirm", p.name)
+	if err := cmd.Run(); err != nil {
+		return
+	}
+	p.installed = true
+	// the output format is "pkgname version\n", so if we split the string on
+	// whitespace, the version is the second item.
+	p.versions = []string{strings.Fields(cmd.Stdout.String())[1]}
```
aelsabbahy: Does pacman support multiple versions of the same package being installed? (ex: multiple kernels)

This code looks like it assumes only 1 package is installed.

siddharthist: I'm not aware of any support for multiple package versions.

integration-tests/test.sh
```diff
@@ -25,17 +29,19 @@ for arch in amd64 386;do
   out=$(docker exec goss_int_test_$os bash -c "time /tmp/goss/$os/goss-linux-$arch -g /tmp/goss/$os/goss.json validate")
   echo "$out"
 
-  grep -q 'Count: 39, Failed: 0' <<<"$out"
+  egrep -q 'Count: [0-9]{2}, Failed: 0' <<<"$out"
```
aelsabbahy: The original code is brittle, but until unit tests are in place for parsing JSON/YAML, it's the only validation there is that all the tests ran and none were skipped. Can you undo this change.

siddharthist: What do you want me to do? Just add a conditional for if `$os == arch`? It's a different number of tests.

aelsabbahy: oh crap, forgot arch is unique.. screw it.. keep it as is, I'll fix later.

aelsabbahy:
This looks good enough to merge, can you squash the PR and I'll merge it when tests pass again.

siddharthist:
@aelsabbahy Github has a new squash-on-merge button you can use :smile:

aelsabbahy:
Today I learned :)

Thanks for the contribution and thanks for the squash-on-merge tip!

aelsabbahy:
FYI.. not sure if random travis-ci hiccup or related to squash-on-merge button, but a build didn't kick off for master *shrug*.

siddharthist:
@aelsabbahy Thanks for merging :fireworks:

PR #61, opened 2016-03-28
Add `sudo` to Makefile for testing with Docker

### LINKED ISSUE #58: Difficult Makefile
The Makefile for goss depends on `goupx`, which probably introduces unecessary complexity. A few mb here or there isn't a big deal to anyone who can't run it themselves.

Additionally, `make` depends on Docker access. Most developers will run `make` as an unpriveleged user; the only secure access to `docker` can be through `sudo`. However, if you run `sudo make`, your GOPATH and GOROOT probably aren't set correctly to build `goss`.

siddharthist:
Fixes #58 

Looks like it doesn't break Travis!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,13 +43,13 @@ release: build
 test-int: centos6 wheezy precise alpine3
 
 centos6: build test
-	cd integration-tests/ && ./test.sh $@
+	cd integration-tests/ && sudo ./test.sh $@
 wheezy: build test
-	cd integration-tests/ && ./test.sh $@
+	cd integration-tests/ && sudo ./test.sh $@
 precise: build test
-	cd integration-tests/ && ./test.sh $@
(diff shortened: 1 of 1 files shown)


aelsabbahy:
Upon second thought.. I think I would prefer some sort of $DOCKER_CMD variable that can be set to `docker` or `sudo docker` with one of them being the default. I don't think other contributors would be comfortable running `sudo ./test.sh`, I know I wouldn't want to do so for a project I didn't know.

Sorry for not clarifying this stance earlier.

siddharthist:
The only way to use Docker securely is to use `sudo`. I think that's a good point, though, we should add it directly to the commands in `test.sh`. But it can't be optional; there's just no other way to do it. Having access to the `docker` command/socket is synonymous with root access.

PR #63, opened 2016-03-30
Add `sudo` to docker commands in test.sh

siddharthist:
Since the docker socket effectively gives root access, the only secure way to access it is using the `sudo` command. See discussion on #61

### DIFF
--- a/integration-tests/test.sh
+++ b/integration-tests/test.sh
@@ -6,36 +6,36 @@ os=$1
 
 for arch in amd64 386;do
   cp ../release/goss-linux-$arch goss/$os/
-  if ! docker images | grep aelsabbahy/goss_$os;then
-    docker build -t aelsabbahy/goss_$os - < Dockerfile_$os
+  if ! sudo docker images | grep aelsabbahy/goss_$os;then
+    sudo docker build -t aelsabbahy/goss_$os - < Dockerfile_$os
   fi
 
-  if ! docker ps | grep goss_int_test_$os;then
-    if docker ps -a | grep goss_int_test_$os;then
-      docker rm -vf goss_int_test_$os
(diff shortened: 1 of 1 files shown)


aelsabbahy:
Please have this set by a top level variable that the user can override easily by environment variable or top level config. Either that or make it smart enough to detect when this is needed.

I don't care to force a certain configuration on a developer workstation.

siddharthist:
@aelsabbahy Could you outline a configuration that this wouldn't work in? The only ones I can think of are 
 1. Developer is running whole Makefile as root
 2. Developer has sudo-less access to `docker` command

both of which are highly questionable.

aelsabbahy:
Sorry this slipped through the cracks, forgot to respond.

Basically, it'll fail if someone has the `alias` setup described here:
http://www.projectatomic.io/blog/2015/08/why-we-dont-let-non-root-users-run-docker-in-centos-fedora-or-rhel/

siddharthist:
@aelsabbahy Ah, I see. I'm not sure that's true, though:
```
$ ls /root 
ls: cannot open directory '/root': Permission denied
$ sudo sudo ls /root
$ sudo sudo sudo ls /root
$
```

aelsabbahy:
You're assuming the user has full sudo access, and therefor `sudo sudo` access.
If your sudoers looks like this, you'll get errors on the `sudo sudo`
```
user  ALL=(ALL)       /bin/docker
```

You'll get an error along the lines of:
```
Sorry, user <username> is not allowed to execute '/bin/sudo docker info' as root on <hostname>
```

siddharthist:
@aelsabbahy Gotcha, thanks for outlining that! I'll close this until I can think of a better way.

PR #69, opened 2016-04-07
Refactor distro/package/service detection

siddharthist:
Makes that whole process more consistent, and probably more robust.

### DIFF
--- a/system/system.go
+++ b/system/system.go
@@ -80,128 +80,118 @@ func New(c *cli.Context) *System {
 		NewProcess:  NewDefProcess,
 		NewGossfile: NewDefGossfile,
 	}
-	// FIXME: Detect-os needs to be refactored in a consistent way
-	// Also, cache should be its own object
 	sys.detectService()
+	sys.detectPackage(c)
+	return sys
+}
 
+// detectPackage adds the correct package creation function to a System struct
+func (sys *System) detectPackage(c *cli.Context) {
 	p := c.GlobalString("package")
+	if p != "deb" && p != "apk" && p != "pacman" && p != "rpm" {
+		p = DetectPackageManager()
+	}
 	switch p {
-	case "rpm":
-		sys.NewPackage = NewRpmPackage
 	case "deb":
 		sys.NewPackage = NewDebPackage
-	case "alpine":
+	case "apk":
 		sys.NewPackage = NewAlpinePackage
 	case "pacman":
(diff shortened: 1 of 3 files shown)


system/system.go
```diff
@@ -80,128 +80,119 @@ func New(c *cli.Context) *System {
...
-		if bytes.Contains(b, []byte("Ubuntu")) {
-			return true
+// DetectPackageManager attempts to detect whether or not the system is using
+// "deb", "rpm", "apk", or "pacman" package managers. If none are found, it
+// attempts to fall back to detecting the running distro. If that fails, it
+// returns the empty string.
+func DetectPackageManager() string {
+	for _, manager := range []string{"deb", "rpm", "apk", "pacman"} {
+		if HasCommand(manager) {
+			return manager
```
aelsabbahy: I would reverse this and have distro detection take priority. I know it's extremely rare, but I've actually had an Archlinux installation in the past that had both pacman and rpm installed on it. I was using rpm command to get metadata out of rpm files.

siddharthist: @aelsabbahy I'll reverse the order. I think we could have better performance gains by focusing elsewhere :) this seems so minor that it would be impossible to tell the difference.

siddharthist: And that's a good point about the package managers! I have the same situation on my workstation, I don't know why I didn't think of it.

siddharthist:
@aelsabbahy I just rebased, squashed, and made the tests concurrent :) should be ready to merge after your review!

system/system_test.go
```diff
@@ -0,0 +1,51 @@
...
+		}
+	}
+	if failed {
+		t.Errorf("Function %v returned %v, which is not one of %v", name, output, validOutputs)
+	}
+}
+
+func TestPackageManager(t *testing.T) {
+	t.Parallel()
+	testOutputs(DetectPackageManager, []string{"deb", "rpm", "apk", "pacman"}, t)
```
aelsabbahy: Haven't run this yet, but it seems to be testing whether goss can detect the developer workstation. What happens if I run this test on a gentoo or OSX box?

siddharthist: You're right, `""` should also be a valid option. I'll change it.

siddharthist:
@aelsabbahy comment addressed!

siddharthist:
Thanks for merging!

aelsabbahy:
Thanks for contributing :)

PR #103, opened 2016-06-28
Support systemd without needing dbus

### LINKED ISSUE #101: systemd without dbus
It should be possible to use the service resource on a systemd distro like debian without dbus.

aelsabbahy:
Fixes #101


aelsabbahy:
Would like to add tests for systemd prior to merging this.

PR #142, opened 2016-08-28
add failures to Nagios long output

pysysops:
Hi,

I thought I'd open up a PR to output the actual failures in the Nagios ($longserviceoutput$) this makes Goss's Nagios output more contextual showing what has failed improving alerts if used with Sensu, Nagios etc.

Any feedback or discussion welcome.

Thanks,

Tim.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,5 @@
 # Random stuff for my local testing/development that I don't want checked in
 tmp/
 /goss.yaml
+
+/.idea
--- a/README.md
+++ b/README.md
@@ -177,6 +177,7 @@ package:
 * TAP
 * JUnit
 * nagios - Nagios/Sensu compatible output /w exit code 2 for failures.
+* nagios_verbose - nagios output with verbose failure output.
 
 ## Community Contribuations
 * [goss-ansible](https://github.com/indusbox/goss-ansible) - Ansible module for Goss
--- a/docs/manual.md
+++ b/docs/manual.md
@@ -90,6 +90,7 @@ Valid formats:
   * TAP
   * JUnit
   * nagios - Nagios/Sensu compatible output /w exit code 2 for failures.
+  * nagios_verbose - nagios output with verbose failure output.
 * --no-color (disable color)
 
 ### Example:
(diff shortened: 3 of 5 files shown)


aelsabbahy:
It's been years since I've used nagios. What's the output limit on it? For example if I have 30 failed tests, that's 30+ lines, will Nagios handle that correctly? What happens if you go over? Same question for Sensu. Curious to see what this looks like if you have a screenshot handy. 

I'm also wondering if this should be a new output format or if we replace nagios in place like this PR does.

pysysops:
I too wondered about making it a new output or whether to add a "verbose" argument (which might be preferable). According to the Nagios Plugin API:

> Nagios will only read the first 4 KB of data that a plugin returns. This is done in order to prevent runaway plugins from dumping megs or gigs of data back to Nagios. This 4 KB output limit is fairly easy to change if you need. Simply edit the value of the MAX_PLUGIN_OUTPUT_LENGTH definition in the include/nagios.h.in file of the source code distribution and recompile Nagios. There's nothing else you need to change!

Sensu truncates "metric" type checks to 255 characters which goss checks wouldn't be as there's no performance data. 

I can (attempt to) add in a --verbose flag or argument to switch the extra output on or off if that's preferable?

pysysops:
A couple of screenshots:
screen shot 2016-08-29 at 00 52 45
screen shot 2016-08-29 at 00 54 51

aelsabbahy:
Initial thoughts is to create a `nagios_verbose` to keep it simple.. if more outputs overlap, then we can add formatter options.

That said, I will not reject a PR with format options feature added as it's a nice to have feature, just don't want to add extra work if it's not yet necessary. So I'll leave it to you.

Thanks again for your great contributions so far, much appreciated.

pysysops:
I've added the extra nagios_verbose output formatter for the sake of simplicity as, all the other formats are full / verbose output anyway. 

Output example:
```
+ docker exec goss_int_test_centos7_amd64 /goss/centos7/goss-linux-amd64 -g /goss/centos7/goss.json validate -f nagios_verbose
GOSS CRITICAL - Count: 51, Failed: 2, Skipped: 5, Duration: 2.610s
Fail 1 - User: apache: exists: doesn't match, expect: [true] found: [false]
Fail 2 - Group: apache: exists: doesn't match, expect: [true] found: [false]
```

aelsabbahy:
LGTM!

aelsabbahy:
Released in [v0.2.4](https://github.com/aelsabbahy/goss/releases/tag/v0.2.4)

PR #143, opened 2016-08-28
Dns reverse lookup

pysysops:
This PR adds the ability to check DNS reverse lookups which are required for various application servers like Oracle's Weblogic to function properly.

It's basically a clone of the existing DNS resource but takes the IP address to lookup and uses net.lookupAddr to resolve the PTR record(s) for that IP address.

The reverse DNS resource will also help identify old PTR records hanging around (which results in IP's resolving to incorrect hostnames).

I've also made appropriate additions to the tests and documentation.

Any feedback appreciated.

### DIFF
--- a/add.go
+++ b/add.go
@@ -64,6 +64,13 @@ func AddResource(fileName string, gossConfig GossConfig, resourceName, key strin
 			os.Exit(1)
 		}
 		resourcePrint(fileName, res)
+	case "ReverseDNS":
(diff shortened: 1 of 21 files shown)


pysysops:
FYI - 64.233.184.26 - is a Google IP address which will more than likely resolve to the same hostname globally and reliably.

pysysops:
Closed due to discussion re: implementing server and type on dns resource.

Issue #147, opened 2016-09-13
Intermittent command: Error: Command execution timed out


shalomb:
We intermittently see ``goss`` fail on validating a ``gossfile`` that has quite a few `commands` to run.

Here's an excerpt of an autogenerated gossfile that has the seemingly (??) simple ruby one-liners .. this particular file has ~127 similar commands.

```
--- 
command:

  ? |-
    rvm ruby-1.8.7-p358 do ruby -e '...'
  :
    exit-status: 0
  
  ? |-
    rvm ruby-1.8.7-p358 do ruby -e '...'
  :
   exit-status: 0

  ? |-
    rvm ruby-1.9.2-p320 do ruby -e '...'
  :
    exit-status: 0

  ? |-
    rvm ruby-1.9.2-p320 do ruby -e '...'
  :
    exit-status: 0
  
  ? |-
    rvm ruby-2.1.2 do ruby -e '...'
  :
    exit-status: 0
```

(output (from ansible) trimmed for brevity and to hide sensitive code)
```
$ goss -v
goss version v0.1.5


"bash -c ' gossfile=app.yaml; goss --gossfile \"$gossfile\" validate </dev/null | tee \"/tmp/goss/$gossfile.out\"; 
"............................................................................................................................FFF
Failures:
...
rvm ruby-1.8.7-p358 do ruby -e '....' exit-status: Error: Command execution timed out (10s)
rvm ruby-1.9.2-p320 do ruby -e ' .... ': exit-status: Error: Command execution timed out (10s)
rvm ruby-2.1.2 do ruby -e ' ...': exit-status: Error: Command execution timed out (10s)\n\n
Total Duration: 10.111s\nCount: 127, Failed: 3",
...
```

We'd put the failures down to environmental one-offs (load spikes, etc) but there does seem to be a strong correlation to this happening the more commands the gossfiles have (possibly a GOPROCMAX or similar limitation or a `goss` scheduling limitation). 

1. Examining the load of the VMs around the time of the failures does not reveal anything untowards - but we can't rule that out (see 3).
2. Increasing the timeout (tried upto 60 seconds) does not seem to have an effect - and infact, in the failure path, we have to wait longer for `goss` to fail.
3. `goss` is invoked through `ansible` and played out across many VMs - a failure on one VM almost always coincides with failures on all the other VMs. It would suggest a Hypervisor (Xen here) vCPU scheduling issue and if it indeed is, it's strange that we only see `goss` sensitive to this.
[...]

PR #148, opened 2016-09-13
Allow user to override max-concurrent

aelsabbahy:
Possible fix for #147


cmd/goss/goss.go
```diff
@@ -61,6 +61,12 @@ func main() {
 					Value:  0,
 					EnvVar: "GOSS_RETRY_TIMEOUT",
 				},
+				cli.IntFlag{
```
aelsabbahy: Should this be a global option?

Also, this needs to be documented.

PR #168, opened 2016-11-08
Added NoFollowRedirects option, closes #160

### LINKED ISSUE #160: HTTP Redirects
I am unable to use `http` to check for a `301` http code. It looks like the test is following the redirect. Perhaps an option to not follow redirects?

tdi:
I added reversed boolean option `no-follow-redirects` in order not to break compatibility with the previous versions. If you think that it is better to have `follow-redirects` then I can reverse and rebase. 

Another idea is to make this flag and json elements optional but from the overall code convention I can see that all are required. 

Usage: (added `-r` to say "do not follow redirects)

```
➜  resource git:(master) goss add http http://wp.pl/ -r
Adding HTTP to './goss.yaml':

http://wp.pl/:
  status: 301
  allow-insecure: false
  no-follow-redirects: true
  timeout: 5000
  body: []
```

Please comment and review.

### DIFF
--- a/add.go
+++ b/add.go
@@ -16,9 +16,10 @@ import (
 func AddResources(fileName, resourceName string, keys []string, c *cli.Context) error {
 	setStoreFormatFromFileName(fileName)
 	config := util.Config{
-		IgnoreList:    c.GlobalStringSlice("exclude-attr"),
-		Timeout:       int(c.Duration("timeout") / time.Millisecond),
-		AllowInsecure: c.Bool("insecure"),
+		IgnoreList:        c.GlobalStringSlice("exclude-attr"),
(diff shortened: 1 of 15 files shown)


tdi:
I guess I would have to update the tests.

aelsabbahy:
Thanks for submitting this. It looks good to me, but can you update the tests to include one redirect test? Maybe a `google.com` test ensuring it's a 301.

tdi:
@aelsabbahy Ok no problem. I will get back when I do.

tdi:
@aelsabbahy ok I added tests with redirects.

aelsabbahy:
Sorry for the delay in merging, thanks for this!

PR #170, opened 2016-11-09
Add ability to specify a server on DNS resources

pysysops:
Originally, the DNS resource used the golang net library to perform a
net.lookupHost on DNS resourcces. net.lookupHost uses the local DNS
resolver which includes a host file lookup (still useful) as well as DNS lookup.

The following behaviour has been implemented:

Without the server attribute set, net.lookupHost is used which is great
for testing out host entries or just general hostname resolution.

With a server attribute set the github.com/miekg/dns library is used to
query the server (on port 53).

### DIFF
--- a/system/dns.go
+++ b/system/dns.go
@@ -67,7 +67,7 @@ func (d *DefDNS) setup() error {
 	}
 	d.loaded = true
 
-	//for i := 0; i < 3; i++ {
+	for i := 0; i < 3; i++ {
 		addrs, err := DNSlookup(d.host, d.server, d.qtype, d.Timeout)
 		if err != nil || len(addrs) == 0 {
 			d.resolveable = false
@@ -77,16 +77,15 @@ func (d *DefDNS) setup() error {
 				return nil
 			}
 			d.err = err
-			return d.err
-			//continue
+			continue
 		}
 		sort.Strings(addrs)
 		d.resolveable = true
 		d.addrs = addrs
 		d.err = nil
 		return nil
-	//}
-	//return d.err
+	}
+	return d.err
 }
 
 func (d *DefDNS) Addrs() ([]string, error) {
--- a/system/dns.go
+++ b/system/dns.go
@@ -197,7 +197,9 @@ func LookupA(host string, server string) (addrs []string, err error) {
     return nil, fmt.Errorf("No DNS record found")
 	}
 	for _, ans := range r.Answer {
-		addrs = append(addrs, ans.(*dns.A).A.String())
+		if t, ok := ans.(*dns.A); ok {
+			addrs = append(addrs, t.A.String())
+		}
 	}
 
 	return
@@ -216,7 +218,9 @@ func LookupAAAA(host string, server string) (addrs []string, err error) {
     return nil, fmt.Errorf("No DNS record found")
 	}
 	for _, ans := range r.Answer {
-		addrs = append(addrs, ans.(*dns.AAAA).AAAA.String())
+		if t, ok := ans.(*dns.AAAA); ok {
+			addrs = append(addrs, t.AAAA.String())
+		}
 	}
 
 	return
@@ -235,7 +239,9 @@ func LookupCNAME(host string, server string) (addrs []string, err error) {
     return nil, fmt.Errorf("No DNS record found")
 	}
 	for _, ans := range r.Answer {
-		addrs = append(addrs, ans.(*dns.CNAME).Target)
+		if t, ok := ans.(*dns.CNAME); ok {
+			addrs = append(addrs, t.Target)
+		}
 	}
 
 	return
@@ -254,8 +260,10 @@ func LookupMX(host string, server string) (addrs []string, err error) {
     return nil, fmt.Errorf("No DNS record found")
 	}
 	for _, ans := range r.Answer {
-		mxstring := strconv.Itoa(int(ans.(*dns.MX).Preference)) + " " + ans.(*dns.MX).Mx
-		addrs = append(addrs, mxstring)
+		if t, ok := ans.(*dns.MX); ok {
+			mxstring := strconv.Itoa(int(t.Preference)) + " " + t.Mx
+			addrs = append(addrs, mxstring)
+		}
 	}
 
 	return
@@ -274,7 +282,9 @@ func LookupNS(host string, server string) (addrs []string, err error) {
     return nil, fmt.Errorf("No DNS record found")
 	}
 	for _, ans := range r.Answer {
-		addrs = append(addrs, ans.(*dns.NS).Ns)
+		if t, ok := ans.(*dns.NS); ok {
+			addrs = append(addrs, t.Ns)
+		}
 	}
 
 	return
@@ -293,12 +303,13 @@ func LookupSRV(host string, server string) (addrs []string, err error) {
     return nil, fmt.Errorf("No DNS record found")
 	}
 	for _, ans := range r.Answer {
-		sr := ans.(*dns.SRV)
-		prio := strconv.Itoa(int(sr.Priority))
-		weight := strconv.Itoa(int(sr.Weight))
-		port := strconv.Itoa(int(sr.Port))
-		srvrec := strings.Join([]string{prio, weight, port, sr.Target}, " ")
-		addrs = append(addrs, srvrec)
+		if t, ok := ans.(*dns.SRV); ok {
+			prio := strconv.Itoa(int(t.Priority))
+			weight := strconv.Itoa(int(t.Weight))
+			port := strconv.Itoa(int(t.Port))
+			srvrec := strings.Join([]string{prio, weight, port, t.Target}, " ")
+			addrs = append(addrs, srvrec)
+		}
 	}
 
 	return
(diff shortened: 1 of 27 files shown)


pysysops:
Currently this only provides DNS A record lookup (IPv4). I'll work on adding IPv6 support before adding the ability to perform various QTYPE lookups.

pysysops:
I've had a go at enabling advanced users to specify types of DNS record to query. Currently the following record types are available: 

- A
- AAAA
- CNAME
- MX
- NS
- PTR
- SRV
- TXT

The syntax to check a specific record type is:
```
dns:
  # Validate a CNAME record
  CNAME:pysysops.github.io:
    resolveable: true
    server: 8.8.8.8
    addrs:
    - "github.map.fastly.net."

  # Validate a PTR record
  PTR:8.8.8.8:
    resolveable: true
    server: 8.8.8.8
    addrs:
    - "google-public-dns-a.google.com."
``` 

I'm aware that some parts of the code might not be perfect or could be done differently so I'd be grateful for any feedback that can help me improve on this.

All tests and docs have been updated appropriately.

aelsabbahy:
First and foremost, I'd like to apologize for the long delay in responding to this. Overall, this looks really awesome! Lots of great features and a major improvement to the dns checker.

A couple of notes:
1) There are scenarios where the code panics, for example:
```yaml
dns:
  pysysops.github.io:
    resolveable: true
    addrs:
    - github.map.fastly.net.
    timeout: 0
    server: 8.8.8.8
```
```
goss validate
panic: interface conversion: dns.RR is *dns.CNAME, not *dns.A
...
```

2) Docs format has changed a bit, so the manual needs to be rebased.

pysysops:
Hey,

Thanks for the review and catching that nasty error. For some reason I had assumed that asking for an A record would return the A record (flatten the CNAME).

I'll fix this, adding a test for it (I've got a domain for the purposes of DNS testing - dnstest.io) and amend documentation to fit the rework.

Thanks, 

Tim Birkett.

aelsabbahy:
Thanks, again.. awesome enhancement, love it!

pysysops:
The current integration tests fail occasionally due to the addresses github.io addresses I've used. I'll setup some stub DNS records shortly and update the tests / docs to use records I can guarantee won't change.

pysysops:
OK, All changes work well and I got rid of the RR type mismatch error by ensuring only the appropriate RR type is returned like:

```go
 	for _, ans := range r.Answer {
		if t, ok := ans.(*dns.A); ok {
 			addrs = append(addrs, t.A.String())
 		}
  	}
```

pysysops:
@aelsabbahy - currently my dnstest.io domain is down whilst I harass my registrar into getting glue records / DNSSEC functional so I can use my own NS servers. 

I'll comment again once working... soon I hope.

pysysops:
After transferring my domain I was able to get everything up and running including DNSSEC. This will mean I can add further functionality moving forwards such as DNSSEC / DS record validation, SOA validation (ie serial number matches etc...) making goss even more useful for validation of DNS Authoritative servers etc.

For now I think this enhancement is ready to review and release if all goes well. Future improvements can be made iteratively as it becomes clear it might add value to goss.

aelsabbahy:
Did a little bit of testing with this, awesome job. Just needs `goimports` and it's good to go.


Thanks for all these contributions, this PR is an awesome enhancement. :+1:

pysysops:
This was quite a fun piece of work and learned a lot about various things here. This PR will allow us ops type folks to validate our DNS setups and records against specific DNS servers (corporate / public). 

Goss is a fantastic tool! So often Ops dinosaurs don't automate server / system validation because the current tool set out there often takes a bit of Ruby development to get going and it's a barrier to the "I'm not a dev" people out there. Goss has closed that gap.

aelsabbahy:
Merged the other two CRs one of them conflicted with this one. Once conflicts are resolved I'll merge.

Thanks again for all your contributions.

pysysops:
Annoyingly it seems that either Travis has intermittent DNS abilities or there's some finite limit on DNS requests... Occasionally builds will fail and then complete successfully the next run. Another interesting thing is that it is usually (if not always) the precise Docker container that times out or fails.

I'll do some investigation into this before I'd recommend merging it...

aelsabbahy:
I've had issues in the past with goss network features and travis, usually bumping the timeout helps..

pysysops:
I had a look at a similar issue that the go devs were having and added retry logic similar to: https://github.com/golang/go/blob/master/src/net/dnsclient_unix.go#L189-L223 but with a small sleep between retries.

The latest commit has (hard-coded) 5 retries with a 20ms sleep between retries. It's fixed the build :) and my tests go through on an unreliable (~5% packet loss) connection. Part of me feels bad for hard-coding the retries / sleep but another part of me thinks it might be good enough (MVP) and pulling those settings out to be configurable might be unnecessary over-engineering.

pysysops:
Managed to go through successfully twice in a row which just wasn't happening previously. So if you're happy with the change it should be good to merge.

aelsabbahy:
Not a huge fan of the retry/sleep loop.. but I understand that it might end up being the best option.

I think I'd like to test this out a bit further and get some metrics on how many retried on average we're needing and whether this issue is Travis related or not.

pysysops:
I can't say that I'm a huge fan of them either...

I did a fair bit of wiresharking to observe DNS traffic and saw that all query requests are sent as expected but not all responses were received causing goss dns to hit the configured timeout. I had a fmt.Printf.... inside the loop to output the error and retry count, the error is always a timeout.

The main problem here (and the reason increasing timeouts doesn't get the results we'd like) is the connectionless nature of DNS/UDP.

Programmatic retries seem to be how almost every DNS client handles the internet.

I'm sure if we used a DNS resolver on the local network there would be no issue but with Travis CI and public DNS resolvers retries may be a necessity.

The 20ms sleep is unnecessary as if a packet is lost the test will exceed the configured timeout before retrying.

On the subject of timeouts. I'd also love to investigate miekg's dns code a bit more and make sure that any sockets left open by "Exchange" get closed properly or explicitly on timeout in goss.

pysysops:
Looking at Travis troubleshooting guide I think they accept that there are intermittent issues around network connectivity: https://docs.travis-ci.com/user/common-build-problems/#My-builds-are-timing-out. I had it timeout on one of the before_install tasks.

I've made some changes to make sure dns.Client timeout is set to the same as the configured timeout. I've also reduced the number of retries to 3 and removed the unnecessary sleep time.

It feels like the retries logic is purely to keep builds green on Travis rather than be a function or feature so I feel that it's perhaps the most pragmatic approach.

I also flipped the bit on some of the tests around resolveable and found a bug. Because I had explicitly returned an error if no addresses are found. This caused the retry to kick in and tests for resolveable: false to fail which is not great. It now works as expected and returning no addresses does not result in retries or failed tests (unless resolveable: true) which is the expected behaviour.

pysysops:
Any feedback is as always welcome and appreciated. Thanks.

aelsabbahy:
Looks good, need help understanding one of the conditions though.

When do we run into a scenario where `err == nil AND len(addresses) == 0`  and what does that mean?

pysysops:
Looking at it... I too need some help understanding that condition.
I have removed the unnecessary code completely.

system/dns.go
```diff
@@ -23,40 +28,64 @@ type DefDNS struct {
...
-	addrs, err := lookupHost(d.host, d.Timeout)
-	if err != nil || len(addrs) == 0 {
-		d.resolveable = false
-		d.addrs = []string{}
-		// DNSError is resolvable == false, ignore error
-		if _, ok := err.(*net.DNSError); ok {
-			return nil
+	for i := 0; i < 3; i++ {
+		addrs, err := DNSlookup(d.host, d.server, d.qtype, d.Timeout)
+		if err != nil || len(addrs) == 0 {
```
aelsabbahy: This was the line I was trying to better understand.

pysysops: That's a piece of code that's in master. It looks like it is there to handle resolvable = false (*net.DNSError which indicates an empty response received from DNS server) vs an error from timeout or other connection error.

aelsabbahy: That awkward moment when you don't recognize your own code..

Anyways, this looks good to me, CI passing, did some local testing and it's looking good. If this is good to go on your end, I'll merge it in.

pysysops:
Haha... I don't recognise things I did 5 minutes ago.

I'm happy with this feature. Feel free to merge and release at your convenience.

aelsabbahy:
If you have time, can you take a look at the build failures on master. They seem to be related to this PR.

https://travis-ci.org/aelsabbahy/goss/builds/206463972

Seems dns tests on Travis are still flakey. :(

aelsabbahy:
Some findings, I noticed Google dns showing some dnssec errors.
http://dnsviz.net/d/c.dnstest.io/dnssec/

And:
https://dns.google.com/query?name=c.dnstest.io&type=CNAME&dnssec=true

Not sure if this is the culprit or not.

pysysops:
Apologies, this was my failure to schedule re-generation of RRSIG records which are only valid for 30 days. This resulted in records not being returned by any recursive DNS servers with DNSSEC configured like most of the DNS servers on the public web. 

I've fixed this now and re-sign the dnstest.io zone nightly.

aelsabbahy:
Awesome, thanks! Random thought I had during the outage is whether it would make sense to standup a bind/http docker container and keep the tests completely isolated from the Internet.

Just thinking out loud. No action items here.. :)

pysysops:
I had that thought back when I was getting so many timeouts, like a couple of stub containers to test against. I was looking at dnsmasq which is a really simple, lightweight dns (and dhcp) server but at the time it felt a bit overkill at the time.

I might try it out along with my next PR when I get round to it?

aelsabbahy:
Yeah, no rush on this. My main concern with dnstest.io is that it's an external dependency that may or may not be up.

Issue #173, opened 2016-11-17
Better handling of parser errors


velenux:
I was testing goss with a few regular expression checks for content inside a file:

```yaml
file:
  /etc/hosts:
    exists: true
    contains:
    - "!/::1[[:space:]]+localhost\./"
```

The regexp was making the YAML parser fail and it took me a while to find out why, the error message was just **"Unable to determine format from content"**.

I'm not entirely sure this is working as intended, but it turns out I had to use a double backslash (`\\`) to escape the `.`. A better error message (maybe report which line/character fails) would have helped a lot :)

I'm no programmer, but I could try to put in a pull request to clarify this in the documentation if you want.

pysysops:
I've opened a PR that gets some form of useful output for invalid file formats although in YAML when using double quotes you need to escape a range of characters just as you do in JSON.

I believe that in YAML you can use single quotes to avoid escaping characters.

velenux:
@pysysops thank you! The single quotes are actually a better option, I'm editing all my tests to use them

aelsabbahy:
I think this is resolved, if not.. feel free to reopen

Issue #157, opened 2016-10-02
Format flag for add and autoadd


CMeza99:
Right now if I do an autoadd sometimes is in YAML and other times in JSON (but always has the yaml extension). I think, if the user can control the format, that would be the best.

aelsabbahy:
Hmm, I've seen this before, but haven't been able to consistently reproduce. If you know a way to reproduce this let me know. Does specifying `goss -g goss.yaml add/autoadd` force it into YAML mode for you?

CMeza99:
`-g file.(yaml|json)` will format the output accordingly if it is a new file (which is awesome, is this documented?).

So I think I found what was happening to me. If `add` or `aa` has nothing to return, a file with `{}` get created. Any subsequent `add`s or `autoadd`s will be in JSON, regardless of `-g`. I noticed then when I attempted `goss -g apache.yaml aa apache` (which is incorrect for Ubuntu 14.04). Then I did `goss -g apache.yaml aa apache2` and it was JSON.

pysysops:
The suggested edit above (latest commit) is good for #173 and avoids an extra few CPU cycles of hitting setStoreFormatFromData when StoreFormat is already set. 

However a quick test of a non-existent resource in a new file still outputs an evil nil JSON object aka: `{}`

I'm kinda thinking here that a quick test for an empty resource before writing the file and printing a message and not an writing an empty file? 

What do you think?

pysysops:
Good spot @CMeza99 I could reproduce it every time. If the file doesn't exist and you add / autoadd a resource that doesn't exist you get a file with `{}` in. I've created a PR that tackles this and another issue but would appreciate your feedback to add some context around how you're using goss. 

The PR checks if the resulting config is empty, notifies the user that the configuration is empty and **doesn't** create a file however, I'm not sure if this behaviour would cause a problem (missing file) for your use-case, it might be better to create an empty file appropriately: 

JSON:
```
{}
```
YAML:
```
---
```

What do you think?

aelsabbahy:
@pysysops I could be wrong, but I believe the yaml driver in go writes an empty file as:
```yaml
{}
```

Which is valid YAML, but confusing. I don't think it's that big of an issue if subsequent adds are in the correct format. That said, I think the cleanest solution is not creating a file at all (your solution).

PR #176, opened 2016-11-20
Fix error output when an error is encountered in parsing file

pysysops:
This PR fixes the issue reported in #173 

Thought I'd have a look at this issue to learn more about the way in which goss loads configuration. Initially, it was a little confusing but I realised the reason for readJSONData in store.go is to handle configuration passed to STDIN.

The simplest answer seemed to be to amend validate.go and use readJSONData only when reading from STDIN and to make use of readJSON (which sets the StoreFormat based on file type) when passing a file config to goss.

It was also necessary to amend the setStoreFormatFromData function as the use of log.Fatalf was causing goss to exit before functions that output informative errors had run. 

 Examples of output from a `\.` in configuration:

JSON:
```
2016/11/20 22:51:23 Unable to determine format from content
Error: invalid character '.' in string escape code
```
 
YAML:
```
2016/11/20 22:49:07 Unable to determine format from content
Error: yaml: line 5: found unknown escape character
```

If configuration sent in via STDIN is invalid goss will exit with:
```
Error: StoreFormat unset
2016/11/20 23:36:27 Unable to determine format from content
```

Any feedback appreciated!

Thanks,

Tim.

### DIFF
--- a/store.go
+++ b/store.go
@@ -65,7 +65,7 @@ func ReadJSON(filePath string) GossConfig {
 // Reads json byte array returning GossConfig
 func ReadJSONData(data []byte) GossConfig {
 	if StoreFormat == UNSET {
-	  setStoreFormatFromData(data)
+		setStoreFormatFromData(data)
 	}
 	gossConfig := NewGossConfig()
 	// Horrible, but will do for now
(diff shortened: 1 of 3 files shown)


store.go
```diff
@@ -46,7 +46,7 @@ func setStoreFormatFromData(data []byte) {
 		StoreFormat = YAML
 		return
 	}
-	log.Fatalf("Unable to determine format from content")
+	log.Printf("Unable to determine format from content")
```
aelsabbahy: I'm thinking if we revert this back to Fatalf and update readJSONData to be:
```go
func ReadJSONData(data []byte) GossConfig {
    setStoreFormatFromData(data)
    if StoreFormat == UNSET {
        setStoreFormatFromData(data)
    }
   ...
```

The user will get clear error messages in all scenarios, what do you think?

pysysops: That sounds good, I'll try it out and test out autoadd'ing resources in both formats / replicating #157 issues.

pysysops:
Okay for #157 I think maybe informing the user that the resulting config file would be empty and that it can't be written. It doesn't exit with an error (although it could).

I've added some basic tests to validate that an empty file isn't created or destroyed by auto-adding a _nosuchresource_ to a file with resources in.

Of course there's a few options: Create an empty file... Or an empty object in JSON?

I'm not really sure what the right answer is here so some feedback / thoughts would be appreciated.

aelsabbahy:
Tested this, works great, just needs to be run through `goimports` then It's ready to be merged.

pysysops:
Darn Atom and it's 2 spaces... I now have the go-imports package installed and working nicely. Thanks!! 😃

PR #185, opened 2017-01-27
MD5 File checksums

pysysops:
Hey @aelsabbahy,

I thought a good addition to the `file` resource would be the ability to validate an md5 checksum of the file. It was pretty simple to do. 

I've added the appropriate code, docs and a test file / goss test. 

Output on error looks like: 
```
Failures/Skipped:

File: /etc/passwd: md5:
Expected
    <string>: 6b0364b78fb8968a2d15879125556c3c
to equal
    <string>: 12341234

Total Duration: 1.012s
Count: 52, Failed: 1, Skipped: 0'
```

Errors such as validating an md5 on a directory are handled:
```
Failures/Skipped:

/tmp: md5: Error: read /tmp: is a directory

Total Duration: 1.013s
Count: 54, Failed: 1, Skipped: 0'
```

Any feedback appreciated. Thanks!

### DIFF
--- a/system/file.go
+++ b/system/file.go
@@ -1,6 +1,7 @@
 package system
 
 import (
+	"crypto/md5"
 	"fmt"
 	"io"
 	"os"
@@ -8,7 +9,6 @@ import (
 	"strconv"
 	"strings"
 	"syscall"
-	"crypto/md5"
 
 	"github.com/aelsabbahy/goss/util"
 	"github.com/opencontainers/runc/libcontainer/user"
@@ -218,7 +218,7 @@ func realPath(path string) (string, error) {
 }
 
 func (f *DefFile) Md5() (string, error) {
-  var result []byte
+
 	if err := f.setup(); err != nil {
 		return "", err
 	}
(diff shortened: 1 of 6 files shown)


aelsabbahy:
Thanks for the awesome contributions so far. I'll try to process all three outstanding PRs early next week.

pysysops:
Thanks for a really cool project. It's got great potential and it's simplicity / ease of use is a big WIN.

As I think of useful enhancements I'll either open PR's or discuss them as issues as well as try to answer any user issues that I can.

system/file.go
```diff
@@ -8,6 +8,7 @@ import (
 	"strconv"
```
aelsabbahy: Please run this through `goimports` to fix indentation and ordering of imports.

system/file.go
```diff
@@ -214,3 +216,23 @@ func realPath(path string) (string, error) {
 
 	return realPath, err
 }
+
+func (f *DefFile) Md5() (string, error) {
```
aelsabbahy: The `result` variable can be dropped altogether, this code should do the same as the orig:
```golang
func (f *DefFile) Md5() (string, error) {
        if err := f.setup(); err != nil {
                return "", err
        }

        fh, err := os.Open(f.realPath)
        if err != nil {
                return "", err
        }
        defer fh.Close()

        hash := md5.New()
        if _, err := io.Copy(hash, fh); err != nil {
                return "", err
        }

        return fmt.Sprintf("%x", hash.Sum(nil)), nil
}
```

pysysops: Ah yes... That's clear after another look at it 👍 this was a super quick addition initially.

PR #186, opened 2017-02-06
fix explanation of "validate" over ssh in manual

yokogawa-k:
STDIN is currently not implicitly used (#133), so fixed the manual.


aelsabbahy:
Thanks

PR #190, opened 2017-02-15
Add Reference to Ansible "degoss" Role

naftulikay:
This project is an Ansible Role which installs, runs, and removes Goss and test cases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,6 +181,7 @@ package:
 
 ## Community Contributions
 * [goss-ansible](https://github.com/indusbox/goss-ansible) - Ansible module for Goss.
+* [degoss](https://github.com/naftulikay/ansible-role-degoss) - Ansible role for installing, running, and removing Goss in a single go.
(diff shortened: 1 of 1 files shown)


aelsabbahy:
Curious on how this differs from [goss-ansible](https://github.com/indusbox/goss-ansible)?

naftulikay:
Basically, goss-ansible is a simple Ansible module that allows you to intelligently run Goss from Ansible.

degoss is an Ansible role which includes goss-ansible and does the following in order:

 - create a temporary directory
 - install a version (or the latest version) of Goss into that temporary directory
 - copy a Goss spec file or series of spec files into that temporary directory
 - run Goss on all of those spec files
 - if an error occurs, pretty print the output
 - remove Goss and any traces that anything was run

My personal use-case which prompted the development of this was needing to test an AMI without leaving traces on it during the build. I have Packer run Ansible to install and configure the machine, then run degoss to validate it passes all tests. Once finished, there is no trace of Goss or the spec files left on the machine anywhere.

aelsabbahy:
Makes sense, thanks!

PR #197, opened 2017-02-21
Add ability to test HTTP response headers

pysysops:
Here's an attempt at the addition to the http test that I see some value in.

A few times in the past I've been bitten by the presence of, or the lack of various HTTP Headers.

Think 40M+ daily pageviews, 2 small VM's, a 98.7% CDN cache hit and an unwanted no-cache header...

As always I'm totally open to suggestions and questions. It works, but is it the most appropriate way? 

Thanks, 

Tim.

### DIFF
--- a/integration-tests/goss/alpine3/goss-expected-q.yaml
+++ b/integration-tests/goss/alpine3/goss-expected-q.yaml
@@ -104,9 +104,11 @@ http:
     no-follow-redirects: true
     timeout: 5000
     body: []
+    headers: {}
   https://www.google.com:
     status: 200
     allow-insecure: false
     no-follow-redirects: false
     timeout: 5000
     body: []
+    headers: {}
--- a/integration-tests/goss/alpine3/goss-expected.yaml
+++ b/integration-tests/goss/alpine3/goss-expected.yaml
@@ -143,9 +143,11 @@ http:
     no-follow-redirects: true
     timeout: 5000
     body: []
+    headers: {}
   https://www.google.com:
     status: 200
     allow-insecure: false
     no-follow-redirects: false
     timeout: 5000
     body: []
+    headers: {}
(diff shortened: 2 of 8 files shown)


pysysops:
Apologies for the build failures. I'm sat on an Island in Indonesia with an internet connection too poor to run a complete test myself here 😢

aelsabbahy:
This looks great. Two observations:

1) This looks like a lot of code duplication, I wonder if we can minimize that somehow with a refactor of validateContains.

2) Would it be beneficial to add support for checking that a header key does not exist?

pysysops:
Those are 2 thoughts I had myself after thinking further about this feature / reading up and learning more and more about golang so I'm glad you have raised them.

I'm thinking we can possibly accept various types using empty interface{} and depending on the type passed in, decide how best to validate the data? Although, it's possible there's a better way?

As you know my additions are never perfect in my first commits. But I open PRs early so that I can get feedback, share ideas and form some discussion as things iteratively improve.

aelsabbahy:
So thinking about the refactor a little more. It'd probably be best if I get rid of ValidateContains all together and just write a custom gomega matcher that does the same thing.

This way, everything would go through ValidateValue and delegate the matching to gomega.

pysysops:
Coincidently I was getting more familiar with how gomega was used elsewhere in Goss and gomega in general before even touching this again.

aelsabbahy:
So, had an hour or so to tinker with this over the weekend.

Check out my `contains_gomega` branch, that's where I'll be doing work to move the `ValidateContains()` to `ValidateValues()` + using custom gomega matcher:

https://github.com/aelsabbahy/goss/blob/contains_gomega/resource/custom_matchers.go

This isn't tested at all, so not even sure it works.. just some initial thoughts in code.

pysysops:
I'll close this for now as I'm re forking / cloning the goss repo.

Issue #203, opened 2017-03-11
Goss doesn't escape output in generated JUnit XML


AlexandreCarlton:
When generating JUnit XML, Goss does not escape special characters which can result in malformed XML.

As an arbitrary example:

```yml
command:
  echo 'foo' > /dev/null:
    exit-status: 0
```
The resulting XML is:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<testsuite name="goss" errors="0" tests="1" failures="0" skipped="0" time="0.002" timestamp="2017-03-11T06:52:24Z">
<testcase name="Command echo 'foo' > /dev/null exit-status" time="0.002">
<system-out>Command: echo 'foo' > /dev/null: exit-status: matches expectation: [0]</system-out>
</testcase>
</testsuite>
```
Notably, the `>` and `'` characters are not escaped in both the text of `system-out` and the `name` attribute of `testcase`. It would be really great if that could be fixed. :)

LudwigHoff:
+1

PR #204, opened 2017-03-11
Escape special characters in JUnit XML output

AlexandreCarlton:
Previously, characters such as `>` in tests were not escaped in the
generated JUnit XML output (resulting in malformed output).

Now, such characters are escaped.

This is my first experience with Go, so I'm definitely keen to improve
this.

References #203


aelsabbahy:
Apologies, I completely forgot about this PR. I'll merge it in as part of the next release.

PR #216, opened 2017-03-28
Add a silent output

ouroboros8:
When using goss as a helathchecking endpoint, it may be desirable to
suppress verbose test output so that potential malicious entities can't
scrape healthcheck endpoints for system information.

Add the `silent` output, which simply checks whether any tests have
failed and returns 1 or 0 accordingly, without printing any output.

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,6 +204,7 @@ package:
 * JUnit
 * nagios - Nagios/Sensu compatible output /w exit code 2 for failures.
 * nagios_verbose - nagios output with verbose failure output.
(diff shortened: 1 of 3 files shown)


aelsabbahy:
Awesome! Code looks good, can you also update the documentation as part of this pull request.

ouroboros8:
Oh of course! My bad.

ouroboros8:
I've added some one-line output descriptions. They're maybe a bit long, but I thought it was worth expanding on the use-case, since it's perhaps non-obvious.

aelsabbahy:
I'll merge/release with next release. Thanks :+1:

PR #218, opened 2017-04-02
Fixed typo in dgoss Mac instructions

craigbarrau:
Just fixed a simple typo in the Mac instructions
`- o` should be `-o`
otherwise, the user gets this:
```
curl: option -: is unknown
curl: try 'curl --help' or 'curl --manual' for more information
```


aelsabbahy:
Thanks!

PR #221, opened 2017-04-06
HTTP Connection Leak

OmarDarwish:
`ContainsValue` now closes streams if they are of type `io.ReadCloser`

The system HTTP resource also does not cache connections anymore. I don't think they were actually reused since each new http resources created a new client.


aelsabbahy:
Thanks!

Issue #222, opened 2017-04-09
Add glob pattern support to gossfile


AlexandreCarlton:
In order to add easily add tests to an existing goss suite it would be great if the `gossfile` feature supported the use of glob patterns. This could be used to specify all files in a folder when running a test suite:

```yml
gossfile:
  goss_*.yaml: {}
```

Instead of manually having to append filenames to the main goss file, we could merely drop in a file matching the `goss_*.yaml` pattern and be on our merry way.

PR #223, opened 2017-04-09
Add glob pattern support to gossfile

AlexandreCarlton:
This will ease the addition of tests to an existing goss test suite.

All that is necessary is to drop in a file matching the existing pattern. As an example:

```yml
gossfile:
  /etc/goss.d/*.yaml: {}
```

Unfortunately, I wasn't sure where to place this in the existing integration tests; where might a good place for such a test be?

References #222

### DIFF
--- a/docs/manual.md
+++ b/docs/manual.md
@@ -528,11 +528,12 @@ file:
 
 
 ### gossfile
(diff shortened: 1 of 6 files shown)


aelsabbahy:
This looks great!

For integratiin tesrs maybe change these imports: https://github.com/aelsabbahy/goss/blob/master/integration-tests/goss/wheezy/goss.yaml#L26

To: `gossfile: ["../goss-s*.yaml"]` in all applicable server types.

AlexandreCarlton:
Done.

Thank you for creating goss, it's an incredibly useful tool. :)

PR #225, opened 2017-04-22
Add Gomega HaveKeyWithValue

OmarDarwish:
In my fork, I have an EC2 resource that returns a map of instance tags. This Gomega matcher allows you to do checks like the following:

```yml
aws_ec2:
  tags:
     have-key-with-value:
       foo: 1
       bar: baz
```

### DIFF
--- a/resource/gomega.go
+++ b/resource/gomega.go
@@ -118,13 +118,8 @@ func mapToGomega(value interface{}) (subMatchers []types.GomegaMatcher, err erro
 		return nil, fmt.Errorf("Matcher expected map, got: %t", value)
 	}
 
-	var key, val types.GomegaMatcher
-	for k, v := range valueI {
-		key, err = matcherToGomegaMatcher(k)
-		if err != nil {
-			return
-		}
-		val, err = matcherToGomegaMatcher(v)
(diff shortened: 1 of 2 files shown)


resource/gomega.go
```diff
@@ -95,6 +106,29 @@ func matcherToGomegaMatcher(matcher interface{}) (types.GomegaMatcher, error) {
...
 
+func mapToGomega(value interface{}) (subMatchers []types.GomegaMatcher, err error) {
+	valueI, ok := value.(map[string]interface{})
+	if !ok {
+		return nil, fmt.Errorf("Matcher expected map, got: %t", value)
+	}
+
+	var key, val types.GomegaMatcher
+	for k, v := range valueI {
+		key, err = matcherToGomegaMatcher(k)
```
aelsabbahy: What does `matcherToGomegaMatcher()` on the key provide?

Can you give an example of this usage?

OmarDarwish: No use because of constraints with YAML. Removed.

OmarDarwish:
@aelsabbahy any updates with this PR and #233 #226?

PR #226, opened 2017-04-22
Adds Matching resource

OmarDarwish:
This resource lets you define arbitrary data to match and report against. I use this when piping in a generated file and using templates to match against its contents:

Let's say we have a `data.json` file that gets generated as part of some testing pipeline:

```json
{
  "instance_count": 14,
  "failures": 3,
  "status": "FAIL"
}
```

This could be piped into goss: `goss --vars data.json validate`

And then validated against:

```yaml
---
matching:
  check_instance_count: # Make sure there is at least one instance
    content: {{ .Vars.instance_count }}
    matches:
      gt: 0

  check_failure_count_from_all_instance: # expect no failures
    content: {{ .Vars.failure_count }}
    matches: 0

  check_status:
    content: {{ .Vars.status }}
    matches:
      - not: FAIL
```

### DIFF
--- a/docs/manual.md
+++ b/docs/manual.md
@@ -24,6 +24,7 @@
   * [interface](#interface)
   * [kernel-param](#kernel-param)
   * [mount](#mount)
+  * [matching](#matching)
   * [package](#package)
   * [port](#port)
   * [process](#process)
@@ -609,6 +610,30 @@ mount:
     filesystem: xfs
 ```
 
+### matching
+Validates specified content against a matcher. Best used with [Templates](#templates).
+
+```yaml
+matching:
+  has_substr: # friendly test name
+    content: some string
+    matches:
+      match-regexp: some str
+  has_2:
+    content:
+      - 2
+    matches:
+      contain-element: 2
+  has_foo_bar_and_baz:
+    content:
+      foo: bar
+      baz: bing
+    matches:
+      and:
+        - have-key-with-value:
+            foo: bar
+        - have-key: baz
+```
 
 ### package
 Validates the state of a package
--- a/docs/manual.md
+++ b/docs/manual.md
@@ -618,6 +618,39 @@ mount:
 ### matching
 Validates specified content against a matcher. Best used with [Templates](#templates).
 
+#### With [Templates](#templates):
+Let's say we have a `data.json` file that gets generated as part of some testing pipeline:
+
+```json
+{
+  "instance_count": 14,
+  "failures": 3,
+  "status": "FAIL"
+}
+```
+
+This could then be passed into goss: `goss --vars data.json validate`
+
+And then validated against:
+
+```yaml
+matching:
+  check_instance_count: # Make sure there is at least one instance
+    content: {{ .Vars.instance_count }}
+    matches:
+      gt: 0
+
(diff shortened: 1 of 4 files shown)


OmarDarwish:
This PR should be merged after #225 and #233

OmarDarwish:
Once #256 is merged, the tests for this should be passing

OmarDarwish:
@aelsabbahy do you think this will make it into the next release?

docs/manual.md
```diff
@@ -614,6 +615,30 @@ mount:
     filesystem: xfs
 ```
 
+### matching
```
aelsabbahy: Can you provide an example with templates for the readme? or a brief explanation like you did on this PR.

resource/matching.go
```diff
@@ -0,0 +1,107 @@
...
+	stub := func() (interface{}, error) {
+		return a.Content, nil
+	}
+
+	var results []TestResult
+	results = append(results, ValidateValue(a, "matches", a.Matches, stub, skip))
+	return results
+}
+
+func (ret *MatchingMap) UnmarshalJSON(data []byte) error {
```
aelsabbahy: Can this be moved to resource_list_genny.go and generated into resource_list.go? Or is there something special about this particular implementation?

OmarDarwish: I think this resource is unique in that it does not have a system resource pairing. The template in resource_genny attempts to instantiate a matching system resource. I can add a dummy resource if you think that approach is better.

docs/manual.md
OmarDarwish: Yes, can do.

resource/matching.go
aelsabbahy: We can leave it as is for now. If a second non-system resource comes up, then the dummy approach will be preferable.

docs/manual.md
OmarDarwish: Done.

aelsabbahy:
Was doing some thinking about this PR last night. Some high level thoughts on a V2 implementation of this.

What if this was a validator type, that can be used on any JASON/YAML file or url pointing to one. An optional attribute for both the http and file resource. The tricky part would be the template portion, or maybe we can see if go provides a library that does something like jq.

This implementation works, but feels limited and awkward to use the vars file for this.

Any thoughts?

elliotweiser:
@aelsabbahy (in response to your question...)

IMO, this starts to feel like feature-creep. I like the idea of JSON/YAML file validation, but I think that work should be externalized to a different tool/library that could then be called from GOSS. I know of tools that already do this e.g. [yaml_bot](skippyPeanutButter/yaml_bot). Incidentally, I think @skippyPeanutButter has started work on a Golang version of that tool ☝️(and is looking for collaborators). 

Personally, I would like to see this functionality moved to a validator for the `file` resource. GOSS could then call custom rules validation engines then be be called to validate its structure/contents.  Here's a rough example in my head...

```yaml
file:
  /path/to/my.json
    exists: true
    filetype: file
    validator:
      jsonValidatorLibrary:
        with_tests: /path/to/json/file/tests
```

To summarize, I think it's a good idea, but it feels slightly out-of-scope and should probably be handled in a separate library/project. 

That is my two cents, but I may be in the minority here.

PR #227, opened 2017-04-25
allow dgoss to handle yaml files in other directories

roelandvanbatenburg:
### Context
Dgoss expects the files in the current directory, I'm building multiple docker images so I need multiple yaml files.

### What has been done
- added a global environment variable (in the same style as the others) to define the location of the yaml files.

### How to test
- move your current yaml file to another directory and run ```GOSS_FILES_PATH=<dir> dgoss run <image>

### DIFF
--- a/extras/dgoss/dgoss
+++ b/extras/dgoss/dgoss
@@ -3,6 +3,7 @@
 set -e
 
 USAGE="USAGE: $(basename "$0") [run|edit] <docker_run_params>"
+GOSS_FILES_PATH="${GOSS_FILES_PATH:-.}"
 
 info() { echo -e "INFO: $*"; }
 error() { echo -e "ERROR: $*";exit 1; }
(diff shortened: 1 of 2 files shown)


aelsabbahy:
I'm thinking `get_docker_file` should also honor this new variable.

roelandvanbatenburg:
Makes sense, added it.

roelandvanbatenburg:
@aelsabbahy :bell:

aelsabbahy:
I'll test it tonight, if you would like this merged sooner, can you reply on here with errors for run and edit when the directory doesn't exist?

extras/dgoss/dgoss
```diff
@@ -33,7 +34,7 @@ run(){
 get_docker_file() {
     if docker exec "$id" sh -c "test -e $1" > /dev/null;then
         info "Copied '$1' from container to current directory"
```
aelsabbahy: Got a chance to test it today, the error case is a bit unclear, but the fix is easy.

I think if this info line is changed to the GOSS_FILES_PATH param instead of "current directory" the user would have all the info they need.

aelsabbahy:
I'm just gonna merge this and update it. Thanks for the feature!

PR #229, opened 2017-05-03
Fix regex example in docs.

MrSelfDestruction:


aelsabbahy:
Good catch, thanks!

blalor:
Dur…

PR #233, opened 2017-05-09
Refactor resource configuration validation

OmarDarwish:
This refactors the way that resource configurations are checked. The first time I encountered this code, I had a little difficulty understanding the intention and what was happening. I hope these changes make it a little bit clearer. 

This pull request is a precursor to #226, which uses the utility functions introduced in this PR.


aelsabbahy:
Looks good. However, I don't see `ValidateSection` being used in either this PR or #226.

OmarDarwish:
Ah you're right @aelsabbahy! I was using it for my AWS resources. I'll remove it now. Good catch!

PR #245, opened 2017-06-12
Add upstart check for /etc/init/<service>.override

marclop:
Checks wether the service is disabled with the upstart override file `/etc/init/<service>.override`.

Before this, even if you had manually overridden the service to be `manual` it would just check the upstart file `/etc/init/<service>.conf` and report the service as enabled when it wasn't.

### DIFF
--- a/integration-tests/test.sh
+++ b/integration-tests/test.sh
@@ -39,6 +39,9 @@ trap "rv=\$?; docker rm -vf $id; exit \$rv" INT TERM EXIT
 # Give httpd time to start up, adding 1 second to see if it helps with intermittent CI failures
 [[ $os != "arch" ]] && docker_exec "/goss/$os/goss-linux-$arch" -g "/goss/goss-wait.yaml" validate -r 10s -s 100ms && sleep 1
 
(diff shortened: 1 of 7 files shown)


aelsabbahy:
Looks good, can you modify the dockerfile/tests to validate this?

integration-tests/test.sh
```diff
@@ -39,6 +39,9 @@ trap "rv=\$?; docker rm -vf $id; exit \$rv" INT TERM EXIT
 # Give httpd time to start up, adding 1 second to see if it helps with intermittent CI failures
 [[ $os != "arch" ]] && docker_exec "/goss/$os/goss-linux-$arch" -g "/goss/goss-wait.yaml" validate -r 10s -s 100ms && sleep 1
 
+# Disable apache2 service on boot for "precise"
```
aelsabbahy: Can this be moved to the docker file.

Aside from that, this PR looks great!

aelsabbahy:
Thanks, will review, merge and release later in the week.

Looks good.

PR #249, opened 2017-06-20
Enhancement: Support for SHA-256 file checksums

elliotweiser:
Summary
-------

This change adds support for SHA-256 file checksums. Example:

```yaml
file:
  /goss/hellogoss.txt:
    exists: true
    sha256: 7f78ce27859049f725936f7b52c6e25d774012947d915e7b394402cfceb70c4c
```

This change includes:

* Go code (modeled from md5 functionality) for a sha256 equivalent
* A new (shared) integration test for `sha256` file checksums.
* Documentation update showcasing a `sha256` example.

Additional changes
------------------

I had to make some additional changes because `make deps && make` (as per
the `.travis.yml`) didn't work as is. There seemed to be changes in either
the Docker base images or in upstream package repos (e.g. a specific `apache2`
package version was just _gone_). I also noticed that the ArchLinux container
fails to start properly when invoked via the `test.sh` script because `/sbin/init`
doesn't exist. After digging through some community forums and stackoverflow posts
I didn't understand, I decided it wasn't worth the effort to "fix", so I replaced
the command with a `sleep 1h`. Fortunately, the container still gets cleaned up
and it seems to work fine.

To summarize:

* Integration test/Dockerfile updates
* ArchLinux Docker image doesn't have `/sbin/init` (anymore)

### DIFF
--- a/docs/manual.md
+++ b/docs/manual.md
@@ -518,6 +518,8 @@ file:
     filetype: file # file, symlink, directory
     contains: [] # Check file content for these patterns
     md5: 7c9bb14b3bf178e82c00c2a4398c93cd # md5 checksum of file
+    # A stronger checksum alternative to md5 (recommended)
+    sha256: 7f78ce27859049f725936f7b52c6e25d774012947d915e7b394402cfceb70c4c
   /etc/alternatives/mta:
     # required attributes
     exists: true
--- a/integration-tests/goss/goss-shared.yaml
+++ b/integration-tests/goss/goss-shared.yaml
@@ -24,6 +24,7 @@ file:
   "/goss/hellogoss.txt":
     exists: true
     md5: 7c9bb14b3bf178e82c00c2a4398c93cd
+    sha256: 7f78ce27859049f725936f7b52c6e25d774012947d915e7b394402cfceb70c4c
   "/tmp/goss/foobar":
     exists: false
     contains: []
--- a/integration-tests/test.sh
+++ b/integration-tests/test.sh
@@ -44,9 +44,9 @@ out=$(docker_exec "/goss/$os/goss-linux-$arch" --vars "/goss/vars.yaml" -g "/gos
 echo "$out"
 
 if [[ $os == "arch" ]]; then
-  egrep -q 'Count: 64, Failed: 0' <<<"$out"
+  egrep -q 'Count: 65, Failed: 0' <<<"$out"
 else
-  egrep -q 'Count: 78, Failed: 0' <<<"$out"
+  egrep -q 'Count: 79, Failed: 0' <<<"$out"
 fi
 
 if [[ ! $os == "arch" ]]; then
--- a/resource/file.go
+++ b/resource/file.go
@@ -18,6 +18,7 @@ type File struct {
 	Filetype matcher  `json:"filetype,omitempty" yaml:"filetype,omitempty"`
 	Contains []string `json:"contains" yaml:"contains"`
 	Md5      matcher  `json:"md5,omitempty" yaml:"md5,omitempty"`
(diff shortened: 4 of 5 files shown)


aelsabbahy:
Looks good, can you try to lock the apache version down to the new one.

It's a bit hackish, but I use the md5 of the dockerfiles to cache them and speed up the build.

Strange about arch, I can investigate further later..

elliotweiser:
Pinned the version to Apache version on wheezy to `2.2.22-13+deb7u8`. I actually had to make even more changes originally to get all of the `make` targets to work on my Mac (using Docker for Mac). Apparently some of those issues don't come up in the CI environment 🤷‍♂️ .

In any case, let me know if you need anything else to get this merged.

elliotweiser:
My observation about ArchLinux may have only surfaced when attempting to run said container on a Mac. I'll at least see if that part of this PR can be reverted...

elliotweiser:
Yep works in Travis CI 👍

Ready when you are @aelsabbahy 😄

elliotweiser:
@aelsabbahy Any updates on getting this reviewed/merged?

aelsabbahy:
I'll try to get to it this week. Sorry for the delay.

elliotweiser:
@aelsabbahy Just checking in. Any progress?

elliotweiser:
@aelsabbahy Rebased and resolved conflicts 👍

elliotweiser:
@aelsabbahy I shook out the regressions from Travis CI. Should be all fixed now. 

Since I had to update the apache version again on wheezy, I opened #253 for your consideration.

aelsabbahy:
Can you try reverting the `Dockerfile` to the one on master. I think That will maintain the old version and would make all the version changes in the test files unnecessary.

See:
https://travis-ci.org/aelsabbahy/goss/builds/250023452#L1961

elliotweiser:
Sure. Give me a few minutes...

elliotweiser:
I did as you asked, but it looks like a recent change introduced a somewhat brittle unit test (I observed that error in a few upstream Travis CI builds). I could amend and force-push until it "works", but it would probably make more sense to address the failing test first.

aelsabbahy:
@OmarDarwish it appears the have-key-with-value test is brittle. Can you take a look at it when you get a chance?

See: https://travis-ci.org/aelsabbahy/goss/builds/250194911

This is the second time I've seen it failing in the past 3 days or so.

elliotweiser:
I just reproduced it locally. If you run `make test` enough times, it shows up.

```
--- FAIL: TestMatcherToGomegaMatcher (0.00s)
	gomega_test.go:145: For input '{"have-key-with-value": { "foo": 1, "bar": "baz" }}': got *matchers.AndMatcher &{[0xc4202085a0 0xc4202085c0] <nil>}, want *matchers.AndMatcher &{[0xc420114f60 0xc420114f80] <nil>}
FAIL
FAIL	github.com/aelsabbahy/goss/resource	0.012s
ok  	github.com/aelsabbahy/goss/system	0.011s
?   	github.com/aelsabbahy/goss/util	[no test files]
?   	github.com/aelsabbahy/goss	[no test files]
make: *** [test] Error 1
```

aelsabbahy:
Probably has to do with golang maps being randomized. That's just a guess without digging into the code yet.

OmarDarwish:
@aelsabbahy that's my thought as well. Will dig into this today as soon as I can. My bad guys!

elliotweiser:
@OmarDarwish No worries. These things happen. Thanks for looking into it! 👍

aelsabbahy:
No worries, thanks for the quick response and contributions.

OmarDarwish:
#256 should fix flakiness

PR #256, opened 2017-07-05
Fix flakiness of have-key-with-value test

OmarDarwish:
The gomega `have-key-with value` test was flaky because of ranging over a map in Go is not deterministic so the order that submatchers being passed to `gomega.And` was unpredictable. This caused the test to sometimes pass and sometimes fail.

This commit fixes this by ensuring that `mapToGomega` sorts keys before iterating through map values.

### DIFF
--- a/resource/gomega.go
+++ b/resource/gomega.go
@@ -2,6 +2,7 @@ package resource
 
 import (
 	"fmt"
+	"sort"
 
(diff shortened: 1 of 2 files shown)


elliotweiser:
@OmarDarwish Thanks for this!

PR #257, opened 2017-07-05
Run resource/dns and sytem/http through gofmt

OmarDarwish:
I run gofmt before pushing, and these 2 files are "unformatted". This commit fixes that.


aelsabbahy:
Thanks, I really need to make that part of the CI process.

PR #265, opened 2017-07-10
Add json_oneline support

### LINKED ISSUE #247: Single line json output
Hi,

I'm using goss in a systemd timer and JournalBeat to ship the logs to ElasticSearch.
At the moment I'm using the nagios_verbose format and parsing it in logstash but if the json output was a single line json message I could use that directly with no need for parsing.

Can you make the json output format single line and maybe introduce a json_pretty that is multi line?

Thanks.

aelsabbahy:
Fixes #247

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,7 +203,6 @@ package:
 * TAP
 * JUnit
 * nagios - Nagios/Sensu compatible output /w exit code 2 for failures.
-* nagios_verbose - nagios output with verbose failure output.
 * silent - No output. Avoids exposing system information (e.g. when serving tests as a healthcheck endpoint).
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -203,7 +203,6 @@ package:
 * TAP
 * JUnit
 * nagios - Nagios/Sensu compatible output /w exit code 2 for failures.
-* nagios_verbose - nagios output with verbose failure output.
```
taurus-forever: is format nagios_verbose deprecated?

aelsabbahy: Nah, just felt the readme should be at a higher level. It's still in the manual.

Issue #269, opened 2017-07-29
Add the ability to run dgoss when the docker daemon is not local


rmb938:
I have a use case where I want to run dgoss but the docker daemon is not on the local machine and the file system is not shared. 

My idea would be to add a flag i.e ```--not-local``` or something and run ```https://goss.rocks/install``` inside the container to be tested then docker cp the goss yaml files.

elliotweiser:
👍 

 I was playing with a similar concept outside of dgoss, but this seems like a good/valid use-case for it.

Implementation idea: only `docker cp` files instead of mounting volumes during `docker run`.  In theory this would make `dgoss` compatible against both local and remote docker daemons.

aelsabbahy:
Docker cp would be interesting. It would slow down dgoss a little bit and not allow testing of the log content.. but those could be disclaimers. Maybe dgoss could have a cp vs mount mode that a user can toggle though env vars.

I probably won't get around to this, but if someone submits a backwards compatible PR, I would review/merge.

elliotweiser:
I may have some to look into this next week

rmb938:
Here is the script with the docker cp modifications I made. I didn't add a flag, just replaced everything with docker cp so it will need some work before it is ready for a pr https://gist.github.com/rmb938/bfc92e5cfffd69ba66aabb9bb0e656eb

aelsabbahy:
Closed by #271 it's on the master branch now. I usually don't cut releases for dgoss changes, so it'll be on the github "releases" tab in the next release.

Issue #243, opened 2017-06-07
Printing the output of a bash script executed by a goss test


transcedentalia:
I have a goss test that is running a bash script. The bash script is printing some output with echo.  The output of the bash script is not visible in goss' output or anywhere in the console/terminal. 

Is it possible to see that output somehow? Browsing through the documentation I couldn't find anything helpful.

Thank you

aelsabbahy:
Currently no, command output us treated like file content. Goss loops over it line by line and fails or succeeds depending on if it's found.

This would be an interesting feature request to evaluate if this is something you want.

fracasula:
I think it'd be an interesting failure, especially when debugging failing tests. When that happens I'm forced to get into the container myself to run the failing tests manually. Something like a `--force-output` option to enable it on demand would be really nice I think.

:+1:

aelsabbahy:
Being tracked here now: https://github.com/aelsabbahy/goss/issues/483

PR #271, opened 2017-08-07
Add the ability to run dgoss when the docker daemon is not local. 

plispe:
I created pull request which is discussed in #269 and #243. All the changes should be backwards compatible.

### DIFF
--- a/extras/dgoss/dgoss
+++ b/extras/dgoss/dgoss
@@ -42,7 +42,7 @@ run(){
         info "Starting docker container"
         docker start $id > /dev/null
         ;;
-      *) error "Wrong goss files strategy used! Correct options are \"mount\" of \"cp\"."
+      *) error "Wrong goss files strategy used! Correct options are \"mount\" or \"cp\"."
     esac
 
     log_pid=$!
--- a/extras/dgoss/dgoss
+++ b/extras/dgoss/dgoss
@@ -25,9 +25,26 @@ run(){
     [[ -e "${GOSS_FILES_PATH}/goss.yaml" ]] && cp "${GOSS_FILES_PATH}/goss.yaml" "$tmp_dir"
     [[ -e "${GOSS_FILES_PATH}/goss_wait.yaml" ]] && cp "${GOSS_FILES_PATH}/goss_wait.yaml" "$tmp_dir"
     [[ ! -z "${GOSS_VARS}" ]] && [[ -e "${GOSS_FILES_PATH}/${GOSS_VARS}" ]] && cp "${GOSS_FILES_PATH}/${GOSS_VARS}" "$tmp_dir"
-    info "Starting docker container"
-    id=$(docker run -d -v "$tmp_dir:/goss" "${@:2}")
-    docker logs -f "$id" > "$tmp_dir/docker_output.log" 2>&1 &
+
+    # Switch between mount or cp files strategy
+    GOSS_FILES_STRATEGY=${GOSS_FILES_STRATEGY:="mount"}
+    case "$GOSS_FILES_STRATEGY" in
+      mount)
+        info "Starting docker container"
+        id=$(docker run -d -v "$tmp_dir:/goss" "${@:2}")
+        docker logs -f "$id" > "$tmp_dir/docker_output.log" 2>&1 &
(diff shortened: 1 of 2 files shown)


aelsabbahy:
I'm guessing the #243 link was a mistake.

Skimmed through the code and I like this approach. I'll test it and provide feedback this week.

extras/dgoss/dgoss
```diff
@@ -25,9 +25,26 @@ run(){
...
+        ;;
+      cp)
+        info "Creating docker container"
+        id=$(docker create "${@:2}")
+        info "Copy goss files into container"
+        docker cp $tmp_dir $id:/goss
+        info "Starting docker container"
+        docker start $id > /dev/null
+        ;;
+      *) error "Wrong goss files strategy used! Correct options are \"mount\" of \"cp\"."
```
elliotweiser: mount _or_ copy

elliotweiser:
It should be possible to copy the `goss.yaml` file out of the running docker container when using the `copy` option, no?

plispe:
@elliotweiser I think, that goss.yaml is copied only when you use `dgoss edit` not when `dgoss run`  is used. The  `dgoss edit` should work as expected.

extras/dgoss/dgoss
aelsabbahy: Cp is fine, I think it's the "of" instead of "or" that needs fixing.

plispe: yes. "of' is a typo. sorry for that.

elliotweiser:
You are correct about the `goss.yaml` not being copied out of the container when using `dgoss run`. 

I guess I'm responding to this statement:

> With the `'cp'` strategy you lose the ability to write tests or waits against the docker output.

Since the generated `goss.yaml` would normally live on the volume, retrieving it after `dgoss edit` is trivial. When using the `cp` mode, I think the original behavior can be preserved by executing a `docker cp <CONTAINER>:file file`. 

(I personally can't think of a good reason for developing tests against a remote docker daemon, but for the sake of thorough code review, I thought I'd point this out).

Overall, LGTM 👍

extras/dgoss/dgoss
elliotweiser: Yep just a typo: I meant `of` vs `or`.

Whatever your preference is for `copy` vs `cp` ;)

elliotweiser: Sorry about that... my brain did an auto-complete 😅

plispe:
> Since the generated goss.yaml would normally live on the volume, retrieving it after dgoss edit is trivial. When using the cp mode, I think the original behaviour can be preserved by executing a docker cp <CONTAINER>:file file.

There is actually no volume used with `dgoss edit` but, the generated `goss.yaml` is copied with `docker cp` command from the container.  Therefore `dgoss edit` is unaffected by my changes.

> With the 'cp' strategy you lose the ability to write tests or waits against the docker output.

This statement is about the following command  from `dgoss` script `docker logs -f "$id" > "$tmp_dir/docker_output.log" 2>&1 &` whose behaviour can not be preserved with `cp` strategy. Because you have not possibility to stream `docker log -f` output to the running container without a mounted volume.

elliotweiser:
I stand corrected: `dgoss edit` indeed doesn't use docker volumes. In which case...

:shipit: !

plispe:
@elliotweiser 

> (I personally can't think of a good reason for developing tests against a remote docker daemon, but for the sake of thorough code review, I thought I'd point this out).

The only use-case could be the use of [CicleCI Debugging Jobs Over SSH] (https://circleci.com/docs/2.0/ssh-access-jobs/#nav-button) as CircleCI runs jobs in [separated environments](https://circleci.com/docs/2.0/building-docker-images/#separation-of-environments). I debug sometimes my build over ssh.

elliotweiser:
Neat! I'll have to check that out. Thanks for the contribution and the thorough responses!

plispe:
you are welcome

aelsabbahy:
Apologies for the delay on this, looks great. Thanks for the contribution!

plispe:
No problem. I used my fork untill now. Thanks for merge.

PR #274, opened 2017-08-28
Add verbose and extended-perfdata flags

pysysops:
Here's a first draft of something to satisfy #259 

It adds a --verbose (-v) and a --extended-perfdata (-E) BoolFlag to both the validate and serve commands.

I've passed this in using a simple struct "OutputConfig" that has no affect on any output other than the nagios output currently.

With the --verbose flag the output is the same as the nagios_verbose output:
```
GOSS CRITICAL - Count: 76, Failed: 1, Skipped: 0, Duration: 1.008s
Fail 1 - DNS: localhost: addrs: doesn't match, expect: [["127.0.0.1","::1"]] found: [["127.0.0.1"]]
```

With the --extended-perfdata flag:
```
GOSS CRITICAL - Count: 76, Failed: 1, Skipped: 0, Duration: 2.124s|total=76 failed=1 skipped=0 duration=2.124s
```

With both flags:
```
GOSS CRITICAL - Count: 76, Failed: 1, Skipped: 0, Duration: 1.009s|total=76 failed=1 skipped=0 duration=1.009s
Fail 1 - DNS: localhost: addrs: doesn't match, expect: [["127.0.0.1","::1"]] found: [["127.0.0.1"]]
```


timbirk:
This does mean you can add --verbose and --extended-perfdata arguments to _any_ validate / serve command but it isn't implemented in any other output (currently) and doesn't throw a warning or error to the user although it _could_.

aelsabbahy:
I'm mixed on this implementation, it'll work, but not very clear for the user which flags are for which formatter.

I wonder how these approaches would compare:
```
# this would allow goss validate nagios --help to list the correct flags, but would not be backwards compatible
goss validate nagios --verbose

# or
goss validate --format nagios --format-options verbose,perfdata

# This wouldn't provide a working --help, but the format options are more clear that they vary from format to format
```

Thoughts?

pysysops:
Hey, thanks for the feedback, it was a first "draft" and as always I greatly appreciate feedback and suggestions it only helps create a better product and improves my taming and understanding of golang. 

I've committed a change to allow passing of format-options as a StringSliceFlag. This means options can be passed like:

```
goss validate --format nagios --format-options verbose  --format-options perfdata
```

or short form: 
```
goss validate --format nagios -o verbose  -o perfdata
```

I'm looking into making the help output more informative to the user and will hopefully hava a commit to improve it soon(ish).

pysysops:
Hi @aelsabbahy hop you're well :) 

Thought I'd revisit this PR. I've tidied up the commits a bit.

I thought I'd take an approach that is already used to add output-ers to the list of valid `--format` options to help output. I've also updated the manual with an example of output options and removed the nagios_verbose `Outputer`.

Help output:
```
NAME:
   goss-linux-amd64 validate - Validate system

USAGE:
   goss-linux-amd64 validate [command options] [arguments...]

OPTIONS:
   --format value, -f value          Format to output in, valid options: [documentation json json_oneline junit nagios nagios_verbose rspecish silent tap] (default: "rspecish") [$GOSS_FMT]
   --format-options value, -o value  Extra options passed to the formatter, valid options: [perfdata verbose] [$GOSS_FMT_OPTIONS]
   --color                           Force color on [$GOSS_COLOR]
   --no-color                        Force color off [$GOSS_NOCOLOR]
   --sleep value, -s value           Time to sleep between retries, only active when -r is set (default: 1s) [$GOSS_SLEEP]
   --retry-timeout value, -r value   Retry on failure so long as elapsed + sleep time is less than this (default: 0s) [$GOSS_RETRY_TIMEOUT]
   --max-concurrent value            Max number of tests to run concurrently (default: 50) [$GOSS_MAX_CONCURRENT]
```

Adding options to the cli that aren't implemented / don't exist for a format has no impact.

aelsabbahy:
Thank you for this, looks great!

Issue #272, opened 2017-08-21
Is goss available for armv7?


sayanee:
I realised that goss is [available for download](https://github.com/aelsabbahy/goss/releases) for `i386` or `amd64`. 

Is there a release for `armv7` to run on boards like Beaglebone and Raspberry PI?

aelsabbahy:
There was a request a while back for this #211.

I never followed up on it, but would review and merge a PR that added it.

sayanee:
Tracking in #211

Issue #211, opened 2017-03-21
ARM release


bamarni:
Would it be possible to have a release for linux arm? Thanks

aelsabbahy:
I'm not as familiar with the ARM side of go.. I notice there's a bunch of versions of it:
https://github.com/golang/go/wiki/GoArm#supported-architectures

Are they backwards compatible or would this add 4 new binary releases?

bamarni:
AFAIK they're pretty much backward compatible. I'm not that much familiar with the whole ARM ecosystem but ARMv5 seems very old, and was even considered for removal (https://github.com/golang/go/issues/17082).

I think a release with `GOARM=6	GOARCH=arm` should be fine for most ARM users.

elfgoh:
@aelsabbahy I think it will work fine just to support the default ARM version, as [suggested by Dave Cheney](https://dave.cheney.net/2015/08/22/cross-compilation-with-go-1-5). Or at least until there is a huge demand for other ARM versions

PR #275, opened 2017-08-30
Adding ARM building in Makefile

elfgoh:
This PR enables an ARM binary to be build by running `make release/goss-linux-arm`. The resulting binary was ran on Beaglebone black debian and works fine when running 

``` 
./goss-linux-arm  autoadd sshd
./goss-linux-arm validate
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,6 +41,11 @@ release/goss-linux-386: $(GO_FILES)
 release/goss-linux-amd64: $(GO_FILES)
 	$(info INFO: Starting build $@)
 	CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -ldflags "-X main.version=$(TRAVIS_TAG) -s -w" -o release/$(cmd)-linux-amd64 $(exe)
+release/goss-linux-arm: $(GO_FILES)
+	$(info INFO: Starting build $@)
+	CGO_ENABLED=0 GOOS=linux GOARCH=arm go build -ldflags "-X main.version=$(TRAVIS_TAG) -s -w" -o release/$(cmd)-linux-arm $(exe)
+
+
 
 release:
 	$(MAKE) clean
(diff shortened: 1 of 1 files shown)


elfgoh:
Addresses #272 and #211

aelsabbahy:
What is the default value of GOARM when left blank?
https://github.com/golang/go/wiki/GoArm#supported-architectures

aelsabbahy:
Also, need to add this new target as a dependency of build in the makefile.

elfgoh:
The default value is 6. Thought I have read a few sources stating that compiled ARMv6 binaries are backward compatible with v7 processors.

There are a few sources indicating that it is ideal to compile for ARMv6 as this will allow older RPI models to be supported. I recommend max compatibility using ARMv6

- https://www.alexruf.net/2016/01/16/cross-compile-with-go-1-5-for-raspberry-pi.html
> You my have recognized that the optional variable GOARM has been set to 7. It enables support for ARMv7 instead of using the default value 6 for ARMv6. The difference is, that while the application will not run on old Raspberry Pi models, it takes advantage of a few more floating point registers and operations of the ARMv7 CPU from the Raspberry Pi 2. If you are in doubt, leave this variable unset for maximum compatibility unless you have a reason to adjust it.

- https://dave.cheney.net/2015/08/22/cross-compilation-with-go-1-5
> Sure, knock yourself out, but that means your program won’t run on all models of the Raspberry Pi. The difference between GOARM=6 (the default) and GOARM=7 is enabling a few more floating point registers, and a few more operations that allow floating point double values to be passed to and from the ARMv7 (VPFv3) floating point co processor more efficiently. IMO, with the current Go 1.5 arm compiler, it’s not worth the bother.

elfgoh:
Great! So I guess there will be ARM binaries in the next release?

aelsabbahy:
Yup as of 3.5:

https://github.com/aelsabbahy/goss/releases/tag/v0.3.5

Wonder how many ARM users want to use goss :)

elfgoh:
1 is a start :D

Issue #278, opened 2017-09-05
matching not working


chiefy:
```
~/tmp
❯ goss -g ./goss.yaml v
Error: found 0 tests, source: ./goss.yaml

~/tmp
❯ cat goss.yaml
---
matching:
  has_2:
    content:
      - 2
    matches:
      contain-element: 2

~/tmp
❯ goss --version && uname -a
goss version v0.3.2
Darwin ROSELCDV0001SSU 16.4.0 Darwin Kernel Version 16.4.0: Thu Dec 22 22:53:21 PST 2016; root:xnu-3789.41.3~3/RELEASE_X86_64 x86_64
```

Same with `0.3.4` am I crazy?

signull:
:+1:
I am seeing the same issue.

signull:
possibly the issue is because the "matching" test resource is missing from 
https://github.com/aelsabbahy/goss/blob/master/resource/resource_list.go

signull:
made a PR: I think https://github.com/aelsabbahy/goss/pull/279 should correct the issue

PR #279, opened 2017-09-05
Adding Matching resource to resource_list

signull:
the testing resource "Matching" was missing from resource_list.go

This change addresses: https://github.com/aelsabbahy/goss/issues/278

### DIFF
--- a/resource/resource_list.go
+++ b/resource/resource_list.go
@@ -1513,3 +1513,100 @@ func (ret *HTTPMap) UnmarshalYAML(unmarshal func(v interface{}) error) error {
 	*ret = tmp
 	return nil
 }
+
+type MatchingMap map[string]*Matching
+
+func (r MatchingMap) AppendSysResource(sr string, sys *system.System, config util.Config) (*Matching, error) {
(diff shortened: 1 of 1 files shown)


OmarDarwish:
Hi @signull , thanks for looking into this!

There were 2 issues that needed to be addressed:
*  [`GossConfig.Matchings`](https://github.com/aelsabbahy/goss/blob/master/goss_config.go#L25) need to be added added to the `GossConfig` [constructor](https://github.com/aelsabbahy/goss/blob/master/goss_config.go#L44) and the [`mergeGoss`] function (https://github.com/aelsabbahy/goss/blob/master/goss_config.go#L101).  
* [`Mathching.Validate`](https://github.com/aelsabbahy/goss/blob/master/resource/matching.go#L30) needed to change it's function signature from `Validate(sys system.System) []TestResult` to `Validate(sys *system.System) []TestResult` to match the `Resource` [interface](https://github.com/aelsabbahy/goss/blob/master/resource/resource.go#L14)

I've pushed #280 which should fix this.

signull:
@OmarDarwish has made the correct fix

PR #291, opened 2017-09-21
Print dgoss info/error to stderr

aelindeman:
When using a different formatter (like json or junit), having the dgoss debug messages print to stdout along with the actual goss test result output is annoying when redirecting the output to a file, e.g. `dgoss run my-container > results.json`. This simply changes the `echo` functions to print to stderr instead.


aelsabbahy:
Makes sense.

Thanks for the contribution!

shreyasgune:
Woohoo!

Issue #191, opened 2017-02-15
Windows support


petemounce:
Similarly to #26, it would be great to get Windows support. I assume since you welcomed PRs for FreeBSD, you might also welcome ones that add Windows support?

(Obviously, Windows being what it is, I expect this would require more work than FreeBSD... ;) ).

JackLeo:
What are the missing bits given the existing PR?

petemounce:
Packages (maybe via chocolatey)?

JackLeo:
Something like this is more likely:

https://gallery.technet.microsoft.com/scriptcenter/Script-to-get-installed-80cd3798

by the looks of it - digging through the registry should produce the most universal results for base system

petemounce:
I really can't recommend taking a dependency on anything written in VB. wmic probably has a longer deprecation horizon; VB is very legacy.

JackLeo:
So it's either [wmic](https://msdn.microsoft.com/en-us/library/aa394531(v=vs.85).aspx) or [inline .NET](https://msdn.microsoft.com/en-us/library/microsoft.win32.registry.getvalue(v=vs.110).aspx?cs-save-lang=1&cs-lang=csharp#code-snippet-1) doing the query. For MVP I would suggest just calling wmic subprocess and retrieving the results then allow someone with deeper Windows knowledge to build further.

petemounce:
Definitely wmic is the easiest bar, though won't detect portable apps installed via chocolatey. I'd say that's going to be an edge case for systems administration vs consumer users.

petemounce:
Re: supported - wmi dates back a decade or so at least. Probably fine for the set of likely user population of this library.

JackLeo:
wmic should detect anything installed via an installer since that puts uninstaller registry entry. Packages like Docker could be both installed via an installer and chocolatey so there might be some edge cases with using them both.

petemounce:
Having come back to this, I'm now inclined to think that `package` should assert against package-manager installs only. If software is present on the machine, and the person wants to write tests to assert that, I bet they're _almost certainly_ already aware of chocolatey.

And, @JackLeo is entirely correct, it's possible for a thing to be installed via chocolatey and other means, to completely separate places.

Is it reasonable to make one (initial) chocolatey implementation, and one wmic-based? I saw there's [a `--format` argument to specify](https://github.com/aelsabbahy/goss/blob/master/docs/manual.md#package) which flavour...?

JackLeo:
[...]

PR #298, opened 2017-10-25
Decompose system/file.go for POSIX/Windows to allow building for Windows

eripa:
**Note**: The minor changes here was done to allow for an internal specific use case, I'm just putting this here in case someone wants to pick it up. Personally I probably won't be working more on this.

This change enables goss to build for Windows. Initial step for #191.

Some caveats:

- a test suite hasn't been created
   - cannot really use Docker for this, maybe Vagrant in the future?
   - this means that there are probably many bugs that I haven't seen
- I've only tested it on Windows 8.1, using cygwin x86_64
   - which also means that there are probably many more issues on "pure" Windows
- the Mode, Owner and Group function of `system/file.go` has been mocked with dummy data to build on Windows. Ideally these should be implemented in some platform native manner (that I'm not familiar with)

Example from my usecase:

```
vagrant@test_goss_windows ~
$ ./goss-windows.exe -g goss.yaml validate
.........

Total Duration: 0.033s
Count: 9, Failed: 0, Skipped: 0

vagrant@test_goss_windows ~
$ cat goss.yaml
file:
  C:\cygwin64\home\vagrant\some_file_that_doesnt_exist:
    exists: false
    contains: []
  C:\cygwin64\home\vagrant\consul.exe:
    exists: true
    mode: "0000"
    size: 43419136
    owner: Other Organization
    group: Other Organization
    filetype: file
    contains: []
command:
  /home/vagrant/disk_usage.exe -warning 50 -critical 60:
    exit-status: 0
    stdout: []
    stderr: []
    timeout: 10000
process:
  consul.exe:
    running: true

vagrant@test_goss_windows ~
$ uname -a
CYGWIN_NT-6.3 test_goss_windows 2.0.4(0.287/5/3) 2015-06-09 12:22 x86_64 Cygwin
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,14 +44,15 @@ release/goss-linux-amd64: $(GO_FILES)
 release/goss-linux-arm: $(GO_FILES)
 	$(info INFO: Starting build $@)
 	CGO_ENABLED=0 GOOS=linux GOARCH=arm go build -ldflags "-X main.version=$(TRAVIS_TAG) -s -w" -o release/$(cmd)-linux-arm $(exe)
-
-
+release/goss-windows.exe: $(GO_FILES)
+	$(info INFO: Starting build $@)
+	CGO_ENABLED=0 GOOS=windows GOARCH=amd64 go build -ldflags "-X main.version=$(TRAVIS_TAG) -s -w" -o release/$(cmd)-windows.exe $(exe)
 
 release:
 	$(MAKE) clean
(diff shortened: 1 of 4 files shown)


JackLeo:
what about other parts of the tool? ports, services, etc?

petemounce:
Packages (check via chocolatey)?

aelsabbahy:
closed by #585

PR #300, opened 2017-10-29
Add report flag for sending reports to remote endpoints

malnick:
This PR suggests a way to send the validate report to a remote endpoint. I want to add this functionality so I can run `goss validate` as a systemd unit on a timer. There were several ways I could have achieved the same functionality, but my circumstances led me to push over pull method. 

The primary circumstance was hosts coming up in an AWS autoscaling group, and trying to grok their IPs and keep those IPs in sync on an aggregating service. Configuring a systemd unit/timer with configuration management to send the report output to a well known DNS entry seemed like my best and most durable option. 

If you have a better idea for how this could be implemented I'm more than happy to go down any road suggested. My main goal is to be able to send reports from goss on an interval via push method to a remote host. 

#### Report Flag
I added a `-report` flag to the `validate` command. This is an optional `.StringFlag` that defaults to `localhost:9090`. 

#### Outputer Type 
I updated the [outputer type](https://github.com/aelsabbahy/goss/compare/master...malnick:malnick/report#diff-dd638c0b5108f6e96b5159e674df9106R18) with a `SetReportURL()` method that accepts a string and returns an error. 

My reasoning for doing it this way over adding a new `Outputer` implementation for `report` (which would simply push a JSON formatted POST to the URL) was the fact that I didn't want to be opinionated about the format of the POST body. JSON is an obvious choice, but in the future someone might want to add a prometheus format `Outputer` type or one of the already created `Outputer` implementations. 

In order to make sure the URL was correctly formatted I added a `error` to the return of `output.getOutputer` [signature](https://github.com/aelsabbahy/goss/compare/master...malnick:malnick/report#diff-4fbfd0840928e279693ded5ea4cde702R59). I felt like this was a good focus point for validating the URL present in the `report` flag. 

#### JSON Outputer Implementation
I [updated](https://github.com/aelsabbahy/goss/compare/master...malnick:malnick/report#diff-dedf33f690ad3522cdba12a0df204ecfR54) the `Json{}` implementation with a `makeMap()` function. This was an artifact from my original plan to make a report `Outputer` type and I didn't want to duplicate the code that created the JSON. I left it as is since it seemed like a nice consolidation of that bit of code (happy to revert it though). 

#### URL Input
Getting the URL from the CLI flag (or some other place like ENV) also affected my thinking around this implementation. I really didn't have a good plan for this and went with a "just make it work" ethos for now. 

#### Tests & Thoughts
I ran some initial manual tests, but since this was looking like a hefty change I wanted to open the PR first before dedicating time to an effort that might be rejected.

### DIFF
--- a/cmd/goss/goss.go
+++ b/cmd/goss/goss.go
@@ -49,6 +49,12 @@ func main() {
 					Usage:  fmt.Sprintf("Format to output in, valid options: %s", outputs.Outputers()),
 					EnvVar: "GOSS_FMT",
 				},
+				cli.StringFlag{
+					Name:   "report",
+					Value:  "localhost:9090",
+					Usage:  "Optional report URL to send output to",
+					EnvVar: "GOSS_REPORT",
+				},
 				cli.BoolFlag{
 					Name:   "color",
 					Usage:  "Force color on",
--- a/outputs/documentation.go
+++ b/outputs/documentation.go
@@ -10,6 +10,8 @@ import (
 
 type Documentation struct{}
 
+func (r Documentation) SetReportURL(stringified string) error { return nil }
+
 func (r Documentation) Output(w io.Writer, results <-chan []resource.TestResult, startTime time.Time) (exitCode int) {
 	testCount := 0
 	var failedOrSkipped [][]resource.TestResult
--- a/outputs/http.go
+++ b/outputs/http.go
@@ -1,0 +1,22 @@
+package outputs
+
+import (
+	"bytes"
+	"fmt"
+	"net/http"
+	"net/url"
+)
+
+func postReport(json []byte, u *url.URL) error {
+	resp, err := http.Post(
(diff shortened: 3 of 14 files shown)


aelsabbahy:
Hey @malnick, thanks for the PR and detailed explanation.

There are a few concerns I have with this approach as it feels like it's building too much into the Goss tool and it opens up pandora's box in what should be supported for remote endpoints.. different protocols, authentication, etc.

I think a much simpler approach for this would be a wrapper script that runs goss and `curl -X POST` the result, this would be a few lines of shell and would allow users to easily tweak this.

I would be open to a wrapper script/readme being submitted to the [extras/](https://github.com/aelsabbahy/goss/tree/master/extras) folder, similar to how [dgoss](https://github.com/aelsabbahy/goss/tree/master/extras/dgoss) is managed.

Issue #164, opened 2016-10-14
Add tests for environment variables


mspiegel:
Is there any interest in adding tests that checked against environment variables? The test could check for the existence of the variable or check that the variable is a specific value. I know this request is a little strange and possibly not a good usage of goss. Since the other classes of tests are idempotent to which process is running the tests. The environment variable tests would have to make sure that the environment variables were populated before goss was run. It would still be a nice feature to have.

If there's an interest in this feature I can make an attempt at submitting a pull request.

aelsabbahy:
I would accept a pull request on this feature. I had thought about it in the past as I see some benefit to having this test when using `goss` with docker, or when using `goss` to verify that pre-conditions are in-place before executing a service.

Take a look at `kernel-param` for reference.

mspiegel:
Thanks! I'll try to work on it next week.

aelsabbahy:
Cool, just a heads up, you'll need to use genny and run `make gen` when modifying the resource list.

Apologies in advance for the bad code. Not sure if it's my amateur Golangism or if it's a consequence of no generics.

mspiegel:
I'm getting an error when I run "make deps"
```
[ERROR]	Failed to set version on github.com/docker/docker to 0f5c9d301b9b1cca66b3ea0f9dec3b5317d3686d: Unable to update checked out version
```

aelsabbahy:
Just tested it with glide v0.12.2 and it worked. travis-ci is using `0.10.2` which was working.

What version of glide do you have? Any chance its an intermittent issue?

mspiegel:
Tried again this morning. Checked out a clean version and got off the VPN. "make deps" returns the same error.

aelsabbahy:
It looks like the `glide` code that's throwing the error is here: https://github.com/Masterminds/glide/blob/3895d9e41edbe9186d65c9dde9c38148d9bbd49c/vendor/github.com/Masterminds/vcs/git.go#L134

I would try to:
```
# from the goss checked out folder
cd vendor/github.com/docker/docker
git fetch origin
```

And see what that says, depending on the output, maybe nuke away the `vendor/` folder all together and try running glide again? My guess is that it's either a weird `glide` issue, a network problem or corrupt `vendor/` dir?

Hope that helps.

bryanlatten:
Looking forward to this work @mspiegel - can I make a feature request within your feature request? It would be great if I could use environment variables within my tests. For example, instead of saying 
```
file:
[...]

PR #306, opened 2017-12-10
add env-var resource

twhiston:
W.I.P commit for issue #164 

It still needs some integration tests adding, but I thought I would run it past you first to see if you agree with the direction this is going.

it currently just captures the name and the value, as per the kernal-param implementation, but I suppose it could be extended to capture a state of presence and value, so you could either check for its existence or for an explicit value.

### DIFF
--- a/integration-tests/goss/goss-shared.yaml
+++ b/integration-tests/goss/goss-shared.yaml
@@ -177,3 +177,6 @@ matching:
         - have-key-with-value:
             foo: bar
         - have-key: baz
+env-var:
+  GOSS_TEST: true
+
--- a/add.go
+++ b/add.go
@@ -150,6 +150,13 @@ func AddResource(fileName string, gossConfig GossConfig, resourceName, key strin
 			os.Exit(1)
 		}
 		resourcePrint(fileName, res)
+	case "EnvVar":
+		res, err := gossConfig.EnvVars.AppendSysResource(key, sys, config)
+		if err != nil {
+			fmt.Println(err)
+			os.Exit(1)
+		}
+		resourcePrint(fileName, res)
 	default:
 		panic("Undefined resource name: " + resourceName)
 	}
--- a/cmd/goss/goss.go
+++ b/cmd/goss/goss.go
@@ -313,6 +313,14 @@ func main() {
 						return nil
 					},
 				},
+				{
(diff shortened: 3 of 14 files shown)


twhiston:
@aelsabbahy please can you give me some guidance on what you expect in the tests and what else i need to get this PR accepted?
To be honest I find the current (non go test) testing setup a little confusing, and trying to run the tests on a mac is a bit of a non starter as well, so it would be super useful to get a rough outline of what I need to do and the general methodology of the testing. Thanks very much!

aelsabbahy:
Unfortunately the testing for goss isn't the most ideal. Currently the Docker/e2e tests are the main form of testing. There are a few `*_test.go` files, but not for the changes you're adding.

To add tests for this:
1. Update the Dockerfiles if needed
2. Update the integration-tests/goss/**/*.yaml as needed
3. Update the `integration-test/test.sh` with the correct test count

Your current run is failing with:
```
+docker exec goss_int_test_centos7_amd64 /goss/centos7/goss-linux-amd64 --vars /goss/vars.yaml -g /goss/centos7/goss.yaml validate
+out='Error: yaml: unmarshal errors:
  line 189: cannot unmarshal !!bool `true` into map[string]interface {}'
```

See my other comment on where I think the mistake is.

integration-tests/goss/goss-shared.yaml
```diff
@@ -177,3 +177,6 @@ matching:
         - have-key-with-value:
             foo: bar
         - have-key: baz
+env-var:
+  GOSS_TEST: true
```
aelsabbahy: I think this needs to be `GOSS_TEST: "true"` to ensure "true" is a string not a boolean value.

timbirk:
This is of interest to me,

One consideration, Environment variables depend heavily on the process or user that they get passed to. Is there some way to validate that for example my Java app has certain env variables set or not?

Should that be part of my applications healthcheck or something goss could check?

berney:
@timbirk in linux you could check `/proc/$pid/environ`, where `$pid` is the process ID of the process you want to check the environment variables of. The file contains them separated by NULLs (\x00).

aelsabbahy:
Seems this PR has gone stale and never passed testing. Closing.

PR #327, opened 2018-03-07
Add basic auth for http resource

Ntr0:
This patch will enable optional basic auth for the http resource. Therefore
it introduces two new parameters: username and password. If any of both is
set, the request will be performed using HTTP Basic Auth.

And, thanks for goss. Its awesome.


aelsabbahy:
Sorry for the delay in reviewing this.

Can you add a test for this. Thinking using something like: https://httpbin.org

Ntr0:
@aelsabbahy Can you give me a hint why the ci build fails? It seems that this check should pass: Addr: tcp://localhost:80: reachable: true  But I cannot see how I this is related to my change?

aelsabbahy:
The assertion output in the `test.sh` need to be updated to reflect the new test count.
https://github.com/aelsabbahy/goss/blob/master/integration-tests/test.sh#L46-L49

aelsabbahy:
This is awesome, thanks for this!

PR #328, opened 2018-03-12
[dgoss] fix info redirected to stderr

lutangar:
Hey there,

When running `dgoss` I noticed `info` was redirected to `stderr`, I don't think it was the intended behavior, but I might have missed something.
Let me know.

Anyway, thank you for this great tool!

Cheers.


aelsabbahy:
It was added in a PR a little while back, here's the rational for it: https://github.com/aelsabbahy/goss/pull/291

Issue #319, opened 2018-02-23
Service systemd implementation should not fallback to use sysvinit one


guillemj:
The current service_systemd.go implementation is falling back to use the sysvinit one (service_init.go) when the commands give an error exit core. This is wrong, because systemd should have either generated a service file for those sysvinit init scripts, or it or the admin might have disabled or masked those services, so falling back will give the wrong result. But it will definitely not pay attention to those sysvinit services outside of its compat support anyway, so either systemctl knows about it, or for all practical purposes it does not matter at all.

The case in point here for us, was the motd service, which is masked on Debian, so systemctl is-enabled motd will return an error code, but while having the Debian initscripts package installed, there will be sysvinit enabled service present, so a check expecting the service to be disabled and not running will fail with an unexpected result when it should have succeeded.

taurus-forever:
@aelsabbahy Is it possible to pay attention to the current pull request?
At the moment we have stuck on 0.3.5+patch from pull request,
but we would like to upgrade 0.3.7+ one day.

Our problem is simple: goss doesn't behave well on systemd based servers
where some services are disabled but they are also provide sysvinit scripts.
Race condition happens when we disable/mask such service in
systemd using the standard systemd approaches. In this case goss calls sysvinit scripts.

We need a way to disable fallback to sysvinit if service is masked/disabled in systemd.
The proposed pull request just removes fallback as logically is it wrong, IMHO.
Maybe you are worrying for backward compatibility here, in this case maybe
we can have agree on some goss configuration option which will keep current defaults
but allows us to disable fallback? Tnx!

aelsabbahy:
Sorry for missing this in the last release. I merged the PR into master. Can you confirm that master looks good.

Once we sort out #446 I can cut a new release.

aelsabbahy:
Closing since this is merged/released.

Issue #109, opened 2016-07-20
debian jessie systemd is-enabled for service without unit file


brechsteiner:
The service enabled function on services without systemd unit files is not working correct.

OS: Debian 8.5
goss: v0.1.8

root@test01-01:~# service --status-all | grep ntp
 [ + ]  ntp
root@test01-01:~# goss add service ntp
Adding Service to './goss.yaml':

ntp:
  enabled: false
  running: true

aelsabbahy:
Hmm.. so goss at the start of a run determines if a system is using systemd, upstart, or init and does its tests based on that detection.

So Debian is using both systemd and init side by side? I'm assuming `systemctl is-enabled ntp` returns false, right?

brechsteiner:
A few service has already a new systemd unit file, but the most of it has just the old init files. 

root@test01-01:~#  systemctl is-enabled ntp
Failed to get unit file state for ntp.service: No such file or directory
root@test01-01:~# echo $?
1

philpep:
Yes, this should be fixed in next debian release: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=760616

As a workaround I fallback to SysV in testinfra: https://github.com/philpep/testinfra/blob/master/testinfra/modules/service.py#L103

aelsabbahy:
@philpep Yup, that's what I had to do for upstart, the fix for this will be very similar:
* https://github.com/aelsabbahy/goss/blob/master/system/service_upstart.go#L51

BTW, testinfra is awesome. That and serverspec were a great inspiration when I started writing goss :)

aelsabbahy:
Fixed in [v0.1.9](https://github.com/aelsabbahy/goss/releases/tag/v0.1.9)

brechsteiner:
cool, thanks for your fast reactions. Works great

PR #343, opened 2018-04-11
Fix systemd service implementation to not fallback to sysv (#319)

guillemj:
The current service_systemd.go implementation is falling back to use the
sysvinit one (service_init.go) when the commands give an error exit code.
This is wrong, because systemd should have either generated a service
file for those sysvinit init scripts, or itself or the admin might have
disabled or masked those services, so falling back will give the wrong
result. Because it will definitely not pay attention to those sysvinit
services outside of its compat support anyway, so either systemctl
knows about them, or for all practical purposes it does not matter at
all.

### DIFF
--- a/system/service_systemd.go
+++ b/system/service_systemd.go
@@ -30,11 +30,6 @@ func (s *ServiceSystemd) Exists() (bool, error) {
 	if strings.Contains(cmd.Stdout.String(), fmt.Sprintf("%s.service", s.service)) {
 		return true, cmd.Err
 	}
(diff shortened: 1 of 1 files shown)


aelsabbahy:
The history of that change is the following:
* Issue: https://github.com/aelsabbahy/goss/issues/109
* related PR: https://github.com/aelsabbahy/goss/pull/110

Older debian needs this fallback to work properly.

Feel free to reopen this ticket if there's a way to have your changes while not breaking the issue highlighted above.

PR #344, opened 2018-04-12
Allow setting of Source IP for Addr Test

CBeerta:
This Change allows the Source IP with which the Addr test is run to be
set.


resource/addr.go
```diff
@@ -6,11 +6,12 @@ import (
...
 type Addr struct {
-	Title     string  `json:"title,omitempty" yaml:"title,omitempty"`
-	Meta      meta    `json:"meta,omitempty" yaml:"meta,omitempty"`
-	Address   string  `json:"-" yaml:"-"`
-	Reachable matcher `json:"reachable" yaml:"reachable"`
-	Timeout   int     `json:"timeout" yaml:"timeout"`
+	Title        string  `json:"title,omitempty" yaml:"title,omitempty"`
+	Meta         meta    `json:"meta,omitempty" yaml:"meta,omitempty"`
+	Address      string  `json:"-" yaml:"-"`
+	LocalAddress string  `json:"localaddress" yaml:"localaddress"`
```
aelsabbahy: Should have an "omitempty" for this.

PR #346, opened 2018-04-17
Add ability to loop n times in a template.

### LINKED ISSUE #345: Support for asserting against systemd template units?
I'm setting up some [buildkite agents](https://buildkite.com/docs/agent]). I'm running several instances per host, via their documentation.

I've got 8 running:

```
$ systemctl list-units | grep buildkite
buildkite-agent@1.service                                                                   loaded active running   Buildkite Agent
buildkite-agent@2.service                                                                   loaded active running   Buildkite Agent
buildkite-agent@3.service                                                                   loaded active running   Buildkite Agent
buildkite-agent@4.service                                                                   loaded active running   Buildkite Agent
buildkite-agent@5.service                                                                   loaded active running   Buildkite Agent
buildkite-agent@6.service                                                                   loaded active running   Buildkite Agent
buildkite-agent@7.service                                                                   loaded active running   Buildkite Agent
buildkite-agent@8.service                                                                   loaded active running   Buildkite Agent
system-buildkite\x2dagent.slice                                                             loaded active active    system-buildkite\x2dagent.slice
```

I've set up assertions as this, and they pass:

```yaml
service:
  buildkite-agent:
    enabled: false
    running: false
  buildkite-agent@1: # I'd like to ideally avoid duplicating this 8 times where only the index differs, since that will mean adjusting my tests if I change my mind about the number I want running. So I'd prefer this to be config/parameter-driven.
    enabled: true
    running: true
process:
  buildkite-agent:
    running: true
```

```shell
$ goss v -f tap
1..5
ok 1 - Process: buildkite-agent: running: matches expectation: [true]
ok 2 - Service: buildkite-agent@1: enabled: matches expectation: [true]
ok 3 - Service: buildkite-agent@1: running: matches expectation: [true]
ok 4 - Service: buildkite-agent: enabled: matches expectation: [false]
ok 5 - Service: buildkite-agent: running: matches expectation: [false]
```

I'd like to be able to make assertions like (as in, this is a sketch, I don't know whether it's a good one):

```yaml
process:
  buildkite-agent:
    running: true
    count: 8

service:
  buildkite-agent:
    enabled: true # when template, then I would expect this to pass only if all n are running
    running: true
    templated: true
    count: 8
```

I'd like to be able to supply the `count` as a parameter. Failing that, I'd like to template it in, somehow - I'm probably happy to do that externally rather than ask to complicate goss by building in templating, but I also don't know whether that's a reasonable thing to suggest or not.

What do you think?

petemounce:
Attempt to fix #345. Incomplete.

This is based on https://stackoverflow.com/questions/28917530/golang-how-to-create-loop-function-using-html-template-package

I'm trying to enable

```yaml
{{range $idx $item := loop 8}}
service:
  foo@{{$idx}}:
    running: true
    enabled: true
{{end}}
```
... and generate 8 foo service assertions, each with 2 checks. I then want to supply that `8` from a vars file or an environment variable, but I think support for that exists already.

Happy to write tests but I think I need a steer for where to do so and how to assert?

### DIFF
--- a/template.go
+++ b/template.go
@@ -33,6 +33,10 @@ func getEnv(key string, def ...string) string {
 	return os.Getenv(key)
 }
 
+func loop(iterations int) []struct{} {
+    return make([]struct{}, iterations)
+}
+
 func regexMatch(re, s string) (bool, error) {
 	compiled, err := regexp.Compile(re)
 	if err != nil {
@@ -46,6 +50,7 @@ var funcMap = map[string]interface{}{
 	"mkSlice":    mkSlice,
 	"readFile":   readFile,
 	"getEnv":     getEnv,
+	"loop":       loop,
 	"regexMatch": regexMatch,
 }
 


aelsabbahy:
I'm fine with merging this if it's documented, not all the current template functions have tests.

If you find a good way to unit test this in go, that would be ideal, but I won't block the PR on that.

petemounce:
@aelsabbahy thanks for the feedback - I'll try to figure out whether I can reasonably test it, and I'll definitely document it.

aelsabbahy:
Is this abandoned or something you're still interested in?

For what it's worth, I'm going to create a ticket to add [sprig](https://github.com/Masterminds/sprig) support into goss.

I think the `until` function will take care of this need:
https://masterminds.github.io/sprig/integer_slice.html

aelsabbahy:
Closing this in favor for #553 which should be in goss v0.4.0

petemounce:
Oh; that should cover it nicely.

PR #348, opened 2018-04-19
If no matched files were found, we throw an exception.

### LINKED ISSUE #347: Include a not existed file should get an error.

axot:
This PR will fix #347.


aelsabbahy:
Much clearer error, thanks!

Issue #315, opened 2018-02-02
Bug: HTTP check fails when http content is too large with 'body' attribute


evancarter-iex:
If you have something like this:
```
http:
  http://api.magicthegathering.io/v1/cards?page=311:
    # required attributes
    status: 200
    timeout: 1000
    body: [colors] # Check http response content for these patterns
```
Goss will return:

``` Body: Error: bufio.Scanner: token too long```

and thus "failing"

aelsabbahy:
Fixed in latest release

PR #351, opened 2018-04-29
Increase buffer size

pysysops:
This "fixes" #315 where big lines of test (like single line JSON) are not handled well by the default buffer size. This increases it from the default to 512K which _should_ be enouch for most situations although this might warrant review / improvement in future.


stell-aelsabbahy:
Let's bump this to 1024, and I'll merge it.

timbirk:
Buffer bumped to 1024.

aelsabbahy:
Thanks!

Issue #220, opened 2017-04-05
Clarify kernel-param value as string


aelsabbahy:
Clarify the value is a string, to avoid users doing something like this.
```
KernelParam: net.ipv4.ip_forward: value:
Expected
    <string>: 0
to equal
    <int>: 0
```

walidshaari:
encourge people to do : goss  add  to find the syntax too, e.g. goss  add KernelParam goss  add net.ipv4.ip_forward

zylad:
The PR lookss good. Any chance to get it merged?

aelsabbahy:
Need to just update the docs for now.

PR #352, opened 2018-04-29
Handle integers and boolean kernel params

pysysops:
This PR addresses the issue raised in #220 although we would hope that users fully understand types in YAML it's not always the case. As a defensive measure we could cast the user specified value to a string (expected type) before validating. This will allow some flexibility to the end user (I forget to add " to string values myself) giving a smoother experience.

### DIFF
--- a/resource/kernel_param.go
+++ b/resource/kernel_param.go
@@ -1,6 +1,8 @@
 package resource
 
 import (
+	"strconv"
+
 	"github.com/aelsabbahy/goss/system"
 	"github.com/aelsabbahy/goss/util"
 )
@@ -24,7 +26,18 @@ func (a *KernelParam) Validate(sys *system.System) []TestResult {
 	sysKernelParam := sys.NewKernelParam(a.Key, sys, util.Config{})
 
 	var results []TestResult
-	results = append(results, ValidateValue(a, "value", a.Value, sysKernelParam.Value, skip))
+	// Cast to string
+	var value matcher
+	switch v := a.Value.(type) {
+	case int:
+		value = strconv.Itoa(v)
+	case bool:
+		value = strconv.FormatBool(v)
+	default:
+		value = v
+	}
+
+	results = append(results, ValidateValue(a, "value", value, sysKernelParam.Value, skip))
 	return results
 }
 


zylad:
Any chance to get this merged, @aelsabbahy?

stell-aelsabbahy:
I'm mixed on this feature. On the one hand, I like that it simplifies the usage and possibly lowers the risk of erros on the other hand I have two concerns:
1. Inconsistent behavior, if we type-cast kernel-params/Value, we should probably type cast all other resource types.
2. Is there a risk of magic/confusion when the wrong type is input into YAML? For example, a `0644` turning into a `"644"`? (Haven't thought much about this one and whether there's a real world impact)

timbirk:
Valid thoughts @aelsabbahy a risk of magic is indeed always a threat... As smoothly as I like to make a users experience perhaps it hides away the fact that strings are expected.

I've found some people don't "get" that YAML is typed and things sometimes require escaping. Perhaps a better route to take is for me to write up a small "newbies guide to YAML" section to cover escaping (or preferably using single quotes) and some of the things to be aware of with various resources like using `'0644'` not `0644`, `'true'` over `true` and working with the `kernel-param` values (strings).

colinbruner:
The problem that I'm personally having with this is that Goss' rendering, from my --var files, does not seem to respect YAML types. For example, I have a variable file that looks like this:
```
kernelparams:
  - param: net.core.rmem_default
    value: "1048575"
```

A "common" kernel-params test file that looks like this:
```
{{ if .Vars.kernelparams }}
kernel-param:
{{ range .Vars.kernelparams }}
  {{ .param }}:
    value: {{ .value }}
{{ end }}
{{ end }}
```

Which is rendered into my goss.yaml file looking like this:
```
kernel-param:
  net.core.rmem_default:
    value: 1048575
```

But ends up erroring out like this:
```
KernelParam: net.core.rmem_default: value:
Expected
    <string>: 1048575
to equal
    <int>: 1048575
```

Does anyone have any advice...? My process looks something like this:

1. Template all JINJA .tmpl files from CMDB.
   - Creates a "$testname.test.yaml" file
2. Generate additional variable files from shell scripts.
   - Creates "$testname.vars.yaml" file
3. Create a combined variable file required by host.
   - Creates "$HOSTNAME.vars.yaml" file
4. Create a gossfile "importing" other test files.
   - Creates "$HOSTNAME.tests.yaml" file
4. Render variables and tests into a goss.yaml file for testing.
   - `goss -g $HOSTNAME.tests.yaml --vars $HOSTNAME.vars.yaml render > goss.yaml`

Should I just not both rendering kernel-param values? I could possibly add them statically to my goss.yaml file after I use the gossfile resource to generate my goss.yaml file?

Issue #353, opened 2018-05-12
dgoss is not properly labeling 'goss' binary


tgeens:
When running 'dgoss' on a SELinux enabled system, you run into issues:

```
$ dgoss run -it --rm ubuntu
INFO: Starting docker container
INFO: Container ID: 087e3b67
INFO: Sleeping for 0.2
INFO: Running Tests
sh: 1: /goss/goss: Permission denied
INFO: Deleting container
```

In the audit logs I can find this entry:

```
type=AVC msg=audit(1526112518.888:986): avc:  denied  { execute } for  pid=7384 comm="sh" name="goss" dev="tmpfs" ino=1749934 scontext=system_u:system_r:container_t:s0:c397,c432 tcontext=unconfined_u:object_r:user_tmp_t:s0 tclass=file permissive=0
```

To me, this looks like the `goss` binary is mounted into the container, but not properly labeled.

From the `docker run` [documentation](https://docs.docker.com/engine/reference/commandline/run/#mount-volumes-from-container---volumes-from):

> Labeling systems like SELinux require that proper labels are placed on volume content mounted into a container. Without a label, the security system might prevent the processes running inside the container from using the content. By default, Docker does not change the labels set by the OS.
>
> To change the label in the container context, you can add either of two suffixes :z or :Z to the volume mount. These suffixes tell Docker to relabel file objects on the shared volumes. The z option tells Docker that two containers share the volume content. As a result, Docker labels the content with a shared content label. Shared volume labels allow all containers to read/write content. The Z option tells Docker to label the content with a private unshared label. Only the current container can use a private volume.

PR #354, opened 2018-05-13
dgoss SELinux compat - added volume mount flag

tgeens:
Labeling systems like SELinux require that proper labels are placed on volume content mounted into a container. Without this label, SELinux prevents goss from running inside a container.

Docker can apply the proper labels to a mounted volume, when the :z mount flag is applied.

This fixes issue #353


aelsabbahy:
Nice catch, thanks!

PR #358, opened 2018-05-27
Json pretty format-option

timbirk:
Hey 👋 

This is probably suitable for a minor release as it changes the default output of the `json` formatter to single-line JSON and adds a `--format-option pretty` for pretty printing JSON.

It allows us to get rid of the mostly duplicated `json_oneline` output and make use of the new `--format-options`. 

Any discussion always appreciated :) 

Thanks,

Tim.


berney:
Can pip into `jq` to get pretty with color syntax and can filter as well.

PR #359, opened 2018-06-08
Redirect docker output to stderr

burkl:
Command `docker top $id` and eventually `docker logs $id` output to stdout. That means, that one gets an invalid XML in the case of defining junit as the output format.


aelsabbahy:
Fixes #412

aelsabbahy:
Thank you for contributing to goss!

PR #370, opened 2018-08-03
Fix GetMounts, use nil filter to get old behaviour

berney:
GetMounts() behaviour changed, now expects a filter parameter, use nil to get old behaviour.

### DIFF
--- a/system/mount.go
+++ b/system/mount.go
@@ -88,7 +88,7 @@ func (m *DefMount) Filesystem() (string, error) {
 }
 
 func getMount(mountpoint string) (*mount.Info, error) {
-	entries, err := mount.GetMounts()
+	entries, err := mount.GetMounts(nil)
 	if err != nil {
 		return nil, err
 	}


berney:
travis-ci has failed due to unable to go get golint.

berney:
Travis-CI failed for #371 because of this change, the old version of the mount library code is being used with the old GetMounts().

sshipway:
This one (and Travis) really needs to be fixed, it is breaking all the Travis checks in the later pull requests.

Is this project hibernating?  Or abandoned?  It would be a shame if the latter as it is clearly popular.

aelsabbahy:
@sshipway I've merged in a change to disable the `dnstest.io` and master is now green.

I'm not sure I understand the `GetMounts()` issue. If dependencies are installed/locked using a package management tool this should not be an issue.

I'm fine with accepting a PR that updates this to using `GetMounts(nil)` but only if it also updates the dependency version to a newer one that uses the new format.

Currently goss is using `go dep`, previously (when this PR was created) it was using Glide, and soon it will hopefully be using go mod #437

sshipway:
I can't work out how to make 'go dep' work, and it was hard enough working out how to make go source compile at all until I found 'go get' and softlinked to the source from elsewhere.  The GetMounts that 'go get' installed seems to need this parameter that is not in the code, but as yet I have no idea how you configure or use dependencies.  If I ever find out I'll make a separate pull req!  Go seems to be a relatively easy language to read/write but very hard to set up a compile environment...

PR #371, opened 2018-08-03
Add support for OpenRC runlevels

berney:
Alpine Linux (and Gentoo) use OpenRC. Goss was hard-coded to expect a service to be at the sysinit runlevel, if the service was at a different runlevel the service check would erroneously fail.

Behaviour prior to this PR: -

```
alpine38:/tmp# service sshd status
 * status: started

alpine38:/tmp# ./goss-0.3.6-linux-amd64 -g /tmp/goss/goss.yaml v
..........F.

Failures/Skipped:

Service: sshd: enabled:
Expected
    <bool>: false
to equal
    <bool>: true

Total Duration: 0.012s
Count: 12, Failed: 1, Skipped: 0
```

With this PR the service block now supports an optional `runlevel` attribute, which defaults to `sysinit` if omitted maintaining the old behaviour.

Now it is possible to install openssh-server in Alpine linux and properly check the service is running with the following config: -

```
service:
  sshd:
    enabled: true
    running: true
    runlevel: default
```

Additionally you might have:

```
port:
  tcp:22:
    listening: true
    ip:
    - 0.0.0.0
  tcp6:22:
    listening: true
    ip:
    - '::'
user:
  sshd:
    exists: true
    groups:
    - sshd
    home: /dev/null
    shell: /sbin/nologin
group:
  sshd:
    exists: true
process:
  sshd:
    running: true
```

Post this PR:

```
alpine38:/tmp# ~/goss-berne2d -g /tmp/goss/goss.yaml v
runlevel is default
............

Total Duration: 0.012s
Count: 12, Failed: 0, Skipped: 0
```

Note: I'm quite new to reading and writing golang. I've tested the changes and the code works for me. If I have used bad constructs or poor programming please correct it or help me correct it.

Ideally I'd have written test cases but I don't know how to do that yet.

### DIFF
--- a/integration-tests/test.sh
+++ b/integration-tests/test.sh
@@ -45,6 +45,8 @@ echo "$out"
 
 if [[ $os == "arch" ]]; then
     egrep -q 'Count: 84, Failed: 0, Skipped: 3' <<<"$out"
+elif [[ $os == "alpine3" ]]; then
+    egrep -q 'Count: 103, Failed: 0, Skipped: 5' <<<"$out"
 else
     egrep -q 'Count: 101, Failed: 0, Skipped: 5' <<<"$out"
 fi
--- a/util/config.go
+++ b/util/config.go
@@ -18,6 +18,7 @@ type Config struct {
 	Username          string
 	Password          string
 	LocalAddress      string
+	RunLevel          string `default:"sysinit"`
 }
 
 type OutputConfig struct {
--- a/docs/manual.md
+++ b/docs/manual.md
@@ -757,8 +757,13 @@ service:
     # required attributes
     enabled: true
     running: true
+    # Optional attributes - Alpine systems
+    runlevel: default
+    # Optional attributes
     skip: false
 ```
+
+`runlevel` will default to `sysinit` if omitted.
 
 **NOTE:** this will **not** automatically check if the process is alive, it will check the status from `systemd`/`upstart`/`init`.
 
--- a/integration-tests/Dockerfile_alpine3
+++ b/integration-tests/Dockerfile_alpine3
@@ -3,7 +3,8 @@ MAINTAINER Ahmed
 
 # install apache2 and remove un-needed services
 RUN apk update && \
-  apk add openrc apache2 bash ca-certificates && \
+  apk add openrc apache2 bash ca-certificates openssh && \
   rc-update add apache2 && \
+  rc-update add sshd default && \
   rm -rf /etc/init.d/networking /etc/init.d/hwdrivers /var/cache/apk/* /tmp/*
 RUN mkfifo /pipe
--- a/integration-tests/goss/alpine3/goss.yaml
+++ b/integration-tests/goss/alpine3/goss.yaml
@@ -3,6 +3,13 @@ service:
   autofs:
     enabled: false
     running: false
+  apache2:
+    enabled: true
+    running: true
+  sshd:
+    enabled: true
+    running: true
+    runlevel: default
 user:
   apache:
     exists: true
--- a/resource/service.go
+++ b/resource/service.go
@@ -12,7 +12,7 @@ type Service struct {
 	Enabled  matcher `json:"enabled" yaml:"enabled"`
 	Running  matcher `json:"running" yaml:"running"`
 	Skip     bool    `json:"skip,omitempty" yaml:"skip,omitempty"`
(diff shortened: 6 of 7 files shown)


berney:
Hmm on my system its pulling in the newer version of mount so it wants a filter (I raised PR #370 for this) but in travis-ci it's using the old version and complains of too many arguments. I don't understand glide and golang enough to understand why the dependency is upgraded on my system but not in travis-ci. I'm using Ubuntu 16.04 LTS , go v1.10.3, glide v0.13.1.

berney:
I rebased onto master to remove the GetMounts change (#370). I also made the new `runlevel` attribute omitted if it is blank, which makes all the integration tests pass.

I fixed the other make test errors, updating the Dockerfile for Wheezy to fix apache2.2 package error and the corresponding goss files.

There is a warning about the hash in the  `glide.lock` file not matching the `glide.yaml` file. This appears to be from preexisting changes where the `glide.lock` file was not updated after changes to the `glide.yaml` file. This PR hasn't touched those files.

travis-ci test is now passing.

berney:
I've rebased to master.

I've updated the documentation to include the optional `runlevel` attribute and its default if omitted.

I've updated the integration tests. I modified the alpine3 Dockerfile, the MD5 sum won't match anymore. I have not updated the MD5 sum, since @aelsabbahy will need to rebuild the docker images.

integration-tests/test.sh
```diff
@@ -45,6 +45,8 @@ echo "$out"
 
 if [[ $os == "arch" ]]; then
     egrep -q 'Count: 84, Failed: 0, Skipped: 3' <<<"$out"
+elif [[ $os == "alpine3" ]]; then
```
aelsabbahy: Can the tests be re-written in a way that avoids this  if condition?

Want to avoid maintaining too many separate test counts.

util/config.go
```diff
@@ -18,6 +18,7 @@ type Config struct {
 	Username          string
 	Password          string
 	LocalAddress      string
+	RunLevel          string `default:"sysinit"`
```
aelsabbahy: Curious, why use fields here?

I'd like to keep the patterns consistent as much as possible. As in, I'd like for all the config default values to be handled the same way.

integration-tests/test.sh
berney: This PR adds an OpenRC (Alpine) specific attribute so I added tests only to the Alpine goss file.

I think it might be better to refactor the integration tests to move the Dockerfiles under the OS directories and split `test.sh` to have an OS specific part in the OS directory that will include the expected counts of pass/failed/skipped tests.

util/config.go
berney: It was the way I worked out how to add a default.

Which other code has an attribute with a default string value value that I can use an example of how to do it?

aelsabbahy: http tuneout an example of a default.

cmd/goss/goss.go flags & goss/add.go (AddResources)

berney: I'm not sure I understand. Do you mean I should add code like:

```
if c.RunLevel == "" {
  c.RunLevel = "sysinit"
}
```

berney:
@aelsabbahy Can this PR be re-opened? I've updated my branch to not use the struct tag `default` and reflection, to be consistent with the rest of goss.

util/config.go
berney: @aelsabbahy I've updated this code to no longer use reflection based off the default tag on the struct.
I hope this resolves this objection.

integration-tests/test.sh
berney: @aelsabbahy I hope based on my comment above that the alpine3 `if` condition is OK.

If you still object, what do you think of my idea to refactor integration tests to take expected results from file in OS directory? If you approve I could open a separate PR for this work.

aelsabbahy:
Github won't let me click the  "reopen and comment" because "the *** branch was forced-pushed or recreated"

Can you create a new PR and reference it here in a comment.

Sorry for the inconvenience

berney:
> Github won't let me click the "reopen and comment" because "the *** branch was forced-pushed or recreated"
> 
> Can you create a new PR and reference it here in a comment.
> 
> Sorry for the inconvenience

Ah sorry about that, after rebasing I did a force-push. I didn't know about this interaction with PRs.

I've opened a new PR https://github.com/aelsabbahy/goss/pull/650 to continue this.
Thanks

PR #380, opened 2018-08-29
Attempt to add cross-platform builds

petemounce:
Also 
* wrote down how I tried to get set up for local development on macOS (failed to get glide to run, so far, it just panics. I have golang 1.10).
* worked around the fact that `owner`, `group` or `mode` are not applicable to Windows filesystems.


petemounce:
Superceded by #585.

PR #382, opened 2018-09-03
Fix debian wheezy CI

petemounce:
according to https://packages.debian.org/search?keywords=apache2 there was a security fix, so use later package

I've seen this in a couple of open PRs, so here's the fix in isolation if that's easier to merge.


davidcollom:
@aelsabbahy Any chance we could get this merged? seems that this would fix multiple PR's failing tests.

petemounce:
Thanks for merging!

PR #384, opened 2018-09-05
Add ToUpper and ToLower to Template funcMap

### LINKED ISSUE #383: Add ToUpper and ToLower support for Template Pipelines
I have a series of URLs in my Docker image that I want to test and I already have separate tests written for mixed-case URLs. It would be nice to allow Template pipelines to change the case of a Template string to force (e.g.) lowercase URL testing to try to maintain separation of concerns while not forcing the caller to lowercase the URLs to test.

For example:

```
http:
  # mixed-case URL tests
  # <snip>

  # URL tests from --vars
  {{range .Vars.urls}}
    http://localhost/{{. | ToLower}}
  {{end}}
```

Currently, that syntax gives an error:

```
2018/09/05 12:33:07 template: test:34: function "ToLower" not defined
```

mattrice:
Add basic ToUpper and ToLower support for Template pipelines by injecting functions from strings

Fixes #383

### DIFF
--- a/docs/manual.md
+++ b/docs/manual.md
@@ -848,6 +848,8 @@ Available functions beyond text/template [built-in functions](https://golang.org
 * `readFile "fileName"` - Reads file content into a string, trims whitespace. Useful when a file contains a token.
   * **NOTE:** Goss will error out during during the parsing phase if the file does not exist, no tests will be executed.
 * `regexMatch "(some)?reg[eE]xp"` - Tests the piped input against the regular expression argument.
+* `ToLower` - Changes piped input to lowercase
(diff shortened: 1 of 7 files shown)


mattrice:
I have not been working with Go long, so I am unsure how this functionality could be tested.

aelsabbahy:
I think changing: https://github.com/aelsabbahy/goss/blob/master/integration-tests/goss/goss-shared.yaml

To `/etc/PAssWord` and using ToLower would be sufficient.

aelsabbahy:
Thank you

PR #394, opened 2018-10-23
Bump Go versions

HaraldNordgren:


stell-aelsabbahy:
I think 1.11.x is sufficient. No need to build using multiple revisions. Aside from that, LGTM!

PR #396, opened 2018-10-26
remove double entry in doc

gimler:


aelsabbahy:
Since it's a required attribute the duplication is necessary. Otherwise, copying/pasting that code wouldn't work.

PR #400, opened 2018-11-06
Fix resource name

dud225:


aelsabbahy:
When doing an add, the command is:
`goss add goss ..`

The entry in the yaml file is `gossfile`

PR #403, opened 2018-11-07
Add CI for macOS and Windows

petemounce:
Travis supports Windows now. Let's try that out... I'll add to-do items as I go. I'll be aiming at

* keeping builds green on Linux
* adding golang 1.10 and 1.11 (since #394 did, and why not)
* making changes to enable the build to pass on the 2 new OS'
  * that might be "skip tests with a TODO" at this stage - want to limit the scope of the PR, because having passing CI is a good foundation for then making tests pass, which could happen in parallel in a child branch.

### OSX

* [x] no `docker` travis service
  * it _probably_ doesn't make sense to do `docker` things on OSX, because one cannot dockerise macOS apps.
* [ ] run integration tests on macOS, but not inside docker.
  * I think it's _probably_ reasonable to run the integration tests on the naked host. While that opens up environment-drift possibility, the hosts are still pretty predictable, and there's a changelog (https://docs.travis-ci.com/user/build-environment-updates/#macos-build-environment-updates)

### Windows

* [x] syntax of find command is wrong.
  * Windows `find` is definitely not GNU find. Windows env gets git-bash, a stripped down msys environment, which doesn't appear to have find, or doesn't have it in PATH.
  * Changed the command to use `git ls-files...` to list all the go non-test, non-vendor files.
* [ ] run integration tests on Windows, but not inside docker


petemounce:
@aelsabbahy I'd be happy to split to individual OS'. I'll continue here to get to green, then split; that will allow me to do that in a way that allows merge conflicts to be avoided.

I have a few questions:

1. "I don't want to merge CI in until the code changes for that OS are part of Goss." - I don't think I understood that. Do you mean; you don't want to merge CI changes until goss works end-to-end for the OS being added? If so - please help me understand what your criteria are, since
    * macOS tests-in-docker are not going to be possible, since macOS cannot be containerised, one can only run linux-containers on a macOS host
        * I propose to run the integration tests on the Travis node directly - however, this will almost certainly require some changes to the tests implementation, or scope, depending on as-yet-unknown platform-differences. Does that sound like an acceptable approach?
    * Windows travis hosts do not (yet?) support docker
        * I propose same approach as macOS, run on the node directly - same caveat applies
1. Please would you elaborate on over-complicate? That will allow contributors to understand the constraints to work within. Would it be possible to write a CONTRIBUTING.md for those guidelines?
1. How you foresee supporting Windows / macOS in the future? Have contributors do that, end-to-end? Will that involve widening the pool of people with commit-rights?

petemounce:
Superceded by #585.

PR #410, opened 2018-11-26
Fix build error, fixes #402

### LINKED ISSUE #402: Not enough arguments in call to mounts.GetMounts
Apologies if I'm misunderstanding something. I'm trying to build goss on my Mac using "go get" and I get this error.

github.com/aelsabbahy/goss/system/mount.go:91:33: not enough arguments in call to mount.GetMounts
	have ()
	want (mount.FilterFunc)

According to this, https://github.com/moby/moby/blob/master/pkg/mount/mount.go#L51 if there are not any arguments you should pass in nil.

SimonBaeumer:
Build failed due to missing filter function for mount points. I added an empty filter function to solve it. Fixes #402


.travis.yml
```diff
@@ -10,11 +10,8 @@ services:
 - docker
 
 before_install:
-- curl -L https://github.com/Masterminds/glide/releases/download/0.10.2/glide-0.10.2-linux-amd64.zip
-  > glide.zip
-- unzip glide.zip
-- export PATH="$PATH:$PWD/linux-amd64"
 - go get -u github.com/golang/lint/golint
+- go get -u github.com/golang/dep/cmd/dep
```
petemounce:
```suggestion
DEP_RELEASE_TAG=0.5.0 curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh
```

(That's dep's documented installation method - the script supports cross-platform)

Possible to pin to a particular version? I think dep has had breaking changes before.

Makefile
```diff
@@ -93,7 +93,7 @@ test-all: lint test test-int
 
 deps:
 	$(info INFO: Starting build $@)
-	glide install
+	dep ensure
```
petemounce:
```suggestion
./dep ensure
```

(Only if you go with the previous suggestion though)

novendor.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+set -eu
```
petemounce:
```suggestion
set -euo pipefail
```

To trap failures inside pipes.

novendor.sh
```diff
@@ -0,0 +1,20 @@
...
+# Bash replacement for glide novendor command
+
+DIRS=$(ls -ld */ . | awk {'print $9'} | grep -v vendor)
+
+for DIR in ${DIRS}; do
+    GOFILES=$(find ${DIR} -name "*.go")
```
petemounce: Use `git ls-files`? Its behaviour will not vary cross-platform.

See also https://github.com/aelsabbahy/goss/pull/403/files#diff-b67911656ef5d18c4ae36cb6741b7965R7

.travis.yml
SimonBaeumer: Yes, very good hint!

Makefile
SimonBaeumer: :+1:

novendor.sh
SimonBaeumer: :+1:

novendor.sh
SimonBaeumer: Implemented it, thank you for your feedback!

petemounce:
I'm a fan of vendoring dependencies in general because it's a very explicit way of dealing with them, and cuts away external-calls during builds. Where I am at present develops within a [monorepo](https://trunkbaseddevelopment.com/monorepos/) with [bazel](https://bazel.build) as our build tool; it works very well.

As long as there's a good process in place to maintain vendored dependencies close to their HEADs (renovatebot, dependabot both do this for free for OSS), that removes most of the fragility. To be honest, if a tool is used to manage the dependencies in the first place, it matters less whether they are vendored or on-demand; the important part is to keep control over sprawl and staleness.

petemounce:
Regards: review, will you push your changes where you have resolved comments?

SimonBaeumer:
Thanks for your explanation, some very useful tools I didn't know before :+1: 

I pushed my changes, forgot it yesterday. ;)

SimonBaeumer:
Is travis on vacation? :(

petemounce:
Try pushing an empty commit to re-trigger? `git commit --allow-empty`

SimonBaeumer:
Couldn't push `--allow-empty`. Now I squashed my last two commits and hopefully travis will work^^

SimonBaeumer:
Hui, passed! :)

petemounce:
LGTM!

aelsabbahy:
Thanks for this contribution, few notes:
1. You mention the travis build failing? I'm not seeing this error, how can I reproduce this? I branched off of master to trigger a build and it travis built it without any errors: https://travis-ci.org/aelsabbahy/goss/builds/469518749

2. I like the switch to `dep` from glide, but always ran into missing features that blocked me from migrating to it.. seems those are gone now.

3. I prefer not to have the `vendor/` directory committed to the repo (mostly due to the cons listed in the FAQ, primarily git history/diff being a mess). Seems there's no clear best practice on this: https://github.com/golang/dep/blob/master/docs/FAQ.md#should-i-commit-my-vendor-directory

petemounce:
Where I am, we commit vendor because we want reproducible builds. We use `.gitattributes` and github's linguist to suppress diffs in `vendor/` being expanded by default via

```gitattributes
# make github know about vendored code
# https://github.com/github/linguist#vendored-code
go/src/us.com/vendor linguist-vendored
**/vendor/** linguist-vendored
```

SimonBaeumer:
Thanks for your feedback @aelsabbahy!
The problem with travis was that it was not triggered to run a new build, so probably nothing we can fix.

Removed `vendor/` and squashed commits.  :+1:

aelsabbahy:
@petemounce few questions on that approach:
1. How do you suppress git diff on the cli?
2. How painful are merge conflicts in the `vendor/` directory
3. How does one validate (preferably in CI) that the gopkg.lock and the contents of `vendor/` match?

petemounce:
1. `git diff -- [directory]` to scope the diff to the given tree(s) or file(s)
1. Not painful at all; they do not occur, because the `vendor/` tree is functionally immutable. That is, it's a cache that the dependency manager owns, not the developers in the repo.
> dep wholly controls vendor, and will blow away any manual changes or additions made to it that deviate from the version of an upstream source dep selects.
1. `dep check` for ci. `dep ensure` to make changes to `vendor/` and `gopkg.lock` file based on local intentional changes to the `gopkg.toml`; https://golang.github.io/dep/docs/daily-dep.html#basics

aelsabbahy:
@SimonBaeumer LGTM, thanks! I'll merge this most likely tomorrow.

I meant the Travis issue that caused this whole glide -> dep migration and this PR, curious on what it was. Even if nothing is broken I still like the switch to dep.

@petemounce
We can continue the discussion in this PR or create an issue for the discussion. I see pros/cons to both approaches. I would want to experiment with the vendor idea and address some of the concerns I have prior to deciding to add it.

PR #415, opened 2019-01-08
Test the container health right after starting it

dud225:
Hello

PR #304 added additional output in case the container doesn't start. That's great, however this code path doesn't run if one uses the wait feature.
So this PR displays the logs when using this feature as well.


aelsabbahy:
Looks great, thank you for this.

Issue #420, opened 2019-01-29
goss exits with 0 when flags are bad


jim-raney-physiq:
```
[bash]$ goss -g goss.yaml validate -o verbose --color --sleep 5s --retry-tiemout 2m
Incorrect Usage: flag provided but not defined: -retry-tiemout

NAME:
   goss validate - Validate system

USAGE:
   goss validate [command options] [arguments...]

OPTIONS:
   --format value, -f value          Format to output in, valid options: [documentation json json_oneline junit nagios rspecish silent tap] (default: "rspecish") [$GOSS_FMT]
   --format-options value, -o value  Extra options passed to the formatter, valid options: [perfdata verbose] [$GOSS_FMT_OPTIONS]
   --color                           Force color on [$GOSS_COLOR]
   --no-color                        Force color off [$GOSS_NOCOLOR]
   --sleep value, -s value           Time to sleep between retries, only active when -r is set (default: 1s) [$GOSS_SLEEP]
   --retry-timeout value, -r value   Retry on failure so long as elapsed + sleep time is less than this (default: 0s) [$GOSS_RETRY_TIMEOUT]
   --max-concurrent value            Max number of tests to run concurrently (default: 50) [$GOSS_MAX_CONCURRENT]
   
[bash]$ echo $?
0
```

Would be better if it exited non-zero for tools that actively check error codes for test results.

PR #421, opened 2019-01-31
Handle invalid CLI flags

mkobit:
Previously, the flags were not handled so the application would
Looking at the documentation for https://github.com/urfave/cli , the error is propogated up through the `app.Run` call.
Now, the error is checked before fatally logging it.

closes https://github.com/aelsabbahy/goss/issues/420


aelsabbahy:
Nice catch, thanks!

PR #427, opened 2019-03-19
Add "dcgoss flavour" of dgoss

fbartels:
I am currently playing with goss and a setup coming from docker-compose. Looking through Github I came across https://github.com/telephoneorg/dcgoss

The Dockerfile in there did not build anymore and the website of the org does not work anymore so it seems this project is not maintained any longer. So I decided to import their latest version (which seems anyways be based on dgoss, although their repo is missing a license file) and merge it with changes from dgoss.

Additionally I let it run through shellcheck and fixed some warnings.


aelsabbahy:
Can you move this to a dcgoss folder and add a brief readme.md.

Doesn't have to be too detailed. The extras are really there for inspiration on ways to use goss.

fbartels:
Hi @aelsabbahy,

sorry for the update taking so long. I have now added a readme and moved dcgoss into its own subdir.

fbartels:
Looking into this again, should I also add this to the travis file? e.g. after https://github.com/aelsabbahy/goss/blob/master/.travis.yml#L28

aelsabbahy:
Looks great, and no need to put it in .travis.yml. I don't think I want to package the extras folder as part of goss. dgoss is an exception currently.. but that might not always be the case.

PR #434, opened 2019-04-05
Add new options to Command resource

sshipway:
This pull request does the following:
* Add 'exec' attribute to Command resource defaulting to hash key
* Add 'skip' attribute to most resources to allow them to be easily disabled when running a deepmerge of the yaml (e.g. in puppet config management tool)
* Documentation changes for the above
* Update of acceptance tests to reflect new exec attribute defaults

### DIFF
--- a/docs/manual.md
+++ b/docs/manual.md
@@ -432,18 +432,23 @@ Validates the exit-status and output of a command
 
 ```yaml
 command:
-  go version:
+  version:
     # required attributes
     exit-status: 0
+    # defaults to hash key
+    exec: "go version" 
     # optional attributes
     stdout:
     - go version go1.6 linux/amd64
     stderr: []
     timeout: 10000 # in milliseconds
+    skip: false
 ```
 
 `stdout` and `stderr` can be a string or [pattern](#patterns)
 
+The `exec` attribute is the command to run; this defaults to the name of
+the hash for backwards compatibility
 
 ### dns
 Validates that the provided address is resolvable and the addrs it resolves to.
@@ -535,6 +540,7 @@ file:
     # optional attributes
     filetype: symlink # file, symlink, directory
     linked-to: /usr/sbin/sendmail.sendmail
+    skip: false
 ```
 
 `contains` can be a string or a [pattern](#patterns)
@@ -560,6 +566,7 @@ group:
     exists: true
     # optional attributes
     gid: 65534
+    skip: false
 ```
 
 
@@ -578,6 +585,7 @@ http:
     body: [] # Check http response content for these patterns
     username: "" # username for basic auth
     password: "" # password for basic auth
+    skip: false
 ```
 
 
@@ -695,6 +703,7 @@ package:
     # optional attributes
     versions:
     - 2.2.15
+    skip: false
 ```
 
 **NOTE:** this check uses the `--package <format>` parameter passed on the command line.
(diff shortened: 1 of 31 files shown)


sshipway:
Looks like the travis build check is failing because it is using an outdated version of golang that expects no arguments to getMounts.  Part of this patch adds the (nil) parameter to the getMounts() call for compatibility.

aelsabbahy:
@sshipway the travis check uses `go dep` to install specific versions of the dependences. The correct use is `GetMounts()` for the version that's locked in the https://github.com/aelsabbahy/goss/blob/master/Gopkg.lock

Running `make deps` locally may resolve your local build issue.

sshipway:
Looks like I need to install 'godep' and set up a whole lot of softlinks to make it work.  I'll revert the GetMounts change in my branch and see if it then passes travis.

sshipway:
@aelsabbahy - It now works using the older docker/docker library and I have godep doing its thing, and it's passed Travis.  Go seems to be rather fond of github.  Thanks for the pointers, hope these improvements can make it into a new release.

sshipway:
Squashed all commits to a single one for clarity

aelsabbahy:
A few things:
1. Goss add command shouldn't add the exec element by default. It should be a manual change that a user can make.
2. Can you add some tests showcasing the skip flag, maybe a skipped test for a file or a service in `goss-shared.yaml`
3. Related to 2, when I compiled a local version of this and ran it with a skipped test (file) the code panicked.

sshipway:
1. Removed the default exec=command setting when adding
2. Added these for command, file and service; as well as a test of the exec attribute
3. When using skip attribute to skip the first test in a series, shouldSkip() on subsequent tests would then fail as the Found array would be empty.  Added some checks on array size to the function to prevent this.

Note that there seems to be an issue on my host with running test.sh where the containers need to be started with --privileged for the systemd to start the httpd.  Not sure why this should be, I'm looking into it.

sshipway:
All done.  There was a conflict with the other DNS test changes when I tried to squash, so I merged and resolved at my end.  Hope this is OK.
This has been an interesting first experience with go and travis...

aelsabbahy:
The PR looks great so far, did some initial quick test and it appears to be working.

Couple of things:
1. Can you run all the changed files through either `gofmt` or `goimports`
2. I'm wondering if it's clearer to add `Found:        []string{"false"},` to https://github.com/aelsabbahy/goss/blob/master/resource/validate.go#L58 rather than changing `shouldskip` the conditions for true/false currently seem a little confusing to me

Thanks for the awesome contribution.

sshipway:
1. I've run `gofmt` on it (another new utility for me); it apparently loves tabs and vertical space...
2. I was trying to fit in with the existing skip functionality, so as to minimise the required changes; TBH I'm not certain how the `Found` change would affect behaviour.  I felt it clearer and safer to extend `shouldskip` to remove the assumption that there are preexisting test results instead, as this make the code more robust.  Personally I think this way clearer; however, your project, your decision.

This is the first time I've used Travis, and I'm a total convert.  For open-source projects with multiple PR coming in, and an Agile workflow, it will save hours of work.  If gofmt is a non-negotiable for PR acceptance, maybe add a gofmt check to the Travis as well?

aelsabbahy:
Looks great,  thanks.

Opened an issue to #445 to add gofmt check as part of CI, thanks!

PR #437, opened 2019-04-10
Migrate from dep to go modules.

stefan-improbable:
Go modules seem to be the future of managing golang dependencies. This PR also simplifies `.travis.yml` and `Makefile` a bit.

Travis tests are failing due to `dnstest.io` being down, but apart from that, build and unit tests work.

@aelsabbahy if you want, I can send you a PR to disable those dnstest.io tests temporarily.


aelsabbahy:
Merged #440 to disable the test.  Rebase when you get a chance.

This change is great by the way, thanks for doing this!

stefan-improbable:
@aelsabbahy thanks! The tests are passing now.
As for DNS, I was thinking about using https://godoc.org/github.com/miekg/dns to spin up a DNS server in unit test `resource/dns_test.go` (that does not exist yet). In order to test this via Dockerfiles in `integration-tests/`, we could compile the test binaries via `go test -c` and execute them in docker. What do you think abou that?

stefan-improbable:
@aelsabbahy ping

aelsabbahy:
I will merge this in as part of the next release.

I really like the idea of unit testing dns. Though not sure the value of running it in docker.

stefan-improbable:
> I will merge this in as part of the next release.
> 
> I really like the idea of unit testing dns. Though not sure the value of running it in docker.

Thanks. I think the unit tests of DNS have lower priority for me right now given that `http://dnstest.io` is working now. The value of running in docker would be to test integration with different linuxes, provided goss uses libc which happens when `CGO_ENABLED` is set to `true` when building goss. But even then the value of running it in docker would not be that huge.

PR #441, opened 2019-04-28
Account for multiple colon's in dns key (IPv6 PTR) 

### LINKED ISSUE #428: Checking the the DNS PTR of a IPv6 address
Goss supports the checking of a reverse-DNS record of a IPv4 address fine.

But it fails for a IPv6 address like this (using goss 0.3.6):

    dns:
      PTR:2001:4860:4860::8888:
        resolvable: true
        server: 8.8.8.8
        addrs:
        - "google-public-dns-a.google.com."

It looks like the key parser doesn't expect IP addresses with colons:

https://github.com/aelsabbahy/goss/blob/2506b9448a4ec97f92c2c8556efdd52741658899/system/dns.go#L38-L41

Perhaps there also also code parts that need to be changed.

pysysops:
This fixes #428 and adds a very naive check that the type part of a key is an uppercase string.

Fixed the dnstest.io NSD servers. some cron tasks had been silently failing when re-signing records 🤷‍♂

### DIFF
--- a/system/dns.go
+++ b/system/dns.go
@@ -3,6 +3,7 @@ package system
 import (
 	"fmt"
 	"net"
+	"regexp"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


aelsabbahy:
Looks good. I'll look into reverting the disabled tests.

PR #454, opened 2019-06-06
Use GOSS_FILE environment variable for dgoss

jfcastroluis:
Allows use `GOSS_FILE` environment variable for `dgoss`. If not defined then use `goss.yaml` as it's now.

Only set the filename to copy into the container. In the container the name is `goss.yaml` to avoid more changes in the script.


jfcastroluis:
Yes, sorry :man_facepalming:
It's already fixed.

aelsabbahy:
Sorry for the late response on this. Looks good to me, can you resolve the conflict and add the new variable to the dgoss README?

Aside from that, good to go.

jfcastroluis:
I hope it's ready to merge. Let me know any change that I should do. Thanks.

aelsabbahy:
LGTM! Thank you for your contribution.

PR #456, opened 2019-06-11
Add check for mount disk usage

josephvoss:
This pull request extends the mount system check to also store disk usage information. It can be used with a mount resource to either check exact disk usage, or combined with the gomega matcher to ensure that disk usage is below/above a certain percent. Please let me know what tests I can add to verify this change, or if this information is out-of-scope for the mount resource checks.

Thanks


aelsabbahy:
rebased and merged, thanks!

PR #378, opened 2018-08-23
Adding support to provide DNS server port

davidcollom:
Provide support to provide a server and port when querying DNS Servers.

**Usecase:**
Ensuring consul is performing DNS queries as expected, you may have configuration like:
```
dns:
  consul.service.consul:
    resolvable: true
    server: 127.0.0.1:8125
```

PR #461, opened 2019-06-25
Synical/add dns server port

synical:
I was looking into adding some goss tests for a DNS server earlier and noticed I couldn't specify a custom port.

This PR adds support for specifying a port for the configured DNS server, falling back to 53 if no port has been specified.


aelsabbahy:
closed by https://github.com/aelsabbahy/goss/pull/378

PR #464, opened 2019-06-28
fix GetMounts

rmetzler:
`go get` didn't work with master, so I fixed it:

```
$ go get -u  github.com/aelsabbahy/goss
# github.com/aelsabbahy/goss/system
../../../go/src/github.com/aelsabbahy/goss/system/mount.go:91:33: not enough arguments in call to mount.GetMounts
	have ()
	want (mount.FilterFunc)
```

I used `nil` because of the docs

```
$ go doc github.com/docker/docker/pkg/mount.GetMounts
func GetMounts(f FilterFunc) ([]*Info, error)
    GetMounts retrieves a list of mounts for the current running process, with
    an optional filter applied (use nil for no filter).
```

### DIFF
--- a/system/mount.go
+++ b/system/mount.go
@@ -88,7 +88,7 @@ func (m *DefMount) Filesystem() (string, error) {
 }
 
 func getMount(mountpoint string) (*mount.Info, error) {
-	entries, err := mount.GetMounts()
+	entries, err := mount.GetMounts(nil)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


rmetzler:
I'm sorry, I don't know why Go is using `GO111MODULE=auto` and it fails.
When I use `GO111MODULE=on` my code is wrong and your's is correct.

PR #472, opened 2019-08-05
issue-469

pimperator:
introducing new variable LATEST_URL which points to latest tag on github
variable LATEST is filled now dynamically with either wget or curl
validating wether cut and wget or curl is present


aelsabbahy:
added some minor cleanups in this PR: #502

Issue #376, opened 2018-08-22
junit output broken in v0.3.6


aelindeman:
The most recent version includes a change from #304, which prints the Docker logs if the container failed to start, and prints `docker top` if the container ran successfully, both to stdout. This breaks the intent of #291, which was to redirect all non-test-result output to stderr, as to not break parsing junit results with an XML parser.

Before making a pull request I'd like to suggest doing one of two things, to gauge which option sounds better:

- redirecting `docker top` and `docker logs` commands to stderr (easy quick fix)
- adding an `output-file` parameter to the `junit` formatter, using a formatter option like the nagios formatter (slower more "correct"-seeming fix)

aelsabbahy:
Merged/released. Fixed in #473

PR #473, opened 2019-08-08
[dgoss] Use inspect instead of top in health check

jgeorgeson:
Just started using Goss, very nice! I came across #376 trying to publish JUnit results in my CI. This change tests if the container is running without producing any output to be redirected. This change is working in my environment. Additionally tested with your dgoss-examples repo nginx goss.yaml

```bash
$ GOSS_OPTS="--format junit" dgoss run nginx:1.11.10 > junit.xml
INFO: Container ID: 9ddf7dec
INFO: Sleeping for 0.2
INFO: Container health
INFO: Running Tests
INFO: Deleting container
$ head -3 junit.xml 
<testsuite name="goss" errors="0" tests="13" failures="0" skipped="0" time="0.010" timestamp="2019-08-08T22:46:38Z">
<testcase name="Process nginx running" time="0.000">
```

### DIFF
--- a/extras/dgoss/dgoss
+++ b/extras/dgoss/dgoss
@@ -90,7 +90,7 @@ case "$1" in
         fi
         [[ $GOSS_SLEEP ]] && { info "Sleeping for $GOSS_SLEEP"; sleep "$GOSS_SLEEP"; }
         info "Container health"
-        if ! docker top $id; then
+        if [ "true" != "$(docker inspect -f \"{{.State.Running}}\" $id)" ]; then
             docker logs $id
         fi
         info "Running Tests"


jgeorgeson:
Doesn't look like the Travis test includes dgoss. I should have mentioned too that I tested with the 3.7.0 release of the goss executable, didn't compile it from source.

jgeorgeson:
Anything additional I should here?

nathysOne:
I have the same issue as you, but I was able to resolve it by changing the following:

https://github.com/nathysOne/goss/commit/752c612e30fa772814ddbc928f05dafae3986792

strowi:
thx @nathysOne , ran into the same issue and removing the docker top fixed it.:)

aelsabbahy:
Looks good to me. I'll test/merge this after testing locally.

Travis test seems to be broken. Don't worry about the PR successful passing tests.

aelsabbahy:
For what it's worth, this PR seems to have broken the dgoss health check for me. It always said container `ERROR: the container failed to start`.

I committed a fix directly to master, you can find it here:  https://github.com/aelsabbahy/goss/commit/a4a501a4a755c882dd5ab920df3b9c6bb27d9779

jgeorgeson:
Hm, it appears to be the escaped quotes on the format string

```bash
[19:39:32]> justin@chymera in ~ 
$ docker inspect -f \"{{.State.Running}}\" 7cf3008d0e6fde1693d09b6a357858900c266f2f41b7aff4d59965e6c5cd3a59
"true"

[19:39:38]> justin@chymera in ~
$ docker inspect -f "{{.State.Running}}" 7cf3008d0e6fde1693d09b6a357858900c266f2f41b7aff4d59965e6c5cd3a59
true

[19:39:44]> justin@chymera in ~
$ if [ "true" != "$(docker inspect -f \"{{.State.Running}}\" 7cf3008d0e6fde1693d09b6a357858900c266f2f41b7aff4d59965e6c5cd3a59)" ]; then echo ended; else echo running; fi
ended

[19:40:46]> justin@chymera in ~
$ if [ "true" != "$(docker inspect -f "{{.State.Running}}" 7cf3008d0e6fde1693d09b6a357858900c266f2f41b7aff4d59965e6c5cd3a59)" ]; then echo ended; else echo running; fi
running
```
I just have it ingrained to wrap a command output in quotes for equality tests in case the command doesn't produce any output.

aelsabbahy:
> I just have it ingrained to wrap a command output in quotes for equality tests in case the command doesn't produce any output.

Good point, fixed: https://github.com/aelsabbahy/goss/commit/9dddd816d63f7e7217a1021f4e1c5ee0a50f0512

PR #489, opened 2019-10-23
Changes and updates in go.mod and adding go.sum

brejoc:
Building goss I noticed that the content of go.mod changed. Some dependencies have updates and some are simply marked as being indirect dependencies.

I've also added go.sum, since I think it makes sense to add this to VCS. This lets other know that they should get a working build with this specific set of verified dependencies.


aelsabbahy:
Thanks, rebased and merged

Issue #493, opened 2019-11-11
Permission denied when run dgoss with a less privileged user


racarvalho1987:
I am having an issue, regarding yaml file permission (goss_wait.yaml and goss.yaml)
once it has been copied to the container, and it is running with a less priviledged user. 
The permissions are the following:
-rwxr-xr-x 1 1004 1005 9825536 Nov  8 19:48 goss
-rw-r----- 1 1004 1005     186 Nov  8 19:48 goss.yaml
-rw-r----- 1 1004 1005     186 Nov  8 19:48 goss_wait.yaml

and I may need to have at least a 644 permission for goss.yaml and goss_wait.yaml

I already made a change into dgoss script, but I didn't find a "how to" to open a PR. 

How should I procced ?

aelsabbahy:
Can you provide an example of how to reproduce the issue.

As for creating a PR, a regular GitHub pull request with the fix will work just fine.

racarvalho1987:
@aelsabbahy This scenario happens when I execute dgoss in a jenkins pipeline.
I have a jenkins user that will execute the dgoss during the build, and the step that copy the files from $tmp_dir to $id/:goss will keep the original file permissons that is 640. 
My docker file has a nodejs base image, and I am using the node user to avoid security issues. This user don't have root permissions.
I am not sure if I was clear, and if that is easy the reproduce.

aelsabbahy:
Makes sense, as long as the change is to the tmp_folder copy of the files.

When you create a PR please reference it here and I'll review it.

fbartels:
@racarvalho1987 I think it makes sense to use the install command instead of a plain cp for these files. I have done the same already in https://github.com/aelsabbahy/goss/pull/478/files

aelsabbahy:
@fbartels I like `install` command, but eventually someone will try to run dgoss/dcgoss on a system without coreutils.

chmod will not necessarily "better", will work out-the-box on more machines.

racarvalho1987:
@fbartels its odd, but I also use dcgoss, and until now I have no permission problem with it, only dgoss. 
I see what you did and seems to be good, but I will keep the chmod as recommended by @aelsabbahy. 
Thanks for help

aelsabbahy:
Closing since this is already merged.

PR #494, opened 2019-11-12
Issue 493 Change permission from tmp_dir so less privileged user can …

racarvalho1987:
…run goss

Include a chmod 644 to set read permission from any user that runs the goss tests.

Let me know if there is something else that I need to do.

### DIFF
--- a/extras/dgoss/dgoss
+++ b/extras/dgoss/dgoss
@@ -42,6 +42,7 @@ run(){
       cp)
         info "Creating docker container"
         id=$(docker create "${@:2}")
+        chmod -R 644 $tmp_dir
         info "Copy goss files into container"
         docker cp $tmp_dir/. $id:/goss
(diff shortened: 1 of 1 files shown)


aelsabbahy:
Thinking the chmod makes more sense around line 33. For everything below the folder to not conflict with line 29.

aelsabbahy:
Fixes #493

racarvalho1987:
I did a commit to change the line but it won't work
Wait a moment please, I am working on it

aelsabbahy:
Yeah, Ideally goss binary should be 755, all other files 644.

Maybe chmod -R a+r $tmp_dir?

aelsabbahy:
Looks good on a quick eyeball check. I'll test it later today and merge.

racarvalho1987:
Ok @aelsabbahy thanks !

PR #497, opened 2019-11-22
Migrate imports from docker to moby

dacamp:


dacamp:
Actually, looks like docker wants you to still import it as `github.com/docker/docker` so closing this out.

aelsabbahy:
Can you point me to the doc on docker import for my reference.

dacamp:
I don't have any official word from ~Docker~ Moby but their [source code](https://github.com/moby/moby/blob/master/pkg/mount/mount.go) still references the _docker/docker_ import path `package mount // import "github.com/docker/docker/pkg/mount"`

aelsabbahy:
Thanks!

Issue #355, opened 2018-05-14
Semver comparison hints


vromero:
Hi all,

Thanks for this great piece of OSS.

I was wondering if there is a recommended way to bifurcate with the gotpl based semver versions.

I'm using dgoss to test a dockerimage with multiple versionso of the contained software and some tests apply to versions prior to X, and other to verions after Y or both.

I can probably do it with some simple regexes but I was hoping there was something better integrated.

aelsabbahy:
Currently, regex is the only solution. I want to add semver support, but it's hard to find a standard that works across all package managers.

If you have some library (preferably Go) or OSS software that you feel handle this problem in a generic cross platform way let me know and I'll dig into it.

vromero:
I see, thanks for the quick response.

aelsabbahy:
Adding support for this library could be neat: https://github.com/blang/semver#ranges 

It wouldn't work for all package managers, but would work with anything compliant with the semver spec.

vromero:
Agree, if it is not semver then whoever did choose that version scheme didn't mean it to be comparable.

pedroMMM:
@aelsabbahy I was playing with this on my repo and I had success by using https://github.com/hashicorp/go-version and creating another set of advanced mather. What I have already kind of works for me (it needs a better messaging). Would you like me to clean it up and create a PR?

aelsabbahy:
I would be interested in a semver matcher. I would prefer it be created as a gomega custom matcher that uses go-version.

pedroMMM:
@aelsabbahy This can be closed now. Also I will not mention the issue on commits anymore, that got messy fast :(

aelsabbahy:
Yeah works better in pr message.

Also, you don't have to rebase/push if there are no conflicts most of the time. Sometimes test count gets out of wack.

Issue #435, opened 2019-04-08
package version checks should allow ranges (Feature request)


sshipway:
When specifying a required package version, it should be possible to use ranges (> or < ).  This would allow a test for 'at least this version' (to avoid a known bug) or 'no later than this version' (to avoid incompatibilities).  Ideally, the syntax could also allow 'or not installed' as an extension to this.

May be related to #355

aelsabbahy:
It's somewhat related to #355, semver is a check I would like to add as an [advanced macher](https://github.com/aelsabbahy/goss/blob/master/docs/manual.md#advanced-matchers)

However, each package manager uses its own way to compare versions, so that may or may not be helpful depending on the package manager and it's comparison rules.

For example RPM uses the following logic to compare versions:
https://blog.jasonantman.com/2014/07/how-yum-and-rpm-compare-versions/

thatsk:
for Rpm based:-
you should be do 
   cmd := util.NewCommand("rpm", "-q", "--nosignature", "--nohdrchk", "--nodigest", "--qf", "%{VERSION}-%{RELEASE}.%{ARCH}\n", p.name)

pedroMMM:
#508 should take care of this.

Simple check `x > 4.0.0`:
```yaml
package:
  kernel:
    installed: true
    versions:
      version-gt: "4.0.0"
```

Range check `5.0.0 < x > 4.0.0`:
```yaml
package:
  kernel:
    installed: true
    versions:
      and:
      - version-gt: 4.0.0
      - version-lt: 5.0.0
```

aelsabbahy:
Just to clarify, @pedroMMM PR is a great addition for a semver matcher. I'm looking forward to that PR as it will allow goss with minor changes to support semver matching for command/http/file resources.

However, this does not solve package versioning/ranges, a quick test on the debian test image fails to parse for ~7-8% of the installed packages.

Also, rpm does not adhere to semantic versioning, for example.. but goss currently only checks '%{Version}' and not release, so may be a moot point for now.. Although, they may have to be added back in to do proper version comparisons in the future.

Here's an example where semver and rpm do different things:
https://semver.org/#spec-item-11
> Example: 1.0.0-alpha < 1.0.0-alpha.1 < 1.0.0-alpha.beta < 1.0.0-beta < 1.0.0-beta.2 < 1.0.0-beta.11 < 1.0.0-rc.1 < 1.0.0.
```
[...]

PR #508, opened 2019-12-02
[#355] Add Custom Version Matcher

pedroMMM:
Add Custom Version Matcher 
Add Unit Tests
Add Integration Tests
Update Documentation


integration-tests/test.sh
```diff
@@ -43,11 +43,17 @@ trap "rv=\$?; docker rm -vf $id; exit \$rv" INT TERM EXIT
...
-if [[ $os == "arch" ]]; then
-  egrep -q 'Count: 81, Failed: 0, Skipped: 3' <<<"$out"
-else
-  egrep -q 'Count: 97, Failed: 0, Skipped: 5' <<<"$out"
-fi
+case $os in
+  "arch")
+    egrep -q 'Count: 83, Failed: 0, Skipped: 3' <<<"$out"
+    ;;
+  "wheezy")
```
aelsabbahy: I would prefer to keep wheezy in sync with * if possible. It's already annoying that arch count is different than the rest.

docs/manual.md
```diff
@@ -852,9 +852,20 @@ package:
...
 ```
 
+Custom version matcher are available `version-(eq|gt|ge|lt|le)`: 
+
+```yaml
+package:
+  kernel:
+    installed: true
+    versions:
+      version-gt: "4.0.0"
```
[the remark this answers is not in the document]
pedroMMM: I followed your lead with the `BeNumerically` matcher based on the comment: 
> Golang json escapes '>', '<' symbols, so we use 'gt', 'le' instead

But if you favor this approach, I can take a look at the feasibility of having `>` and the other symbols.

integration-tests/test.sh
pedroMMM: The issue was that `wheezy` has a crazy version of `bash` installed that didn't parse correctly. Is that acceptable or should I get another test for `wheezy`?

docs/manual.md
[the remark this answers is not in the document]
pedroMMM: I did a few quick tests since I had some time tonight and the constraints for `go-version` and `semver` are more rigid than the simple functions (`GreaterThan`...). 

I also confirmed that `go-version` with the `NewVersion` constructor is less rigid than `semver` with the `Parse` constructor.

While this approach still doesn't fit all the versions especially since some of them completely disregard standards. I would say that I eyeballed the correct implementation without fully testing the alternatives. 

Of course, I am open to other ideas.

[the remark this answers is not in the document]
aelsabbahy: I've been traveling myself so only eyeballing tbis PR on my phone so far.

Are you saying the following isn't feasible?  
```yaml
semver_constraint: ">4.0.0, <5.0.0, !=4.4.0" 
```

As for package manager support, I think they would have to be handled individualy. Ex:
```yaml
semver_constraint: ">4.0.0, <5.0.0, !=4.4.0"
rpm_constraint: "..."
deb_constraint: "..."
```

I'm almost 100% sure (from memory) that semver clashes with rpm versioning schema.

I see semver as being useful outside the scope of package versioning. Http/command and other checks might be able to leverage it in the future.. once I normalize the non-gomega matchers.

Wow that JSON comment is 4yrs ago, back when Goss was only json based. Seems go offers a toggle for "<>" now:
https://stackoverflow.com/a/39079218

integration-tests/test.sh
aelsabbahy: What happens when you run the code against wheezy?

Also, to be 100% clear.. I love this PR, I don't view it as a solution to package versioning.. just useful for working with things that use semver.

See my comment here for why I don't think semver is sufficient:
https://github.com/aelsabbahy/goss/issues/435#issuecomment-486013785

aelsabbahy:
Leaving this here for reference/potential future additions:

* https://github.com/knqyf263/go-rpm-version
* https://github.com/knqyf263/go-deb-version
* https://linux.die.net/man/1/dpkg `--compare-versions` flag


I should be able to test this PR more later this week.

aelsabbahy:
Just got back from my travels, decided to give this a quick go on my laptop to see how it works.

I see the wheezy error you mention:
```yaml
package:
  bash:
    installed: true
    versions:
      # This is what the command returns
      #- 4.2+dfsg-0.1+deb7u3
      version-gt: "4.4.0"
```
```
dgoss run -t aelsabbahy/goss_wheezy
INFO: Container ID: e82d9a75
INFO: Sleeping for 0.2
INFO: Container health
INFO: Running Tests
Package: bash: installed: matches expectation: [true]
bash: version: Error: Expected a version or a list of versions.  Got:
    <[]string | len:1, cap:1>: [
        "4.2+dfsg-0.1+deb7u3",
    ]


Failures/Skipped:

bash: version: Error: Expected a version or a list of versions.  Got:
    <[]string | len:1, cap:1>: [
        "4.2+dfsg-0.1+deb7u3",
    ]

Total Duration: 0.007s
Count: 2, Failed: 1, Skipped: 0
```

docs/manual.md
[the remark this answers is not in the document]
pedroMMM: I am sorry for the delay, I didn't have any free time yesterday. Reading all of your comments from the past 2 days I understand how we weren't synced on what the functionality we are adding here.

I am fully on board with calling this Matcher a `semver_constraint` and not try to force it into working with the different package managers' version specifications. 

I will modify and clean the PR up. Thank you for your feedback!

integration-tests/test.sh
pedroMMM: Continuing on https://github.com/aelsabbahy/goss/pull/508#issuecomment-561477078

pedroMMM:
> Can you clarify what you mean by this? That seems concerning if it's not parsing them correctly is it doing the correct thing when diffing non-semver strings?
> 
> > I did a few quick tests since I had some time tonight and the constraints for go-version and semver are more rigid than the simple functions (GreaterThan...).

I have 2 examples from my unit tests:

```go
"4.3.42-r".GreatherThan("4.0.0") // returns true
"4.4.019-1".GreatherThan("4.0.0") // returns true
```

```go
go-version.Range("4.3.42-r > 4.0.0") // returns false
semver.Range("4.4.019-1 > 4.0.0") // errors out since no version segment can start with "0"
```

The `go-version` constraint fails because pre-release versions need to be pin down:
> A constraint with a pre-release can only match a pre-release version with the same base segments.

With all of this in mind, I propose we follow your plan of naming the Matcher `semver-constraint` and use the `semver` package since it follows the specifications to the letter.

I will wait for your feedback on the plan before making changes.

aelsabbahy:
semver-constraint + `semver` package sounds good to me.

This is a great start, really hoping other package managers get solved soon.

The semver matcher might not provide immediate value, but will be valuable in the near future once other work is done.

Preview of some upcoming work that will pair-up greatly with this in my opinion:
* Migrating http/command/file to use gomega matchers. They were an older implementation and it's a little tricky to convert to gomega matchers because they're `io.Reader`. I don't want to switch them to arrays, since that could drastically increase the memory usage of `goss` (ex. someone doing validation on a 1GB log file). Basically gotta write a custom implementation of [ContainElement](https://onsi.github.io/gomega/#containelementelement-interface) that uses readers.
* [jmespath](https://github.com/jmespath/go-jmespath) gomega matcher
* Whenever a package type is introduced that strictly adheres to semver (not sure if there are many to be honest)

milind69:
Do you think this feature will be released soon?
Interesting feature

integration-tests/test.sh
pedroMMM: The `wheezy` differences were removed.

docs/manual.md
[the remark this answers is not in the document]
pedroMMM: I have pivoted to implement the `semver-constraint` matcher.

integration-tests/goss/goss-shared.yaml
```diff
@@ -211,3 +211,9 @@ matching:
         - have-key-with-value:
             foo: bar
         - have-key: baz
+  semver:
+    content:
+      - 1.0.1
+      - 1.9.9
+    matches:
+      semver-constraint: ">1.0.0 <2.0.0 !=1.5.0"
```
aelsabbahy:
```suggestion
      semver-constraint: ">1.0.0 <2.0.0 !=1.5.0"
  semver2:
    content:
      - 1.0.1
      - 1.5.0
      - 1.9.9
    matches:
      not:
        semver-constraint: ">1.0.0 <2.0.0 !=1.5.0"
```

This adds a test proving the negation works. It might not be needed given all the unit tests.. guess I'm just being a little extra thorough for now.

The fact that you added unit tests for this is greatly appreciated by the way. Hoping with time goss moves to depending more on unit tests for stuff like this.

Side note, @OmarDarwish thanks for the `matching` resource, fit great for this test! Lightbulb just turned on for me on how useful it is for testing/debugging matchers.

integration-tests/test.sh
```diff
@@ -44,9 +44,9 @@ out=$(docker_exec "/goss/$os/goss-linux-$arch" --vars "/goss/vars.yaml" -g "/gos
 echo "$out"
 
 if [[ $os == "arch" ]]; then
-  egrep -q 'Count: 81, Failed: 0, Skipped: 3' <<<"$out"
+    egrep -q 'Count: 82, Failed: 0, Skipped: 3' <<<"$out"
```
aelsabbahy:
```suggestion
    egrep -q 'Count: 83, Failed: 0, Skipped: 3' <<<"$out"
```

integration-tests/test.sh
```diff
@@ -44,9 +44,9 @@ out=$(docker_exec "/goss/$os/goss-linux-$arch" --vars "/goss/vars.yaml" -g "/gos
...
 else
-  egrep -q 'Count: 97, Failed: 0, Skipped: 5' <<<"$out"
+    egrep -q 'Count: 98, Failed: 0, Skipped: 5' <<<"$out"
```
aelsabbahy:
```suggestion
    egrep -q 'Count: 99, Failed: 0, Skipped: 5' <<<"$out"
```

pedroMMM: Added manually

integration-tests/goss/goss-shared.yaml
pedroMMM: Added manually

pedroMMM: By all means, extra tests are always a good thing!

pedroMMM:
The PR is ready on my side, thank you for all the feedback @aelsabbahy 👍

integration-tests/goss/goss-shared.yaml
OmarDarwish: >Side note, @OmarDarwish thanks for the matching resource, fit great for this test! Lightbulb just turned on for me on how useful it is for testing/debugging matchers.

Glad it's providing use!

aelsabbahy:
Planning on doing an in depth look/merge of this over the weekend.

pedroMMM:
Take your time, I just want to be sure it's ready when you are.

matchers/semver_constraint.go
```diff
@@ -0,0 +1,104 @@
...
+	if err != nil {
+		return nil, false
+	}
+
+	return &v, true
+}
+
+func toVersions(in interface{}) ([]*semver.Version, bool) {
+	if v, ok := toVersion(in); ok {
+		return []*semver.Version{v}, ok
```
aelsabbahy: I think this solves my concern.. but want to get your take.

pedroMMM: You are correct it will work with a single `string` or `slice`. I didn't consider this use case but as it stands it will work just as you described it.

aelsabbahy:
For what it's worth, this works just fine:
```yaml
matching:
  semver:
    content:
      - 1.9.9
      - 6.13.2
    matches:
      contain-element:
        semver-constraint: ">1.0.0 <2.0.0 !=1.5.0"
```

Thanks for the contribution!

PR #514, opened 2019-12-09
Fix documentation

frederikbosch:
Missing letter, request-headers is plural.

### DIFF
--- a/docs/manual.md
+++ b/docs/manual.md
@@ -583,7 +583,7 @@ http:
     allow-insecure: false
     no-follow-redirects: false # Setting this to true will NOT follow redirects
     timeout: 1000
-    request-header: # Set request header values
+    request-headers: # Set request header values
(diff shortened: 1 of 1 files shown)


aelsabbahy:
Good catch, thanks!

frederikbosch:
Thank you, great library!

aelsabbahy:
Out of curiosity does the whole request-headers vs header thing seem inconsistent?

They were two different PRs by two contributors.. wondering if both should be normalized to "*headers"

frederikbosch:
I agree. I would choose for either singular or plural and not use them interchangeably.

frederikbosch:
What also is worth mentioning, or maybe I should create a PR for this, is that the host header is ignored, together with some headers. See [this line in net/http in the go library](https://github.com/golang/go/blob/master/src/net/http/request.go#L94).

PR #516, opened 2019-12-10
Implement GOSS_FILES_STRATEGY for dcgoss

frederikbosch:
`dcgoss` lacks GOSS_FILES_STRATEGY. It is not used while readme.md suggest it does.

### DIFF
--- a/extras/dcgoss/dcgoss
+++ b/extras/dcgoss/dcgoss
@@ -34,9 +34,23 @@ run(){
     [[ -e "${GOSS_FILES_PATH}/goss_wait.yaml" ]] && install -m ugo+rw "${GOSS_FILES_PATH}/goss_wait.yaml" "$tmp_dir"
     [[ -n "${GOSS_VARS}" ]] && [[ -e "${GOSS_FILES_PATH}/${GOSS_VARS}" ]] && install -m ugo+rw "${GOSS_FILES_PATH}/${GOSS_VARS}" "$tmp_dir"
 
-    info "Starting docker container"
-    docker-compose run -d -T --name "$service" -v "$tmp_dir:/goss" "$service"
-    docker logs -f "$service" > "$tmp_dir/docker_output.log" 2>&1 &
+    # Switch between mount or cp files strategy
+    GOSS_FILES_STRATEGY=${GOSS_FILES_STRATEGY:="mount"}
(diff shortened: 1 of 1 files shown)


aelsabbahy:
LGTM, but I didn't test it.

Tagging @fbartels since he is the original author of dcgoss. Can you review these changes?

If he doesn't respond I'll test/merge this later in the week.

aelsabbahy:
Thank you for taking the time to contribute to goss.

frederikbosch:
@aelsabbahy Thank you, again, for the wonderful library. The `cp` strategy is required in my case for running goss during CI/CD. 

The implementation is a tiny little different from `dgoss`. With `dgoss` you first create the container, copy the files and then run it. However, the `create` subcommand with docker-compose is deprecated and would not work anyhow. Therefore the files are copied after the run command for the container has been called. This is not a problem, because the files in question are the goss files, and they are not needed during the boot of the container.

aelsabbahy:
Makes perfect sense. Thanks for the explanation, been ages since I used docker-compose.

frederikbosch:
It's especially useful when the container has dependencies to other containers.

extras/dcgoss/dcgoss
```diff
@@ -34,8 +34,22 @@ run(){
...
+    case "$GOSS_FILES_STRATEGY" in
+      mount)
+        info "Starting docker container"
+        docker-compose run -d -T --name "$service" -v "$tmp_dir:/goss" "$service"
+        ;;
+      cp)
+        info "Creating docker container"
+        docker-compose run -d -T --name "$service" "$service"
+        info "Copy goss files into container"
+        docker cp $tmp_dir/. $service:/goss
```
fbartels:
```
In dcgoss line 48:
        docker cp $tmp_dir/. $service:/goss
                  ^------^ SC2086: Double quote to prevent globbing and word splitting.
                             ^------^ SC2086: Double quote to prevent globbing and word splitting.
```

frederikbosch:
Fixed

frederikbosch:
Thanks for the help

fbartels:
You fixed only the first ;-) `$service:/goss` also needs to be quoted.

frederikbosch:
That was stupid, thanks for your help!

aelsabbahy:
Thanks for the contribution!

PR #517, opened 2019-12-11
Adds support for custom docker files to dcgoss

AustinSchoen:
This adds support for custom docker compose files. Honestly there is probably a better way to pass options through to `docker-compose`, but my bash-fu level isn't high enough. For now this works with just the `-f` flag.

##### Checklist

### Description of change
Captures the `-f` parameter after script invocation and passes to the `docker-compose` call to use services defined a non "docker-compose.yaml" file. Also runs the config validation to check if there are invalid options or errors in the provided file.

### DIFF
--- a/extras/dcgoss/README.md
+++ b/extras/dcgoss/README.md
@@ -9,7 +9,7 @@ dcgoss is a convenience wrapper around goss that aims to bring the simplicity of
 
 ### Run
 
-Run is used to validate a docker container defined in `docker-compose.yml`. It expects both a `docker-compose.yml` and `goss.yaml` file to exist in the directory it was invoked from. Container configuration is used from the compose file, for example:
+Run is used to validate a docker container defined in `docker-compose.yml` by default. A custom dockerfile can be provided via the `-f` parameter, ie. `./dcgoss -f custom-docker-compose.yaml [run|edit] <service>`. Container configuration is used from the compose file, for example:
 
 **run:**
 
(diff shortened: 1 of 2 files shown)


aelsabbahy:
@fbartels I would love your take/review on this.

@AustinSchoen I haven't been involved in the development/maintenance of this script, deferring to @fbartels since he was the original author.

aelsabbahy:
@fbartels I would love your take/review on this.

@AustinSchoen I haven't been involved in the development/maintenance of this script, deferring to @fbartels since he was the original author.

fbartels:
Hi @AustinSchoen,

I was actually already thinking as well how to pass options to docker-compose. I am not sure about your approach, though.

For example I do have the feeling it would break my current tests, as I am loading multiple compose files through the COMPOSE_FILE environment variable at the moment, and this would be overwritten, when a file is specified with `-f`.

Maybe a mechanism modelled after $GOSS_OPTS would be a better approach. But since you then cannot count on a docker-compose.yml (or .yaml) to exist that check may need to go.

AustinSchoen:
@fbartels are you referring to the env var here? https://docs.docker.com/compose/reference/envvars/

Totally open to other implementations, this was just a quick adjustment to support the flag option.

fbartels:
Yes, exactly. You can either have it as an env, or put it into your `.env` and it will be picked up by docker-compose.

AustinSchoen:
Maintainer mentioned a workaround so I will close this. @fbartels if you'd like help implementing other docker-compose options or checks feel free to tag me in an isssue!

Issue #386, opened 2018-09-17
How to preserve /goss/docker_output.log on failure?


eifinger:
Hi,
I use dgoss in my CI pipeline to test my docker images.
I have a http test defined and if it fails I would like to know what exactly went wrong.

The way I understand the dgoss script it cleans up after a failure and thus deletes /goss/docker_output.log

What would be the best way to get the output of my docker container on failure and print it to my build log?

Thank you!

tholok97:
Wondering about this as well.

tholok97:
Got dgoss to print to the build log by adding a `docker logs "$id"` to the EXIT trap before cleanup like this:

```
# Main
tmp_dir=$(mktemp -d /tmp/tmp.XXXXXXXXXX)
chmod 777 "$tmp_dir"
trap 'ret=$?;docker logs "$id";cleanup;exit $ret' EXIT
```

Could someone please explain how this is normally acieved? Having the container log in the build log seems like a very legitimate thing to want, so I'm surprised it's not there by default.

aelsabbahy:
merged/relesed fix in #519

PR #519, opened 2019-12-12
[#386] Allow accessing container logs

m-jarzebowski:
##### Checklist

- [x] `make test-all` (UNIX) passes. CI will also test this.
- [ ] unit and/or integration tests are included

### Description of change

Added functionality to dgoss that allows persisting container logs


aelsabbahy:
Looks great!

Can you clarify that the logs are only retained if the variable is set in the documentation.

Fixes #386

m-jarzebowski:
Should be better now : )

Issue #445, opened 2019-05-12
Check for gofmt compliance as part of CI


aelsabbahy:

pedroMMM:
@aelsabbahy It can be closed now.

aelsabbahy:
Thanks for catching this and thanks for all your contributions. A release is being cut as we speak with some of your changes.

By the way in the future you can add fixes #tickernumber and it'll automatically close it: https://help.github.com/en/github/managing-your-work-on-github/closing-issues-using-keywords

I'll update the PR template to include a section for this.

PR #523, opened 2019-12-14
[#445] Add gofmt validation to CI

pedroMMM:
Add gofmt validation to CI requested on #445 

##### Checklist

- [x] `make test-all` (UNIX) passes. CI will also test this.

### Description of change

Add a `make fmt` that checks if the code formating is valid. Ensure it's part of the main `make`  invoked by the CI and fail the build if the formating invalid.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,7 @@ fmt:
 	$(info INFO: Starting build $@)
 	{ \
 set -e ;\
-fmt=$$(gofmt -l .) ;\
+fmt=$$(gofmt -l ${GO_FILES}) ;\
 [ -z "$$fmt" ] && echo "valid gofmt" || (echo -e "invalid gofmt\n$$fmt"; exit 1)\
 }
 
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,14 @@ lint:
 	#go tool vet .
 	golint $(pkgs) | grep -v 'unexported' || true
 
+fmt:
+	$(info INFO: Starting build $@)
+	{ \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -24,6 +24,14 @@ lint:
 	#go tool vet .
 	golint $(pkgs) | grep -v 'unexported' || true
 
+fmt:
+	$(info INFO: Starting build $@)
+	{ \
+set -e ;\
+fmt=$$(gofmt -l .) ;\
+[ -z "$$fmt" ] && echo "valid gofmt" || (echo "invalid gofmt"; exit 1)\
```
aelsabbahy:
```suggestion
[ -z "$$fmt" ] && echo "valid gofmt" || (echo -e "invalid gofmt\n$$fmt"; exit 1)\
```

When I ran it it just said: "invalid gofmt" without any other details, this adds the actual errors.

pedroMMM: Great catch!

pedroMMM:
I added both of your suggestions.

aelsabbahy:
`GO_FILES` ignores `*_test.go` files. Is this intentional?

pedroMMM:
> `GO_FILES` ignores `*_test.go` files. Is this intentional?

Yes, analysis tools usually will ignore test files. But with that said, I don't feel strong enough about it so I am open to change it if you want.

aelsabbahy:
Ok, cool. Just wanted to make sure it was intentional. Thanks for this (and all the other work so far)!

Issue #544, opened 2020-02-08
usage as a library


ripienaar:
**Describe the feature:**

I would like to be able to use goss as a package from within other go applications, there are various infra automation scenarios where its desirable to deliver the system as a single binary.

I'd like to do an integration into choria.io and some other things

**Describe the solution you'd like**

At least I would like to be able to do a validate, a few of the functions in validate.go, store.go, template.go and a few others do os.Exit() and require a number of unexported functions.

Would you be open to adjusting things a bit so a external package can use goss as a package for at least doing Validate?

Similar to #112 

I'd roughly keep things are, even accepting a cli context in validate but I'd add a few helpers to let me do:

```
err = goss.ValidateWithOptions(goss.GossFile(...), goss.VarsFile(...), goss.OutputFormat(goss.JSON))
```

additional options for pkg type, output format and options, color, duration, retry duration and max concurrency

**Describe alternatives you've considered**

I tried to write this from the outside - alas, the os.Exit() etc is a pain

pedroMMM:
This is a reasonable request and to be honest, I am not a fan of using `os.Exit()` and `cli.Context` outside of the CLI functions.

> additional options for pkg type, output format and options, color, duration, retry duration and max concurrency

This is confusing, could you go into more detail of your use case, please? 

My only worry is that @aelsabbahy will not prioritize this highly. Then you could work on this and I would love to help out.

ripienaar:
> This is a reasonable request and to be honest, I am not a fan of using `os.Exit()` and `cli.Context` outside of the CLI functions.
> 
> > additional options for pkg type, output format and options, color, duration, retry duration and max concurrency
> 
> This is confusing, could you go into more detail of your use case, please?

simply they are the other options the package today get from the CLI - I will give equivelant in code.

aelsabbahy:
Marking this as approved. It's very much an improvement to the current code. I probably won't dedicated personal time to this due to other issues being higher priority for me. That said, I will gladly review and merge changes that introduce this with testing.

Honestly, the parts of the code that will benefit most from this refactor are the parts that would benefit most from having proper go based tests in place.

aelsabbahy:
> This is a reasonable request and to be honest, I am not a fan of using os.Exit() and cli.Context outside of the CLI functions.

![6b7und8rs1v21](https://user-images.githubusercontent.com/6783261/74108545-45378000-4b30-11ea-9377-7eff6e6388f9.png)


[...]

PR #534, opened 2020-01-02
[#439] Add --vars-inline that is capable of overwriting --vars

pedroMMM:
##### Checklist

- [x] `make test-all` (UNIX) passes. CI will also test this
- [x] unit and/or integration tests are included (if applicable)
- [x] documentation is changed or added (if applicable)

### Description of change

Add `--vars-inline` to dynamically pass variables which can overwrite variables sourced from `--vars`, but with no deep merging.

Closes #439

PR #546, opened 2020-02-11
remove cli.Context from main goss packages

ripienaar:
##### Checklist

### Description of change

This turns the cli.Context into a plain struct and passes the
struct, or fields from the struct, into all the various goss
internals. We then remove use of cli.Context everywhere
but in cmd/goss

This is a step one towards being able to use goss as a package
without requiring the CLI stuff

This is a one of several smaller PRs that I will send over the next
while towards resolving #544

### DIFF
--- a/cmd/goss/goss.go
+++ b/cmd/goss/goss.go
@@ -46,7 +46,7 @@ func newRuntimeConfigFromCLI(c *cli.Context) *goss.RuntimeConfig {
 	}
 
 	if c.Bool("color") {
-		cfg.NoColor = bp(true)
+		cfg.NoColor = bp(false)
 	}
 
 	return cfg
--- a/cmd/goss/goss.go
+++ b/cmd/goss/goss.go
@@ -14,7 +14,8 @@ import (
 
 var version string
 
-func NewRuntimeConfigFromCLI(c *cli.Context) *goss.RuntimeConfig {
+// converts a cli context into a goss RuntimeConfig
+func newRuntimeConfigFromCLI(c *cli.Context) *goss.RuntimeConfig {
 	bp := func(b bool) *bool { return &b }
 
 	cfg := &goss.RuntimeConfig{
@@ -127,7 +128,7 @@ func main() {
 				},
 			},
 			Action: func(c *cli.Context) error {
-				code, err := goss.Validate(NewRuntimeConfigFromCLI(c), startTime)
+				code, err := goss.Validate(newRuntimeConfigFromCLI(c), startTime)
 				if err != nil {
 					color.Red(fmt.Sprintf("Error: %v\n", err))
 				}
@@ -178,7 +179,7 @@ func main() {
 				},
 			},
 			Action: func(c *cli.Context) error {
-				goss.Serve(NewRuntimeConfigFromCLI(c))
+				goss.Serve(newRuntimeConfigFromCLI(c))
 				return nil
 			},
 		},
@@ -193,7 +194,7 @@ func main() {
 				},
 			},
 			Action: func(c *cli.Context) error {
-				fmt.Print(goss.RenderJSON(NewRuntimeConfigFromCLI(c)))
+				fmt.Print(goss.RenderJSON(newRuntimeConfigFromCLI(c)))
 				return nil
 			},
 		},
@@ -202,7 +203,7 @@ func main() {
 			Aliases: []string{"aa"},
 			Usage:   "automatically add all matching resource to the test suite",
 			Action: func(c *cli.Context) error {
-				goss.AutoAddResources(c.GlobalString("gossfile"), c.Args(), NewRuntimeConfigFromCLI(c))
+				goss.AutoAddResources(c.GlobalString("gossfile"), c.Args(), newRuntimeConfigFromCLI(c))
 				return nil
 			},
 		},
@@ -221,7 +222,7 @@ func main() {
 					Name:  "package",
 					Usage: "add new package",
 					Action: func(c *cli.Context) error {
-						goss.AddResources(c.GlobalString("gossfile"), "Package", c.Args(), NewRuntimeConfigFromCLI(c))
+						goss.AddResources(c.GlobalString("gossfile"), "Package", c.Args(), newRuntimeConfigFromCLI(c))
 						return nil
 					},
 				},
@@ -229,7 +230,7 @@ func main() {
 					Name:  "file",
 					Usage: "add new file",
 					Action: func(c *cli.Context) error {
-						goss.AddResources(c.GlobalString("gossfile"), "File", c.Args(), NewRuntimeConfigFromCLI(c))
+						goss.AddResources(c.GlobalString("gossfile"), "File", c.Args(), newRuntimeConfigFromCLI(c))
 						return nil
 					},
 				},
@@ -243,7 +244,7 @@ func main() {
 						},
 					},
 					Action: func(c *cli.Context) error {
-						goss.AddResources(c.GlobalString("gossfile"), "Addr", c.Args(), NewRuntimeConfigFromCLI(c))
+						goss.AddResources(c.GlobalString("gossfile"), "Addr", c.Args(), newRuntimeConfigFromCLI(c))
 						return nil
 					},
 				},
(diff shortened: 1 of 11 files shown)


validate.go
```diff
@@ -9,28 +9,51 @@ import (
...
 	"github.com/aelsabbahy/goss/outputs"
 	"github.com/aelsabbahy/goss/resource"
 	"github.com/aelsabbahy/goss/system"
 	"github.com/aelsabbahy/goss/util"
-	"github.com/fatih/color"
-	"github.com/urfave/cli"
 )
 
-func getGossConfig(c *cli.Context) GossConfig {
+type RuntimeConfig struct {
```
ripienaar: Arguably not the best way to go here, I'll see how it works out and refactor a bit while working on #544, for now it gets me moving with the other changes mentioned and already this is usable as a package - minus logging uglyness and os.Exit()

pedroMMM: If @aelsabbahy isn't bothered by this we can leave it here for now until we find a better home for it.

ripienaar:
> Correct me if I am wrong but I don't see the need to use a pointer for the `NoColor` property on the `RuntimeConfig`. Using the boolean value would reduce the assignment to a single line and remove the small mental overhead.

perhaps not but the use of color and no-color options basically makes it a 3 state boolean, if we stay color is on by default then sure we can easily simplify things.

I'll add some comments, but the prevailing style in the code base is not to comment so seems a bit of a stretch to insist on that, but sure I'll add.

pedroMMM:
> perhaps not but the use of color and no-color options basically makes it a 3 state boolean, if we say color is on by default then sure we can easily simplify things.

I stand correct then! I am sorry I missed that. Thank you.

pedroMMM:
> I'll add some comments, but the prevailing style in the code base is not to comment so seems a bit of a stretch to insist on that, but sure I'll add comments to new code I introduce.

I am sorry if I came out as demanding comments. I some times still have hiccups with English since it's the 3rd language I learned. I merely wanted to you add the comments if it wasn't a big brother to you.

We very recently added Code Climate to help improve the quality of the code and this kind of non-functional details that help users and contributors alike. @aelsabbahy will not be enforcing all of the Code Climate problems as of now but we should try to improve on them when feasible. 

Again I am sorry for the miscommunication. Thank you for improving Goss.

aelsabbahy:
This looks great, thanks for the contribution.

I'll do some local testing and merge in the coming days. I don't anticipate any issues.

Can you update the #544 ticket with all the different PRs you plan on submitting for this effort. It would help mentally group them up and in seeing the big picture.

The codeclimate addition is to slowly improve the code quality/unit-test coverage of goss.

ripienaar:
No worries about the comments thing.

I put a 2nd commit here that gets rid of all the os.Exit(), can make it a 2nd PR if you prefer, I had some spare cycles so just got through that while waiting for review here.

Tests failing on my local system so will fix whatever travis has to say if any

aelsabbahy:
Awesome.

Drop a comment on here when you're done making changes to the PR so I know that it's ready for review/merge.

ripienaar:
I am done for now I think - travis green and some basic local testing passed, but I am not super familiar with all the goss cli edge cases so I think I'm read for review.

Maybe pick up on this next week again, unfortunately I am not really sure what would be a list of PRs I'll do, mainly just working through it till I have it in a place I like - also first time in this code base so hard to anticipate what's going to be needed

store.go
```diff
@@ -24,7 +23,7 @@ const (
 )
 
 var OutStoreFormat = UNSET
-var TemplateFilter func(data []byte) []byte
+var TemplateFilter func(data []byte) ([]byte, error)
```
pedroMMM: The changes around the `TemplateFilter` and the rest of the `store.go` and `template.go` will conflict with another PR #534. @aelsabbahy #534 is ready to be merged and maybe it should so @ripienaar makes the right modifications to his code. Funny enough I ended up tweaking my code to make it easier for Package usage as well.

ripienaar: ok, once you merge that one I'll rebase here

aelsabbahy: Merged #534

ripienaar:
hmm, I guess I missed something in the rebase, but I cant see it, any ideas? @pedroMMM?

pedroMMM:
> hmm, I guess I missed something in the rebase, but I cant see it, any ideas? @pedroMMM?

I will take a good look at the code soon but one of the little bits that PR did was re-enabling Travis to run the Integration Tests that had been disabled by mistake. So the issue might have been there before the rebase. Never the less I will go through it.

ripienaar:
Yeah it appears to be that a YAML file is being read as JSON and some fatal exit I removed - turning it into a err instead - is changing behaviour, I'll also look in the morning again bit late now.

ripienaar:
OK - that one was me declaring the temlate output block local by accident, will fix what remains tomorrow.

resource/process.go
```diff
@@ -19,17 +19,21 @@ func (p *Process) SetID(id string) { p.Executable = id }
...
 func (p *Process) GetMeta() meta    { return p.Meta }
 
-func (p *Process) Validate(sys *system.System) []TestResult {
+func (p *Process) Validate(sys *system.System) ([]TestResult, error) {
+	var results []TestResult
 	skip := false
-	sysProcess := sys.NewProcess(p.Executable, sys, util.Config{})
+	sysProcess, err := sys.NewProcess(p.Executable, sys, util.Config{})
+	if err != nil {
+		return results, err
```
ripienaar: Here I have to check the error, resource.Process does not let Validate() return any error, whats the right way to construct a Validatevalue here for this situation where the sys.NewProcess fails?

ripienaar: This can fail because sys.GetProcs can fail in the go-ps package

ripienaar: OK I managed to resolve this issue while keeping the error handling, so good for review with all tests green `The command "make all" exited with 0.`

ripienaar:
Anything else to do here?

pedroMMM:
> Anything else to do here?
@ripienaar nothing for now, just waiting on @aelsabbahy to review.

cmd/goss/goss.go
```diff
@@ -8,11 +8,50 @@ import (
...
+		Password:          c.String("password"),
+		Debug:             c.Bool("debug"),
+	}
+
+	if c.Bool("no-color") {
+		cfg.NoColor = bp(true)
+	}
+
+	if c.Bool("color") {
+		cfg.NoColor = bp(true)
```
aelsabbahy:
```suggestion
		cfg.NoColor = bp(false)
```

validate.go
```diff
@@ -9,87 +9,138 @@ import (
...
+	}
 
 	if len(gossConfig.Resources()) == 0 {
-		fmt.Printf("Error: found 0 tests, source: %v\n", source)
-		os.Exit(1)
+		return nil, fmt.Errorf("found 0 tests, source: %v", source)
 	}
-	return gossConfig
+
+	return &gossConfig, nil
```
aelsabbahy: curious, why change this to a pointer?

system/system.go
```diff
@@ -48,14 +48,17 @@ func (s *System) Ports() map[string][]GOnetstat.Process {
...
-		s.procMap = GetProcs()
+		s.procMap, err = GetProcs()
 	})
-	return s.procMap
+
+	return s.procMap, err
 }
 
-func New(c *cli.Context) *System {
+func New(packageManager string) *System {
```
aelsabbahy: I can see some benefits to passing in RuntimeConfig object to function.

This can be changed as a later PR.

validate.go
aelsabbahy: This can probably move to a separate file.. but I'm fine with this being a later PR.

validate.go
ripienaar: It’s easier to return nil on error. Also NewGossConfig does the same so I made this do that too. Can revert if you like.

In general I think returning nil, err vs empty config that will anyway just be discarded is more idiomatic - old code just exit’d so seemed like a sane change.

aelsabbahy: Makes sense

system/system.go
ripienaar: yup, however it would cause an import cycle.

We can resolve this when, towards the end, I look to merge util.Config and goss.RuntimeConfig, i held off on that to keep the PR from just going totally crazy cross many boundaries.

ripienaar:
rebased to master

aelsabbahy:
Just a quick update on this. I'm planning on cutting a goss release and then merging this right after.

I was going to cut the release yesterday, but ran into some issues due to recent travis.yml changes, I think that's fixed now (https://github.com/aelsabbahy/goss/commit/da32f41f505aa8ae9e9bb480ce9ed4dc122a6f17). So it's just a matter of finding some time this weekend or monday to get it all taken care of.

aelsabbahy:
Just cut the goss release, merging this... we can continue the discussion on the other PR.

Thank you so much for taking the time for this.

PR #547, opened 2020-02-15
Correct `headers` for http in manual.md

johnpaulett:
Per [resource/https.go](https://github.com/aelsabbahy/goss/blob/master/resource/http.go#L17) the attribute is plural, `headers`.


aelsabbahy:
Good catch, thanks!

Issue #549, opened 2020-02-20
Invalid Junit output from dcgoss


officerJones:
**Describe the bug**
dcgoss produces an unparseable junit xml due to logging to be present in the test results.

**How To Reproduce**
Set output format to unit & run dcgoss
```
export GOSS_OPTS="--format junit"
dcgoss run *service* > testresults.xml
```

**Expected Behavior**
testresults.xml contains a parsable xml in junit format.

**Actual Behavior**
testresults.xml also contains following output:
- Text: *service name*
- docker top output
- Text: Deleting *service name*

**Environment:**
 - Version of goss: v0.3.9

aelsabbahy:
Dgoss gets around this by redirecting information output to stderr.

Seems a similar thing needs to be added to dcgoss.

I'm approving this, feel free to submit a PR. I personally don't maintain user submitted wrapper scripts (dcgoss & kgoss)

officerJones:
I tried to create a pull request for this. 
It's the first time I try to contribute to a project so if anything should be done differently, just let me know.

aelsabbahy:
Closed by #550

PR #550, opened 2020-02-20
Fix invalid junit output for dcgoss

officerJones:
##### Checklist

### Description of change
extras/dcgoss/dcgoss was changed to:
- automatically remove started containers to make sure line "Deleting service" is not present in output
- suppress logging of docker-compose run command
- use docker inspect instead of docker top conform dcgoss

I created bug #549 for this issue which this PR should fix.


aelsabbahy:
@fbartels I would love your take/review on this.

PR #552, opened 2020-02-22
Use Pattern Rules in Makefile

directionless:
##### Checklist

### Description of change

Convert the release/ section of the Makefile to using pattern rules. This facilitates building for other platforms.


aelsabbahy:
Makes sense, thanks!

PR #561, opened 2020-03-23
Weakiwi/feature/add block device test

weakiwi:
##### Checklist

### Description of change

This PR is to add block device test.
Why set it as `draft` because I don't know how to write intergration test for this feature.


pedroMMM:
You should be able to do some basic testing via Docker. 

Once you got something there you should add and test the `add` functionality. @aelsabbahy might be needed to explain better how that functions since I haven't touched that section of the codebase but from my understanding, it's a simple autogenerated functionality.

weakiwi:
> You should be able to do some basic testing via Docker.
> 
> Once you got something there you should add and test the `add` functionality. @aelsabbahy might be needed to explain better how that functions since I haven't touched that section of the codebase but from my understanding, it's a simple autogenerated functionality.

Yes. The problem is that block device testing is to test hardware infomation. And dosen't like qemu or virtualbox, docker can't specify excepted hardware information. Like:
1. vendor id for disk
2. size of disk
3. type of drive

pedroMMM:
> > You should be able to do some basic testing via Docker.
> > Once you got something there you should add and test the `add` functionality. @aelsabbahy might be needed to explain better how that functions since I haven't touched that section of the codebase but from my understanding, it's a simple autogenerated functionality.
> 
> Yes. The problem is that block device testing is to test hardware infomation. And dosen't like qemu or virtualbox, docker can't specify excepted hardware information. Like:
> 
> 1. vendor id for disk
> 2. size of disk
> 3. type of drive


@weakiwi I have been thinking about this one. The best I can come up with is that as long as we can verify that the device exists and as much data we can get out of Docker then this should be enough to prove that Goss is calling and interpreting the 3d party library correctly. It's not perfect but we will have to assume that dependency works as advertized since testing it isn't Goss responsibility.

While we all know the consumers aren't responsible for testing 3rd libraries, this still feels like a bad shortcut. @aelsabbahy would you prefer to take a different approach?

aelsabbahy:
I started an issue to discuss this feature here #570.

aelsabbahy:
Closing this as per discussion on #570

PR #563, opened 2020-03-24
allow configs to be merged outside of goss

ripienaar:
##### Checklist

### Description of change

As part of making goss usable as a package it would be useful to manipulate configs externally

Merging configs is complicated and dependant on the features available in Goss so it's worth exposing this as a function others can build on.

### DIFF
--- a/goss_config.go
+++ b/goss_config.go
@@ -46,6 +46,70 @@ func NewGossConfig() *GossConfig {
 	}
 }
 
+// Merge consumes all the resources in g2 into c, duplicate resources
+// will be overwritten with the ones in g2
+func (c *GossConfig) Merge(g2 GossConfig) {
+	for k, v := range g2.Files {
+		c.Files[k] = v
+	}
+
+	for k, v := range g2.Packages {
+		c.Packages[k] = v
+	}
+
+	for k, v := range g2.Addrs {
+		c.Addrs[k] = v
+	}
+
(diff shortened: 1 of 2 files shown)


goss_config.go
```diff
@@ -102,65 +166,7 @@ func interfaceMap(slice interface{}) map[string]interface{} {
...
-	}
-
-	for k, v := range g2.HTTPs {
-		g1.HTTPs[k] = v
-	}
-
-	for k, v := range g2.Matchings {
-		g1.Matchings[k] = v
-	}
+	g1.Merge(g2)
```
ripienaar: not clear to me why here `g1.Gossfiles` is set to nil, does not neccesary seems desirable in a generic merge, so I left it here to preserve existing behaviour and didnt put it on the new `Merge()` function.

pedroMMM: @ripienaar this is interesting. I tranced this code to the [conception](https://github.com/aelsabbahy/goss/pull/2) and it's still a mystery why it's needed. 😞

ripienaar:
Relates to https://github.com/aelsabbahy/goss/issues/544

goss_config.go
aelsabbahy: Not sure to be honest, maybe the pointer was causing some panic or something back then.

I'm not sure there's any need for it now. *shrug*

Issue #564, opened 2020-03-31
http resource - Host header not set


johnk-novu:
**Describe the bug**
Go's `net/http` library excludes/skips setting certain request headers, including `Host`.

**How To Reproduce**
Write a `goss` test that relies on a host header being set.
See https://gist.github.com/johnk-novu/fd3671819c59da227296c54ef393411c for example.

**Expected Behavior**
`Host` header is presented to the webserver.

**Actual Behavior**
No `Host` header(s) are presented to the webserver.

**Environment:**
 - Version of goss: v0.3.10 / built off `master`
 - OS/Distribution version (if applicable): MacOS Mojave 10.14.5

johnk-novu:
PR in the works, will re-target to this repo when it is closer to being "ready"

https://github.com/johnk-novu/goss/pull/1

johnk-novu:
PR open https://github.com/aelsabbahy/goss/pull/565

Struggling to get test running locally due to some laptop issues, getting Go setup on another machine to run tests.

johnk-novu:
Confirmed `make test-all` passes. PR should be good to go.

aelsabbahy:
Closed by #565

PR #565, opened 2020-03-31
Host header update

johnk-novu:
##### Checklist

### Description of change
Go's `http` ignores/removes certain headers (n1, n2), including but not limited to the `Host` header. See: [github issue](https://github.com/golang/go/issues/29865) and [go source](https://github.com/golang/go/blob/cad6d1fef5147d31e94ee83934c8609d3ad150b7/src/net/http/request.go#L92-L99)

This change detects the presence of that header (or headers) and sets it on the http request.

Issue opened: https://github.com/aelsabbahy/goss/issues/564

I also took the liberty of making a couple of unrelated changes:
* update README.md
* rename a duplicate test in goss-shared.yml


integration-tests/goss/goss-shared.yaml
```diff
@@ -180,6 +180,14 @@ interface:
       contain-element:
         have-prefix: '172.17'
 http:
+  https://httpbin.org/headers:
```
ripienaar, line 183: this one will be overwrote with the one below on line 191 as its the same url.

system/http.go
```diff
@@ -40,6 +41,7 @@ func NewDefHTTP(httpStr string, system *System, config util.Config) HTTP {
 		str := strings.SplitN(r, ": ", 2)
 		headers.Add(str[0], str[1])
 	}
+	fmt.Printf("\nheaders: %v\n\n", headers)
```
ripienaar, line 44: probably not needed

system/http.go
```diff
@@ -86,6 +88,14 @@ func (u *DefHTTP) setup() error {
 		return u.err
 	}
 	req.Header = u.RequestHeader.Clone()
+
+	if host, ok := u.RequestHeader["Host"]; ok {
+		if len(host) > 1 {
+			fmt.Fprintf(os.Stderr, "multiple host headers detected, using first: (%v)", host[0])
```
ripienaar, line 94: these should not produce output like this, goss can be used as a package by other projects and this is not a helpful way to communicate back this information in that way - there's no way to do this in current goss.

integration-tests/goss/goss-shared.yaml
johnk-novu: good catch 👍

system/http.go
johnk-novu: yup, leftover from debugging, will remove

system/http.go
johnk-novu: 👍 will remove

integration-tests/goss/goss-shared.yaml
johnk-novu: @ripienaar 

I wonder if that would also apply to the second http basic auth check (https://github.com/aelsabbahy/goss/blob/f95fac2bb3dab16dd932d000ebe09453beb767ff/integration-tests/goss/goss-shared.yaml#L203)? If so, I could update that while I am at it.

ripienaar: Yeah I saw that too - it would yes

johnk-novu:
Confirmed tests pass, I think this is good to go.

system/http.go
```diff
@@ -86,6 +86,11 @@ func (u *DefHTTP) setup() error {
 		return u.err
 	}
 	req.Header = u.RequestHeader.Clone()
+
+	if host, ok := u.RequestHeader["Host"]; ok {
+		req.Host = host[0]
+	}
```
aelsabbahy, line 92: Wonder if we should use the Header.Get method: https://golang.org/pkg/net/http/#Header.Get

```suggestion
	if host := req.Header.Get("Host"); host != "" {
		req.Host = host
	}
```

integration-tests/goss/goss-shared.yaml
```diff
@@ -188,11 +188,19 @@ http:
       - "Foo: bar"
     headers: ["Content-Type: application/json"]
     body: ['"Foo": "bar"']
+  https://httpbin.org/headers?host:
```
aelsabbahy, line 191: Can you point me to the documentation for `?host`

integration-tests/goss/goss-shared.yaml
```diff
@@ -188,11 +188,19 @@ http:
...
+    request-headers:
+      - "Host: example.com"
+    headers: ["Content-Type: application/json"]
+    body: ['"Host": "example.com"']
   https://httpbin.org/basic-auth/username/secret:
     status: 200
     username: username
     password: secret
-  https://httpbin.org/basic-auth/username/secret:
+  https://httpbin.org/basic-auth/username/secret?failure:
```
aelsabbahy, line 203: can you point me to the documentation for `?failure`

integration-tests/goss/goss-shared.yaml
johnk-novu: I am just appending an arbitrary GET param that httpbin seems to ignore to prevent duplicating the other check against `https://httpbin.org/headers`. There may be a cleaner way to prevent duplicating the checks, this didn't seem to impact the behavior of httpbin, so I gave it a shot.

johnk-novu:
@aelsabbahy I made the documentation update, let me know what you think!

https://github.com/aelsabbahy/goss/pull/565/commits/cd4aeb3017c44c02b005c83cd14452f2af84b443

aelsabbahy:
Looks great. There's going to be a bit of juggling with other PRs.. but I should have this merged/released in the next ~2weeks.

johnk-novu:
@aelsabbahy awesome thanks! looking forward to it, this will allow me to implement a very helpful test :)

aelsabbahy:
Awesome change, thank you!

PR #571, opened 2020-04-08
Code breaks when resource_list is generated from genny

aelsabbahy:
### Description of change

This isn't really a PR. But it shows what got changed/broken with #546

This is just a commit of resource_list.go after running `make gen`.

@ripienaar The resouce_list.go is a generated file from resource_list_genny.go. It seems the PR linked above breaks development since it modified the file directly.

When you get a chance, can you please take a look at this since it will block any new resources from being added without the code being reverted.

### DIFF
--- a/resource/resource_list.go
+++ b/resource/resource_list.go
@@ -820,11 +820,7 @@ func (ret *PortMap) UnmarshalYAML(unmarshal func(v interface{}) error) error {
 type ProcessMap map[string]*Process
 
 func (r ProcessMap) AppendSysResource(sr string, sys *system.System, config util.Config) (*Process, error) {
-	sysres, err := sys.NewProcess(sr, sys, config)
-	if err != nil {
-		return nil, err
-	}
-
+	sysres := sys.NewProcess(sr, sys, config)
 	res, err := NewProcess(sysres, config)
 	if err != nil {
 		return nil, err
@@ -837,24 +833,19 @@ func (r ProcessMap) AppendSysResource(sr string, sys *system.System, config util
 	return res, nil
 }
 
-func (r ProcessMap) AppendSysResourceIfExists(sr string, sys *system.System) (*Process, system.Process, bool, error) {
-	sysres, err := sys.NewProcess(sr, sys, util.Config{})
-	if err != nil {
-		return nil, nil, false, err
-	}
-
+func (r ProcessMap) AppendSysResourceIfExists(sr string, sys *system.System) (*Process, system.Process, bool) {
+	sysres := sys.NewProcess(sr, sys, util.Config{})
 	// FIXME: Do we want to be silent about errors?
 	res, _ := NewProcess(sysres, util.Config{})
 	if e, _ := sysres.Exists(); e != true {
-		return res, sysres, false, nil
+		return res, sysres, false
 	}
 	if old_res, ok := r[res.ID()]; ok {
 		res.Title = old_res.Title
 		res.Meta = old_res.Meta
 	}
 	r[res.ID()] = res
-
-	return res, sysres, true, nil
+	return res, sysres, true
 }
 
 func (ret *ProcessMap) UnmarshalJSON(data []byte) error {


pedroMMM:
@aelsabbahy I created protection against this situation moving forward. I recommend that it is only merged **into this branch** after the correct changes to the `resource/resource_list_genny.go` have been made.

aelsabbahy:
Yeah, I feel silly for missing it in the original PR, sorry @ripienaar.

The fix seems like it would have to modify a bunch of system/resource files to make them consistent.

I started a branch here for this:
https://github.com/aelsabbahy/goss/compare/genny_bug_partial?expand=1

@ripienaar , that should take a lot of the busy work off your hands for the fix. Juse need a clean interface/api for the resource.Validate() method. The way process was handled won't work for file since it has multiple checks.

ripienaar:
weird, the auto generated header wasnt enough for my IDE to notice else I'd have had a popup when I edited the file, oh well.

ripienaar:
So what's the action here? Should I fix this? I know nothing of a go generics system :)

aelsabbahy:
@ripienaar We'll have to normalize on one approach or the other to make everything consistent.

I'm curious on your take on
https://github.com/aelsabbahy/goss/compare/genny_bug_partial?expand=1 (this is incomplete and requires more changes to work properly the validate() do not work correctly)

vs
https://github.com/aelsabbahy/goss/compare/genny_bug_2?expand=1

The first is normalizing all the system/*.go to return error in the same way that system/process.go does.

The later, is making it so system/*.go doesn't raise an error on initializing, but tracks any errors and will raise it when calling the methods on the system resource. NewResource() however will raise an error as needed.

aelsabbahy:
I definitely want this fixed since it will block future development on the project andI don't want to revert your changes since they are a great step in the right direction.

Since you were the driver behind making goss more api driven, I'd love your take on the two directions above.

Basically, should it be:
```
proc, err := NewProcess(...)
```

or:
```
proc := NewProcess(...)
exists, err := proc.Exists() # raises the error from newprocess if there was one
```

aelsabbahy:
One last thing, with the first example above, the following will need to be adjusted to support multiple tests:
```go
       runningf := func() (bool, error) {                                                                                                                                                     
                sysProcess, err := sys.NewProcess(p.Executable, sys, util.Config{})                                                                                                            
                if err != nil {                                                                                                                                                                
                        return false, err                                                                                                                                                      
                }                                                                                                                                                                              
                                                                                                                                                                                               
                return sysProcess.Running()                                                                                                                                                    
        }                                             
```

it works for resource/process.go because there's only one test.. but if there's multiple tests it doesn't scale the way it's currently implemented.

Hope this makes sense.. either way, I'm waiting for your thoughts on this before proceeding.

aelsabbahy:
Put in a little more time, to get both fully compiling and passing tests. Can you look at both these PRs:

* #574
* #575

let me know which approach makes more sense from an API standpoint. The goal is to make sure all the constructors behave the same.

pedroMMM:
> The later, is making it so system/*.go doesn't raise an error on initializing, but tracks any errors and will raise it when calling the methods on the system resource. NewResource() however will raise an error as needed.

@aelsabbahy I am a fan of the 2nd option. 

To be clear you what you are describing:
The error will be stored/returned when it occurs and any following tests on that resource are short-circuited to return the error.

Thank you for making the changes in the code for both options, it made it so much easier to compare 👍

aelsabbahy:
So I'm clear, your vote is for #575

I'll wait to see if @ripienaar votes the same.

pedroMMM:
Correct, my vote is for #575

ripienaar:
yeah neither is awesome, but #575 seems sane

aelsabbahy:
Merged #575 and closed #574

> yeah neither is awesome, but #575 seems sane

Agree, wanted to fix it quickly. I'm open to suggestions or PRs that improve this. Even opening an issue with some thoughts can be helpful since someone else can pick up the work.

Thank you both for your contributions!

PR #572, opened 2020-04-08
Remove autogenerated file from Git and create it as part of dependent build steps

pedroMMM:
We should automatically avoid scenarios like #571 moving forward. Instead of policing the changes to the `resource/resource_list.go` we could just create it as a part of the build process.

##### Checklist

### Description of change

Simply remove the autogenerated file created by Genny and ensure it's marked as a dependency where it is appropriate.


aelsabbahy:
I'd prefer to not remove the file from git since golang is.. well golang and having the files there can be useful, see: https://stackoverflow.com/a/56417892

But it's definitely a great idea to have CI fail if the files aren't matching.

pedroMMM:
> I'd prefer to not remove the file from git since having the files there can be useful, see: https://stackoverflow.com/a/56417892
> 
> It also makes it easier to understand if you're just looking at the repo and don't have it checked out locally and make gen executed.
> 
> But it's definitely a great idea to have CI fail if the files aren't matching.


I took a look at that approach the problem is that Genny doesn't support that out of the box so I would have to do some silly manual diff. That is why I went with a simpler approach. 

That said I can get my hands dirty 👍

aelsabbahy:
A simple hack for CI might be to just do a `make gen` and check if git sees changes and leaving the user workflow as is.

aelsabbahy:
I would love to get rid of that Genny code one day. It feels like a go generics problem, but I wonder if the code can be re-written in a way to avoid it.

pedroMMM:
> A simple hack for CI might be to just do a make gen and check if git sees changes and leaving the user workflow as is.


I thought of that as well, but I worried about Travis polluting its local Git Repo with other things. Let me play with this and I will get something.


> I would love to get rid of that Genny code one day. It feels like a go generics problem, but I wonder if the code can be re-written in a way to avoid it.


That would be great, but I am not the right person to tackle it (at least right now). My Go experience hasn't been to that degree of fanciness. The best I can think of right now is to go down the path of having to implement an Interface for each resource type.

aelsabbahy:
Yeah, unfortunately it doesn't feel like go provides good solutions to this. But maybe it's my ignorance, since most my day experience has been with Python/Ruby.

Looking forward to seeing what you come up with.

ripienaar:
I have wanted to do a huge refactor here so adding new checks can be done via a API ... but wow that’s a big one :)

aelsabbahy:
> I have wanted to do a huge refactor here so adding new checks can be done via a API ... but wow that’s a big one :)

Yeah, I've wanted to do the same with different parts of the code, here's an experiment I have #573

.travis.yml
```diff
@@ -43,7 +45,7 @@ deploy:
     - release/goss-linux-arm.sha256
     - extras/dgoss/dgoss
     - extras/dgoss/dgoss.sha256
-  skip_cleanup: true
+  cleanup: false
```
aelsabbahy, line 48:
```suggestion
  skip_cleanup: true
```
Seems you and I went back and forth on this. Last time this was changed, it prevented me from cutting a release :(

As far as I can tell the documentation still says to use `skip_cleanup`
https://docs.travis-ci.com/user/deployment

`cleanup` seems to be part of the beta deployer, which goss isn't currently using:
https://docs.travis-ci.com/user/deployment-v2

Once deployment-v2 is no longer in beta, we can make this change.

go.mod
```diff
@@ -5,28 +5,17 @@ require (
...
 	github.com/urfave/cli v0.0.0-20161102131801-d86a009f5e13
-	golang.org/x/sys v0.0.0-20181210030007-2a47403f2ae5 // indirect
-	gopkg.in/check.v1 v1.0.0-20190902080502-41f04d3bba15 // indirect
-	gopkg.in/fsnotify.v1 v1.4.7 // indirect
-	gopkg.in/tomb.v1 v1.0.0-20141024135613-dd632973f1e7 // indirect
-	gopkg.in/yaml.v2 v2.2.2
+	gopkg.in/yaml.v2 v2.2.8
 )
 
 go 1.13
```
aelsabbahy, line 21:
```suggestion
go 1.14
```

If we're not verifying against 1.13 we probably should update this to 1.14 to be consistent.

.travis.yml
```diff
@@ -18,6 +18,8 @@ services:
 
 install:
   - go get -u golang.org/x/lint/golint
+  - go get -u github.com/cheekybits/genny
+  - go get -u golang.org/x/tools/cmd/goimports
```
aelsabbahy, line 22: Curious, why goimports?

go.mod
```diff
@@ -5,28 +5,17 @@ require (
 	github.com/aelsabbahy/GOnetstat v0.0.0-20160428114218-edf89f784e08
```
aelsabbahy, line 5: was this a product of: `go get -u ./...` and a `go mod tidy`?

PR #573, opened 2020-04-08
DRAFT: Reflecting on some things...

aelsabbahy:
##### Checklist

### Description of change

Experimenting with using `reflect` to remove some of the verbosity.

Pros:
* Lots of hardcoded code removed/replaced with dynamic code

Cons:
* Reflect can cause runtime errors, we lose the static typing checks
* Reflect is rather difficult to work with/understand (at least to me it is)
* Reflect is slower, but in our context, this doesn't matter
* As it stands today, goss doesn't have a ton of checks.. so the return on investment might not be worth it

### DIFF
--- a/add.go
+++ b/add.go
@@ -1,8 +1,8 @@
 package goss
 
 import (
-	"fmt"
 	"os"
+	"reflect"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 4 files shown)


aelsabbahy:
@ripienaar , @pedroMMM I'd love your thoughts on this.

I'm on the fence and have resisted merging in the original `small_refactor` branch due to the `Cons` listed above.

Guess it's a fun experiment.. that I'm not sure I want to go ahead with.. but I'm curious on others feedbacks =)

goss_config.go
```diff
@@ -49,64 +49,15 @@ func NewGossConfig() *GossConfig {
...
-	for k, v := range g2.Commands {
-		c.Commands[k] = v
-	}
+	v1 := reflect.Indirect(reflect.ValueOf(c))
+	v2 := reflect.ValueOf(g2)
+	for i := 0; i < v1.NumField(); i++ {
+		f1 := v1.Field(i)
+		f2 := v2.Field(i)
+		for _, k := range f2.MapKeys() {
+			f1.SetMapIndex(k, f2.MapIndex(k))
```
pedroMMM, line 58: @aelsabbahy this looks almost taken from the standard library! So I don't think you are working in the wrong direction. 

I, however, would add a type check between `f1` and `f2` to ensure they are the type.

add.go
```diff
@@ -42,50 +42,14 @@ func AddResources(fileName, resourceName string, keys []string, c *util.Config)
...
-		res, err = gossConfig.HTTPs.AppendSysResource(key, sys, config)
-	default:
-		err = fmt.Errorf("undefined resource name: %s", resourceName)
-	}
-
-	if err != nil {
+	v := reflect.ValueOf(gossConfig)
+	f := v.FieldByName(resourceName)
+	fun := f.MethodByName("AppendSysResource")
+	res := fun.Call([]reflect.Value{reflect.ValueOf(key), reflect.ValueOf(sys), reflect.ValueOf(config)})
```
pedroMMM, line 48: @aelsabbahy this one might, however, have been a bit too over-engineered. Couldn't a simple map of resource "appenders" work?

cmd/goss/goss.go
```diff
@@ -230,14 +230,14 @@ func main() {
 					Name:  "package",
 					Usage: "add new package",
 					Action: func(c *cli.Context) error {
-						return goss.AddResources(c.GlobalString("gossfile"), "Package", c.Args(), newRuntimeConfigFromCLI(c))
+						return goss.AddResources(c.GlobalString("gossfile"), "Packages", c.Args(), newRuntimeConfigFromCLI(c))
```
pedroMMM, line 233: `"Packages"` 

This is where I get uncomfortable with magic strings. If you are refactoring this code section it should be converted to an enum like of resources. Which would work well with the map of resource "appenders"

pedroMMM:
- Reflect can cause runtime errors, we lose the static typing checks

As I mentioned on the code we can add some extra checks in there to ensure everything is still ok. It's still a con but it can be mitigated.

- Reflect is rather difficult to work with/understand (at least to me it is)

True reflection isn't the easier thing to follow, so it should be used sparsely. 

- Reflect is slower, but in our context, this doesn't matter

True reflection isn't cheap, so again it should be used sparsely. 

- As it stands today, goss doesn't have a ton of checks.. so the return on investment might not be worth it

The simpler the code base is, the easier for bigger and more ambitions improvements are to take place.   I understand the redundancy to add reflection but it exists for a reason and sometimes is the correct tool.

Issue #538, opened 2020-01-22
Feature Request: Improve gomega matchers


aelsabbahy:
**Describe the feature:**

Improve gomega matchers.

**Describe the solution you'd like**

1. Allow transforms. For example the following should work (fixes #350)
```yaml
kernel-param:
  net.core.somaxconn:
    value:
      gt: 5000
```

2. Add more gomega matchers, for example, the matchallelements matcher for slices.
https://onsi.github.io/gomega/#gstruct-testing-complex-data-types

pedroMMM:
@aelsabbahy besides `matchallelements` any other matchers you have in mind? 

Maybe the 2 requests should be broken into 2 different issues/PRs? If they are broken apart this could be easier to create and review as well.

aelsabbahy:
Yeah, they probably should be broken apart into two tickets and your ticket is the third enhancement on the gomega matchers for io.

aelsabbahy:
@pedroMMM Finally got some time to hammer out my thoughts on this.. feels good since I've wanted to make progress on this concept forever now.

Check out the following PR: #576, take a look at the examples/ folder.

It works.. but unit tests are broken and the code is quickly thrown together and can use some love. I probably won't be available next week.. but if you're interested in this.. feel free to take my branch and run with it =)

Issue #483, opened 2019-08-29
Expose actual command stdout / stderr on the goss test result


colearendt:
It would be nice to be able to opt into exposing the `stdout` that is compared to a regex / etc. in a `command` resource. This is a super trivial example, but suppose I wanted to test how many files are in a folder, and ensure that the number is greater than 3:

```
  "NUMAPP=$(expr $(ls -l /apps | wc -l) - 1); if [ $NUMAPP -ge 3 ]; then echo success: $NUMAPP; else echo fail: $NUMAPP; fi":
    title: apps_deployed_gt_3
    exit-status: 0
    stdout: [
      "/^success/"
    ]
    stderr: []
```

It would be nice if the stdout (i.e. `success: 4`) and/or stderr could be returned in the lengthy / verbose test output (i.e. JSON, documentation, etc.).

aelsabbahy:
I have a research item that may indirectly fix this, if my research is successful. I'll create a ticket for the actual implementation details and link this ticket to it.

Leaving this open for now, but not marking it as approved since the solution might be a consequence of another enhancement.

shan100github:
May be https://github.com/aelsabbahy/goss/issues/559 request is also similar ?

pedroMMM:
Yes, #559 is pretty much the same, I didn't realize that an FR as already placed when I asked @shan100github to create one. Ups 😞 

At least we have more attention to this FR now.


> I have a research item that may indirectly fix this, if my research is successful. I'll create a ticket for the actual implementation details and link this ticket to it.
> 
> Leaving this open for now, but not marking it as approved since the solution might be a consequence of another enhancement.

@aelsabbahy  did you made any progress there? If not we could move this discussion forward.

aelsabbahy:
@pedroMMM I think this PR (when ready) will indirectly fix the issue: #576 

With that, doing a string comparison will print out why the strings didn't match (expected vs result).

Only the io.reader I scan mode will not print out the contents.

On a side note, once that's implemented the test may be re-written as:
```yaml
ls -l /apps | wc -l:
    title: apps_deployed_gt_3
    exit-status: 0
    stdout:
      ge: 3
```

colearendt:
Bump 😄

aelsabbahy:
@colearendt How does #576 Look to you?

Once that's merged in, the functionality will look as follows:
```
command:
  echo "hello world":
    exit-status: 0
    stdout:
    - goodbye world
```

```
$ goss v
.F

[...]

PR #576, opened 2020-04-10
POC/DRAFT: Add transforms

aelsabbahy:
This is to showcase some of the changes I have in mind for #538 to do a major enhancement on the goss/gomega language. Check out the examples folder for an idea of where I'm going with this.

This is horrible code that needs to be cleaned up, just tried to get a quick braindump out there since the code explains it a lot better than my ticket.

For usage examples take a look at this:
https://github.com/aelsabbahy/goss/tree/add_transforms/examples


paskal:
Hi there, I'm interested in having a fix for #483 and I would strongly prefer contributing to upstream istead of mainating my own fork. If I understand correctly, this MR is a potential fix: is there any way I could assist you in finishing it?

aelsabbahy:
This PR is a potential fix in that it allows direct string matching of the command output. Rather than only limiting it to the current pattern matcher logic.

Here's an example with some work from this PR:
```yaml
command:
  test_exec:
    exec: 'echo -e "This is an example output of a command\nGoss will print this in full when using the new string matcher\nas opposed to using the pattern matcher"'
    exit-status: 0
    stdout:
    - This is an example output of a command, broken
    - Goss will print this in full when using the new string matcher
    - as opposed to using the pattern matcher
    stderr: []
    timeout: 10000
  test_exec2:
    exec: 'echo -e "This is an example output of a command\nGoss will print this in full when using the new string matcher\nas opposed to using the pattern matcher"'
    exit-status: 0
    stdout: |
      This is an example output of a command, broken
      Goss will print this in full when using the new string matcher
      as opposed to using the pattern matcher
    stderr: []
    timeout: 10000
```


notice test_exec2 is testing stdout against a string rather than an array. The output of running this in the new version of goss will be:
```
.F..F.

Failures/Skipped:

Command: test_exec: stdout:
Expected
    {} # not implemented yet, will probably say "io.Reader{}", but it will not contain the command output
to contain patterns
    ["This is an example output of a command, broken","Goss will print this in full when using the new string matcher","as opposed to using the pattern matcher"]
the missing elements were
    ["This is an example output of a command, broken"]

Command: test_exec2: stdout:
Expected
    "This is an example output of a command\nGoss will print this in full when using the new string matcher\nas opposed to using the pattern matcher\n"
to equal
    "This is an example output of a command, broken\nGoss will print this in full when using the new string matcher\nas opposed to using the pattern matcher\n"

Total Duration: 0.002s
Count: 6, Failed: 2, Skipped: 0
```

This PR will be a pretty major refactor of how goss handles pattern matching and how results are output. It is a breaking change so it will be in goss v0.4.0.. and since v0.4.0 will be a breaking change, there's a few other breaking changes that I would like to have in there prior to release.

Here's my todo.md (and progress) for v0.4.0:
```markdown
# TODO:

[X] Find way to get private unexported values from gomega matchers
[X] Implement gomega matcher for validate contains
[X] Remove old validatecontains code
[X] implement gjson transform
[X] Add equal matcher
[X] Use benumerically as defualt numbers
[X] Add toNumeric transformer
[-] create new testresult object
[x] Move all matcher data outside of matchers and into testresult object
[ ] Move all formatting related to testresult into outputs/
[-] Implement clean short failure messages for matchers
[ ] Check have-key message with matcher
[ ] Write docs
[ ] github.com/sanity-io/litter?
[ ] And matcher indent is funky/incorrect
[ ] get rid of have-key-with-value or dry out have-key-with-value?

# optional
[ ] catch .(string) type errors in gomega.go (if input is not controlled)
[ ] Implement have-patterns for array
[ ] all matcher potentially
[ ] investigate why string diff does not trigger for equal matcher
[ ] register matchers using init?

# Other v0.4.0 stories, since they potentially introduce breaking changes
[ ] psutil vs go-ps, possibly switch to the former
[ ] process, comm vs args. Can we match on either or with the psutil library?
[ ] rpm expand full version
[ ] rpm/deb version matchers? create generic generic version of semver.Range that can be used with a version interface. Create wrappers around 3rd party rpm/deb comparison libraries to honor version interface
```

Let me know if there's something in there that you're interested in helping out with. If I had to guess I'm about ~60% done with this PR. However, I haven't even started on any of the other v0.4.0 work.

paskal:
I'll look into these early next week and will either do something or ask more questions then. Thank you for the detailed answer!

aelsabbahy:
Awesome, please reach out to me before putting in work/time, I want to avoid a situation where you do work I already completed.


Also, is the string matcher change what you're looking for with regards to #483?

The reason it can't be done with patterns is because goss reads a file line by line. This is apparent if you run `goss v` on a GB text file or command output, the memory usage stays very low. Using string matcher would read the results into memory, but will have the benefit of displaying it.

paskal:
I don't know yet how I would approach as the only thing I know for now is that delivering more information to the affected asset owner is a necessity. If I understand you correctly and it will work with text match but not with regexp, I am fine with it as it's infinitely better than the current situation which doesn't seem to have a workaround (at least I'm not aware of it).

I think I'll start with looking into this PR and figuring out if I can help with it, and then going trough the list and pinging you in PRs\issues.

paskal:
My apologies for the delay, I don't know if I'll have time to look into this as I got a lot of other stuff on hands. @nloginov is interested in having this in place as well and will likely try to help you at some point in time.

aelsabbahy:
@petemounce Tagging you on here since you are creating a bunch of PRs currently.

This is a major work stream that I'm working on for goss v4. I would love your feedback on it.

The manual.md changes and this file explain the differences:
examples/goss_awesome_gomega.yaml


I will most likely create a v4 branch, this + a few other PRs will be merged into it and hopefully I can get v4 out the door within a month. This is a fairly large re-write and may impact some of the PRs you have out there.

docs/manual.md
```diff
@@ -807,91 +805,294 @@ user:
...
+KernelParam: net.core.somaxconn: value:
+Expected
+    128
+to be >
+    200
+the transform chain was
+    [{"to-numeric":{}}]
+
+Total Duration: 0.001s
+Count: 1, Failed: 1, Skipped: 0
```
petemounce, line 936: As a user, I would find this easier to interpret if
* the front of the messaging included the input value and type
* the end of the messaging showed the desired output type

So, sketching something like this?

```suggestion
KernelParam: net.core.somaxconn: value:
Expected
    (string)  128
to be >
    (numeric) 200
the transform chain was
    [{"to-numeric":{ input: "128" }}]

Total Duration: 0.001s
Count: 1, Failed: 0, Skipped: 0, Error: 1
```

I put the types ahead of the values since the width of the type-string-names is predictable, and then the lines can be padded out to the longest one so that the values are all going to appear starting from the same column, making them easier to eyeball.

I added `Error: 1` because it feels to me that a failure to transform is an internal error as opposed to a failed assertion. That struck me as something I'd like to know separately, since it means there's something wrong in how I've defined my assertions, rather than a for-sure assertion-failure.

docs/manual.md
```diff
@@ -807,91 +805,294 @@ user:
...
+```
+
+Kernel-param (failing) test with transform:
+```yaml
+kernel-param:
+  net.core.somaxconn:
+      value: {gt: 200}
+```
+
+When a transformed test fails, it will detail the transformers used, the `-o include_raw` option can be used to include the raw, untransformed attribute value:
```
petemounce, line 920: This means the tests would need to be re-run to get that extra information. As a user, I would prefer that if a test fails, all of the possible debugging information were given to me so I can have it up-front. Not doing that
* in the best case, costs me the time to re-run the tests
    * but this is harder on machines that are remote
* in the worst case, costs me a lot of time to re-run the tests multiple times in the event of chasing an intermittent failure (much more likely in machine state testing, and more likely again if the test touches an external dependency like `addr`, `http` etc).

docs/manual.md
```diff
@@ -807,91 +805,294 @@ user:
...
+For example, kernel-param attribute returns a string, however, it can be tested using numeric comparisons:
+
+Example kernel-param test:
+```yaml
+kernel-param:
+  net.core.somaxconn:
+      value: "128"
+```
+
+Kernel-param (failing) test with transform:
```
petemounce, line 913:
```suggestion
Example (failing) kernel-param test with transform:
```

docs/manual.md
petemounce: This suggestion accidentally also (I think?) removes the need for the `-o include_raw`, since it includes the same information with no extra output lines needed.

docs/manual.md
```diff
@@ -807,91 +805,294 @@ user:
...
-### Examples
+Goss supports advanced matchers by converting YAML input to [gomega](https://onsi.github.io/gomega/) matchers.
+
+#### String Matchers
 
-Validate that user `nobody` has a `uid` that is less than `500` and that they are **only** a member of the `nobody` group.
+These will convert the system attribute to a string prior to matching.
 
+* `'55'` - Checks that the numeric is "55" when converted to string
+* `have-prefix: pre` - Checks if string starts with "pre"
```
petemounce, line 968: This works because the default type in YAML is a string, right?

docs/manual.md
```diff
@@ -807,91 +805,294 @@ user:
...
+        - 42
+        - 42.0
+        - gt: 40
+        - lt: 45
+```
+
+#### Array matchers
+
+These will convert the system attribute to an array prior to matching. Strings are split on "\n"
+
```
petemounce, line 1010:
```suggestion
```

docs/manual.md
```diff
@@ -807,91 +805,294 @@ user:
...
+
+* `contain-element: matcher` - Checks if the array contains an element that passes the matcher
+* `contain-elements: [matcher, ...]` - checks if the array is a superset of the provided matchers
+* `[matcher, ...]` - same as above
+* `equal: [value, ...]` - Checks if the array is exactly equal to provided array
+* `consist-of: [matcher, ...]` - Checks if the array consists of the provided matchers (order does not matter)
```
petemounce, line 1016: Is it to late to rename `consist-of` to `satisfies`?

petemounce, line 1016: I haven't, from this and the example below, figured out where I would use `contain-element` vs `consist-of`. I find it difficult to understand partly because both of them allow matchers to be present.

I can think of wanting to:

* assert that an array contains particular elements/matches (without caring about ordering)
* assert that an array contains a sequence of particular elements/matches (e.g. contains a then any-single-element then b, but that sequence can be anywhere in the array)
  * same but a then not-b then b
* assert that an array contents matches sequence x, y, z exactly

I think I'm struggling partly because the doc prose is more of the reference style (which I find quite abstract, though of course that's highly subjective!) rather than how-to/use-case style. I think if it contained more worked examples I'd have an easier time.

petemounce (commented):
> @petemounce Tagging you on here since you are creating a bunch of PRs currently.
> 
> This is a major work stream that I'm working on for goss v4. I would love your feedback on it.
> 
> The manual.md changes and this file explain the differences:
> examples/goss_awesome_gomega.yaml
> 
> I will most likely create a v4 branch, this + a few other PRs will be merged into it and hopefully I can get v4 out the door within a month. This is a fairly large re-write and may impact some of the PRs you have out there.

I've reviewed just the docs in this pass; this is a pretty massive PR to feed back on. Is there any possibility of slicing it up into smaller chunks? 

What kinds of feedback are you after? UX/requirements/design/implementation/coding style/tests/something else?

If I've understood the intent - this is so that inputs can be converted into the appropriate types to allow assertion with a richer set of matchers. Is that right?

docs/manual.md
```diff
@@ -807,91 +805,294 @@ user:
...
-    - found
-    - /fou.d/
-    - "\\!alsofound"
-    - "!missing"
-    - "!/mis.ing/"
+      and:
+        - contain-element: "foo"
+        - not: {contain-element: "bar"}
+        - contain-element: {match-regexp: "[Gg]oss"}
+
```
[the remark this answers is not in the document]
aelsabbahy: Was this a typo on your end? github UI isn't clear on what this change actually is. Looks to be adding a `-`?

docs/manual.md
aelsabbahy: Gonna defer this conversation to regular github comments to make it easier to track. I'll detail the different use-cases there.

docs/manual.md
aelsabbahy: you mean `have-prefix: pre` vs `have-prefix: "pre"` if so, yes.. that's yaml-ism.

docs/manual.md
aelsabbahy: again, this looks like `-` being suggested as an addition.. is that correct?

docs/manual.md
aelsabbahy: consist-of has been around since 2016. Maybe as part of 4 or a 4.X release I can add other aliases, but I prefer not breaking the old interface.

aelsabbahy: `contain-element` is there since 2016. This PR adds `contain-elements` which is really just a subset check.

For your scenarios:
> assert that an array contains particular elements/matches (without caring about ordering)

contain-element / contain-elements / ["a", "b"] (this is an alias for contain-elements) will satisfy this

> assert that an array contents matches sequence x, y, z exactly

equal

> assert that an array contains a sequence of particular elements/matches (e.g. contains a then any-single-element then b, but that sequence can be anywhere in the array)
same but a then not-b then b

I don't think this is supported today, sounds like a subset of matchers where order matters. I'm curious about the real-world use-case of this.. can be a new matcher that's added in 4.X

> I think if it contained more worked examples I'd have an easier time.

I love this suggestion, I wonder if it should be:
1. A new section in manual.md
2. A new doc that's linked to from manual.md (examples.md?)
3. Just an example folder(s) of yaml files. Could use either the `matching` or maybe real world examples?

Thoughts?

aelsabbahy:
I like the input suggestion. The problem that I've been having is striking a balance between noisy default output and useful default outputs.

```
$ cat goss.yaml
matching:
  example:
    content: '{"foo": "bar", "moo" {"nested": "cow"}, "count": "15"}'
    matches:
      gjson:
        moo.nested: cow
        foo: {have-prefix: b}
        count: {le: 2}  # this will fail
        '@this': {have-key: "foo"}
        moo:
          and:
            - {have-key: "nested"}
            - {not: {have-key: "nested2"}}
```

And the different executions:
```
$ goss v
F

Failures/Skipped:

Matching: example: matches:
Expected
    15
to be <=
    2
the transform chain was
    [{"gjson":{"Path":"count"}},{"to-numeric":{}}]

Total Duration: 0.001s
$ goss v -o include_raw
F

Failures/Skipped:

Matching: example: matches:
Expected
    15
to be <=
    2
the transform chain was
    [{"gjson":{"Path":"count"}},{"to-numeric":{}}]
the raw value was
    "{\"foo\": \"bar\", \"moo\" {\"nested\": \"cow\"}, \"count\": \"15\"}"

Total Duration: 0.001s
$ goss v --format tap
1..1
not ok 1 - Matching: example: matches: Expected "15" to be <= 2 the transform chain was [{"gjson":{"Path":"count"}},{"to-numeric":{}}]
$ goss v --format tap -o include_raw
1..1
not ok 1 - Matching: example: matches: Expected "15" to be <= 2 the transform chain was [{"gjson":{"Path":"count"}},{"to-numeric":{}}] the raw value was "{\"foo\": \"bar\", \"moo\" {\"nested\": \"cow\"}, \"count\": \"15\"}"
```

`-o include_raw` vs `-o exclude_raw` is really more about what gets set as the default.

Errors are caught like other goss errors are. That said, I think the tansforms need more clearer error messages and to explain which transform encountered it, since this isn't very clear.
```
$ goss v
F

Failures/Skipped:

example: matches: Error: strconv.ParseFloat: parsing "15x": invalid syntax

Total Duration: 0.000s
```

docs/manual.md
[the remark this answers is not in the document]
petemounce: Yes, it's removing a newline - sorry, nit-pick on my part! I have `pre-commit` set up to automatically do things like that.

docs/manual.md
petemounce: Ah, gotcha. Yeah, definitely not worth a breaking change. If it's possible easily to alias, that might be nice, but ... not that nice.

petemounce: I think a new section of the manual.md would work well.

Would you be interested in a PR that adds https://www.mkdocs.org/ -> github pages for documentation?

petemounce:
> I like the input suggestion. The problem that I've been having is striking a balance between noisy default output and useful default outputs.

My preference is for an error to tell me all of the inputs and all of the outputs, so that I don't need to run the test again to find them.

I think that for machine-state assertions, that's more-than-normally valuable as a default due to the risk of intermittent failures. I think that because machine-state is not an isolated system; it's fraught with external influencers, dependencies, and those squishy meatbags changing things on what would otherwise be shining clean examples of computational perfection. :)

> `-o include_raw` vs `-o exclude_raw` is really more about what gets set as the default. I initially had raw included by default, but got the opposite feedback. Need a good way to poll goss users on topics like this, any suggestions?

Maybe a gitter.im or community slack? I don't have any good suggestions, sorry.

aelsabbahy:
@ripienaar Adding you here since you're pretty active with major goss changes as of late. This is probably the defining feature of v0.4.0, would love your input. The major change is the ability to do transforms and compare "15" <= 15, or use `gjson` to extract JSON values from a string (this allows for a much more powerful `command`/`http` tests).

At this point, I think it's just a discussion about UI and sane defaults, but feel free to review any part of this PR.

For example, using tap output and the example from the comment above https://github.com/aelsabbahy/goss/pull/576#issuecomment-689736096

currently the PR defaults to a "quiet" output
```
$ goss v --format tap
1..1
not ok 1 - Matching: example: matches: Expected {"count":"15","foo":"bar","moo":{"nested":"cow"}} to have key matching "foox" the transform chain was [{"gjson":{"Path":"@this"}}]
```

This can be changed by using `-o include_raw` to make it more verbose and include the "Raw" input:
```
$ goss v --format tap -o include_raw
1..1
not ok 1 - Matching: example: matches: Expected {"count":"15","foo":"bar","moo":{"nested":"cow"}} to have key matching "foox" the transform chain was [{"gjson":{"Path":"@this"}}] the raw value was "{\"foo\": \"bar\", \"moo\" {\"nested\": \"cow\"}, \"count\": \"15\"}"
```

@petemounce is proposing removing the include_raw and just having the transform chain include the "input" as part of it.
```
$ goss v --format tap                                                                                                                                                                          
1..1                                                                                                                                                                                           
not ok 1 - Matching: example: matches: Expected {"count":"15","foo":"bar","moo":{"nested":"cow"}} to have key matching "foox" the transform chain was [{"gjson":{"Path":"@this","Input":"{\"foo\": \"bar\", \"moo\" {\"nested\": \"cow\"}"}}] 
```

aelsabbahy:
Close enough.. will tackle issues in the v4 branch.

aelsabbahy:
@petemounce This is very close now, I think..

The failure is in the windows test:

```
...F.F..
Failures/Skipped:
Command: wrap a powershell with quotes - expect 0 because travis does not restrict anonymous logins: stdout:
Error
    HavePatterns matcher expects patterns to be a string. got: 
        <int>: 0
Command: wrap a powershell - expect 0 because travis does not restrict anonymous logins: stdout:
Error
    HavePatterns matcher expects patterns to be a string. got: 
        <int>: 0
Total Duration: 0.848s
Count: 8, Failed: 2, Skipped: 0
```

Any objection to this change/fix? https://github.com/aelsabbahy/goss/pull/576/commits/ac0e1ae65a419885ff0fd6aa967c6079ec310442

PR #579, opened 2020-04-25
DGOSS: Honor ${GOSS_FILE} when copying back from container to host

jgeorgeson:
##### Checklist

### Description of change

While testing docker images with a custom `GOSS_FILE` setting, I found _dgoss_ was copying the modified file back to `goss.yaml` instead of `${GOSS_FILE}`.

Additionally the images I was testing were derived from a minimal Alpine linux image, which basically just has `/bin/sh` and `/bin/ash` as symlinks to busybox, so I also changed the shell spec so dgoss would run there and didn't encounter any POSIX compatibility issues.

### DIFF
--- a/extras/dgoss/dgoss
+++ b/extras/dgoss/dgoss
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 
 set -e
 
--- a/extras/dgoss/dgoss
+++ b/extras/dgoss/dgoss
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 set -e
 
--- a/extras/dgoss/dgoss
+++ b/extras/dgoss/dgoss
@@ -60,8 +60,8 @@ run(){
 get_docker_file() {
     if docker exec "$id" sh -c "test -e $1" > /dev/null;then
         mkdir -p "${GOSS_FILES_PATH}"
-        info "Copied '$1' from container to '${GOSS_FILES_PATH}'"
-        docker cp "$id:$1" "${GOSS_FILES_PATH}"
+        info "Copied '$1' from container to '${GOSS_FILES_PATH}/${GOSS_FILE:-goss.yaml}'"
+        docker cp "$id:$1" "${GOSS_FILES_PATH}/${GOSS_FILE:-goss.yaml}"
     fi
 }
 


aelsabbahy:
Goss_file change makes sense.

For the shebang are you sure `sh` supports double brackets? I am fairly certain it doesn't, but haven't used pure posix-sh in ages.

jgeorgeson:
I would have expected an immediate failure if there was unsupported syntax but everything ran fine for me with the shebang change.

Looking at https://busybox.net/downloads/BusyBox.html they do list `[[` as a supported function. And on https://www.busybox.net/ it appears to have been included since 1.26

> test: make [ and [[ forms individually selectable

I don't see Busybox version mentioned on the Alpine release notes prior to 3.9, but Alpine 3.9 included Busybox 1.29

https://alpinelinux.org/posts/Alpine-3.9.0-released.html

I  don't have Docker available at the moment to test the Alpine images but installing Busybox 1.30.2 and testing it seems to work as expected

```
4/25/2020 4:26:32 PM Justin : ~
> busybox cat --help
BusyBox v1.32.0-FRP-3445-g10e14d5eb (2020-04-11 10:50:43 BST) multi-call binary

Usage: cat [-nbvteA] [FILE]...

Print FILEs to stdout

        -n      Number output lines
        -b      Number nonempty lines
        -v      Show nonprinting characters as ^x or M-x
        -t      ...and tabs as ^I
        -e      ...and end lines with $
        -A      Same as -vte
4/25/2020 4:26:34 PM Justin : ~
> busybox sh
C:/Users/Justin # [[ "foo" == "foo" ]] && echo foo
foo
C:/Users/Justin # [[ "foo" != "foo" ]] && echo foo
C:/Users/Justin # ls .gitconfig*
.gitconfig
C:/Users/Justin # [[ -f .gitconfig ]] && echo foo
foo
C:/Users/Justin # [[ -f .gitconfig2 ]] && echo foo
C:/Users/Justin #

**Comment by aelsabbahy** [comment]
Running the script with `/bin/sh` through https://www.shellcheck.net/ or `shellcheck` cli it prints the following:
```
In extras/dgoss/dgoss line 23:
    if [[ $id ]];then
       ^-- SC2039: In POSIX sh, [[ ]] is undefined.
```

https://github.com/koalaman/shellcheck/wiki/SC2039

Some `sh` implementations are more than just POSIX, but not all:
```
$ docker run -it --rm alpine:3.9 sh -c '[[ "foo" ]] && echo "works"'
works
$ docker run -it --rm debian:jessie sh -c '[[ "foo" ]] && echo "works"'
sh: 1: [[: not found
```

jgeorgeson:
fair point, reverted the shebang change.

aelsabbahy:
Awesome, thanks for the contribution!

PR #582, opened 2020-04-29
dgoss - Collect the logs regardless of GOSS_FILE_STRATEGY

Agares:
##### Checklist

### Description of change
Previously the logs were only collected if the GOSS_FILE_STRATEGY was set to "mount", with this change they'll be collected also for "cp".


aelsabbahy:
Looks great, thanks for the contribution!

PR #586, opened 2020-05-11
Add some bash helpers

petemounce:
This is aimed at making the integration-test script (and other scripts) easier to work with by giving them some consistent baseline behaviour

* automatic bash strict mode - sets options for -e -u and -o pipefail
* places execution inside repo-root, to avoid the reader needing to keep track of current working directory. Side benefit - scripts will be runnable from any directory
* adds some logging helpers
* adds a traceback function so it's possible to see which line an error came from

The intention of this PR:
* [x] Migrate each script to use this standard include
* [x] Replace `set -x` where found in existing scripts with 
    * `log_*` usage
    * `trap` usage that will record error output and place it centre stage

I may also refactor `integration-tests/test.sh` to separate the docker-environment-setup part from the run-tests part, so that's clearer. I don't know whether that would "fit" inside this PR; this PR is more about standardising behaviour inside scripts.

##### Checklist

### DIFF
--- a/ci/lib/setup.sh
+++ b/ci/lib/setup.sh
@@ -1,0 +1,81 @@
+# configure cwd, vars and logging
+_setup_env() {
+  # -ET: propagate DEBUG/RETURN/ERR traps to functions and subshells
+  set -ET
+  # exit on unhandled error
+  set -o errexit
+  # exit on unset variable
+  set -o nounset
+  # pipefail: any failure in a pipe causes the pipe to fail
+  set -o pipefail
+
+  if [[ -n "${SCRIPT_DEBUG:-}" ]]; then
+    set -o xtrace
+    # http://www.skybert.net/bash/debugging-bash-scripts-on-the-command-line/
+    export PS4='# ${BASH_SOURCE}:${LINENO}: ${FUNCNAME[0]:-}() - [${SHLVL},${BASH_SUBSHELL},$?] '
+  fi
+  trap _err_trap ERR
+  # shellcheck disable=SC2034
+  # START_DIR is used elsewhere.
+  START_DIR="${PWD}"
+  export START_DIR
+  readonly SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[2]}")" && pwd)"
+  readonly TOP_SCRIPT="${SCRIPT_DIR}/$(basename "${BASH_SOURCE[2]}")"
+  if [[ -z "${SCRIPT_DIR}" ]]; then
+    echo >&2 -e "setup.sh:\tFailed to determine directory containing executed script."
+    return 1
+  fi
+  if ! cd "$(dirname "${BASH_SOURCE[0]}")/../.."; then
+    echo >&2 -e "setup.sh:\tFailed to cd to repository root"
+    return 1
+  fi
+  REPO_ROOT="$(pwd)"
+  export REPO_ROOT
+  if ! source ci/lib/log.sh; then
+    echo >&2 -e "setup.sh:\tFailed to source logging library"
+    return 1
+  fi
+}
+
+_err_trap() {
+  local err=$?
+  local cmd="${BASH_COMMAND:-}"
+  # Disable echoing all commands as this makes the traceback really hard to follow
+  set +x
+  if [[ -n "${SKIP_BASH_STACKTRACE:-}" ]]; then
+    log_debug "SKIP_BASH_STACKTRACE was set to something; silencing bash stack-trace."
+    exit "${err}"
+  fi
+
+  echo >&2 "panic: uncaught error" 1>&2
+  print_traceback 1
+  echo >&2 "${cmd} exited ${err}" 1>&2
+}
+
+_setup_constants() {
+  export EXIT_SUCCESS=0
+  export EXIT_INVALID_ARGUMENT=66
+  export EXIT_FAILED_TO_SOURCE=67
+  export EXIT_FAILED_TO_CD=68
+  export EXIT_FAILED_AFTER_RETRY=69
+  export EXIT_NOT_FOUND=70
+}
+
+# Print traceback of call stack, starting from the call location.
+# An optional argument can specify how many additional stack frames to skip.
+print_traceback() {
+  local skip=${1:-0}
+  local start=$((skip + 1))
+  local end=${#BASH_SOURCE[@]}
+  local curr=0
+  echo >&2 "Traceback (most recent call first):" 1>&2
+  for ((curr = start; curr < end; curr++)); do
+    local prev=$((curr - 1))
+    local func="${FUNCNAME[$curr]}"
+    local file="${BASH_SOURCE[$curr]}"
+    local line="${BASH_LINENO[$prev]}"
+    echo >&2 "  at ${file}:${line} in ${func}()" 1>&2
+  done
+}
+
+_setup_env || exit $?
--- a/integration-tests/test.sh
+++ b/integration-tests/test.sh
@@ -1,9 +1,12 @@
 #!/usr/bin/env bash
+# shellcheck source=../ci/lib/setup.sh
+source "$(dirname "${BASH_SOURCE[0]}")/../ci/lib/setup.sh" || exit 67
+# preserve current behaviour
+set -x
 
-set -xeu
+os="${1:?"Need OS as 1st arg. e.g. alpine arch centos7 precise wheezy"}"
+arch="${2:?"Need arch as 2nd arg. e.g. amd64 386"}"
 
-os=$1
-arch=$2
 vars_inline="{inline: bar, overwrite: bar}"
 
 seccomp_opts() {
@@ -14,6 +17,9 @@ seccomp_opts() {
     echo '--security-opt seccomp:unconfined'
   fi
 }
+
+# setup places us inside repo-root; this preserves current behaviour with least change.
+cd integration-tests
 
 cp "../release/goss-linux-$arch" "goss/$os/"
 # Run build if Dockerfile has changed but hasn't been pushed to dockerhub
--- a/ci/lib/log.sh
+++ b/ci/lib/log.sh
@@ -1,0 +1,52 @@
+#!/usr/bin/env bash
+
+# comment out the unused-ones so far until they're needed. Otherwise it's a google search to find them again.
+NOCOLOUR='\033[0m'
+RED='\033[0;31m'
+GREEN='\033[0;32m'
+ORANGE='\033[0;33m'
+# BLUE='\033[0;34m'
+# PURPLE='\033[0;35m'
+CYAN='\033[0;36m'
+# LIGHTGRAY='\033[0;37m'
+# DARKGRAY='\033[1;30m'
+LIGHTRED='\033[1;31m'
+LIGHTGREEN='\033[1;32m'
+# YELLOW='\033[1;33m'
+# LIGHTBLUE='\033[1;34m'
+# LIGHTPURPLE='\033[1;35m'
+LIGHTCYAN='\033[1;36m'
+# WHITE='\033[1;37m'
+
+is_ci() {
+  if [[ "${CI:-}" == "true" ]]; then
+    echo "true"
+  else
+    echo "false"
+  fi
+}
+
+log_action() {
+  echo -e "${LIGHTGREEN}${*}${NOCOLOUR}" >&2
+}
+log_warn() {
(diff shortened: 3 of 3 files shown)


petemounce:
Ping @aelsabbahy for review. This is the thinnest slice to introduce it and show that it does not break CI. If you're ok with this, I'd make changes to other scripts to incorporate it.

ci/lib/setup.sh
```diff
@@ -0,0 +1,81 @@
...
+  set -o pipefail
+
+  if [[ -n "${SCRIPT_DEBUG:-}" ]]; then
+    set -o xtrace
+    # http://www.skybert.net/bash/debugging-bash-scripts-on-the-command-line/
+    export PS4='# ${BASH_SOURCE}:${LINENO}: ${FUNCNAME[0]:-}() - [${SHLVL},${BASH_SUBSHELL},$?] '
+  fi
+  trap _err_trap ERR
+  # shellcheck disable=SC2034
+  # START_DIR is used elsewhere.
```
aelsabbahy, line 19: where/how is this used?

petemounce:
> log.sh looks great. However, setup.sh is confusing and hides stuff when set -x is set.
> 
> For example, running a failing `./test.sh precise 386` with set -x, gives this confusing message and hides the debug information fo the trap code.
> 
> ```
> + sleep 1
> ++ docker_exec /goss/precise/goss-linux-386 --vars /goss/vars.yaml --vars-inline '{inline: bar, overwrite: bar}' -g /goss/precise/goss.yaml validate
> ++ docker exec goss_int_test_precise_386 /goss/precise/goss-linux-386 --vars /goss/vars.yaml --vars-inline '{inline: bar, overwrite: bar}' -g /goss/precise/goss.yaml validate
> +++ _err_trap
> +++ local err=1
> +++ local 'cmd=docker exec "$container_name" "$@"'
> +++ set +x
> panic: uncaught error
> Traceback (most recent call first):
>   at ./test.sh:35 in docker_exec()
>   at ./test.sh:50 in main()
> docker exec "$container_name" "$@" exited 1
> panic: uncaught error
> Traceback (most recent call first):
>   at ./test.sh:50 in main()
> docker exec "$container_name" "$@" exited 1
> + out='.......SS......SS..S...................................................F...................................
> 
> Failures/Skipped:
> 
> File: /does/not/exist: exists: skipped
> File: /does/not/exist: contains: skipped
> 
> Service: skippable: enabled: skipped
> Service: skippable: running: skipped
> 
> Command: commandskip: exit-status: skipped
> 
> Service: autofs: running:
> Expected
>     <bool>: false
> to equal
>     <bool>: true
> 
> Total Duration: 1.594s
> Count: 107, Failed: 1, Skipped: 5'
> ++ _err_trap
> ++ local err=1
> ++ local 'cmd=out=$(docker_exec "/goss/$os/goss-linux-$arch" --vars "/goss/vars.yaml" --vars-inline "$vars_inline" -g "/goss/$os/goss.yaml" validate)'
> ++ set +x
> panic: uncaught error
> Traceback (most recent call first):
>   at ./test.sh:50 in main()
> out=$(docker_exec "/goss/$os/goss-linux-$arch" --vars "/goss/vars.yaml" --vars-inline "$vars_inline" -g "/goss/$os/goss.yaml" validate) exited 1
> d1a24ff4e4cf9c290a6fae0d8775ec085f7ccfb6b34d7268f3dbff53e4321567
> ```
> 
> I'm fine with merging this code but not sourcing setup.sh until the bash scripts are cleaned up to the point where they make sense without `set -x`

This first commit was just to integrate it and see CI pass, and gauge whether you were amenable to the include in general.

I take from your reply that you are, if I now go through `test.sh` and remove the need for it being run `set -x` in favour of log statements. I plan to also (whether this PR or followup; depends how it goes) factor `test.sh` so that the docker-setup parts are separate from the run-tests parts, so that it becomes possible (_definitely_ in a followup) to use the run-tests parts nakedly on macOS and Windows CI hosts, and see where that gets me in terms of integration coverage for those platforms vs effort put in.

Does that sound alright? I'll wait to hear before proceeding.

ci/lib/setup.sh
petemounce: Oops, that's in fact _not_ used here, but is over inside a different codebase. Will remove.

aelsabbahy:
Yeah, that sounds good and makes a lot of sense.

The CI process overall needs to be improved. There needs to be a clearer:
* Unit tests
* Integration tests
* Smoke tests

Hopefully then they'll be easier to compose. So that:
* Unit test and smoke test all platforms
* Integration test much smaller set

aelsabbahy:
I assume this is still a work in progress. Correct?

Let me know when it's good to go.

petemounce:
Oh, definitely still work in progress. I don't expect to make progress imminently, though. This is more of a "plinking along in the background" sort of thing.

ci/lib/setup.sh
petemounce: Removed it.

petemounce:
@aelsabbahy I've resumed this and I think it's in a reviewable state.

I _haven't_ tried out the `development/*_images.sh`, but I think everything else is covered because CI uses it.

integration-tests/test.sh
```diff
@@ -15,54 +16,79 @@ seccomp_opts() {
   fi
 }
 
-cp "../release/goss-linux-$arch" "goss/$os/"
-# Run build if Dockerfile has changed but hasn't been pushed to dockerhub
```
petemounce, line 19: I had to add this pushd/popd because the md5 file contains the relative path to the dockerfile, and I'm not sure how to regenerate the md5s (?). I'd be happy to do that, or alternatively leave this the way it is currently.

integration-tests/test.sh
```diff
@@ -15,54 +16,79 @@ seccomp_opts() {
...
+cp "release/goss-linux-${arch}" "integration-tests/goss/${os}/"
+pushd "integration-tests"
 if ! md5sum -c "Dockerfile_${os}.md5"; then
-  docker build -t "aelsabbahy/goss_${os}:latest" - < "Dockerfile_$os"
-# Pull if image doesn't exist locally
-elif ! docker images | grep "aelsabbahy/goss_$os";then
+  log_info "Dockerfile_${os} has changed, rebuilding aelsabbahy/goss_${os}:latest"
+  docker build -t "aelsabbahy/goss_${os}:latest" - < "Dockerfile_${os}"
+elif ! docker images | grep "aelsabbahy/goss_${os}"; then
+  log_info "aelsabbahy/goss_${os} doesn't exist locally; pulling."
```
petemounce, line 25: Sample CI output https://travis-ci.org/github/aelsabbahy/goss/jobs/720718087#L4978 shows up quite prominently.

ci/before-install.sh
```diff
@@ -0,0 +1,17 @@
...
+# shellcheck source=lib/stdlib.bash
+source "$(dirname "${BASH_SOURCE[0]}")/lib/stdlib.bash" || exit 67
+
+os_name="$(go env GOOS)"
+
+if [[ "${os_name}" == "darwin" ]]; then
+  # bash from macOS is too old to have readarray. Install newer version.
+  brew install bash
+elif [[ "${os_name}" == "linux" ]]; then
+  echo "Nothing to do for linux."
```
petemounce, line 11: I noticed the docker-version test inside seccomp_options gives output that looks like an error, which lead me to see that the version of docker by default in travis is truly ancient, so I thought I'd leave space in here to install it, and so I extracted from .travis.yml to here.

integration-tests/test.sh
```diff
@@ -1,68 +1,95 @@
...
-os=$1
-arch=$2
 vars_inline="{inline: bar, overwrite: bar}"
 
 seccomp_opts() {
   local docker_ver minor_ver
   docker_ver=$(docker version -f '{{.Client.Version}}')
+  log_info "docker version: ${docker_ver}"
   minor_ver=$(cut -d'.' -f2 <<<$docker_ver)
   if ((minor_ver>=10)); then
```
aelsabbahy, line 15:
```suggestion
  if ((10#$minor_ver>=10)); then
```

Should fix the whole: value too great for base error
https://stackoverflow.com/questions/12821715/convert-string-into-integer-in-bash-script-leading-zero-number-error/12821845#12821845

development/build_images.sh
```diff
@@ -1,13 +1,9 @@
...
+source "$(dirname "${BASH_SOURCE[0]}")/../ci/lib/stdlib.bash" || exit 67
 
-set -xeu
-
-SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-INTEGRATION_TEST_DIR="$SCRIPT_DIR/../integration-tests/"
-
-
-for docker_file in $INTEGRATION_TEST_DIR/Dockerfile_*; do
+for docker_file in "integration-tests/Dockerfile_*"; do
```
aelsabbahy, line 5: This is a breaking change, the double quotes shouldn't be there

integration-tests/test.sh
aelsabbahy: It either needs to exist in both or be removed from both.

I build the images to speed up CI, but CI will fallback on building the images itself if the checksum doesn't match.

As long as the publish and this script are in sync or expect the same relative path, this mechanism will work.

development/push_images.sh
```diff
@@ -1,18 +1,15 @@
...
 images=$(docker images | grep '^aelsabbahy/goss_.*latest' | awk '$0=$1')
 
 # Use md5sum to determine if CI needs to do a docker build
-pushd "$SCRIPT_DIR/../integration-tests";
-  for dockerfile in Dockerfile_*;do
-    [[ $dockerfile == *.md5 ]] && continue
-    md5sum "$dockerfile" > "${dockerfile}.md5"
-  done
-popd
+for dockerfile in "integration-tests/Dockerfile_*"; do
```
aelsabbahy, line 8: same.. double quotes should be removed

development/build_images.sh
```diff
@@ -1,13 +1,9 @@
 #!/usr/bin/env bash
+# shellcheck source=../ci/lib/stdlib.bash
+source "$(dirname "${BASH_SOURCE[0]}")/../ci/lib/stdlib.bash" || exit 67
 
-set -xeu
-
-SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-INTEGRATION_TEST_DIR="$SCRIPT_DIR/../integration-tests/"
```
aelsabbahy, line 6: I'm guessing this is no longer needed because the helper? stdlib.bash? (haven't fully combed through the code yet)

integration-tests/test.sh
petemounce: 🤦 thanks!

development/build_images.sh
petemounce: Yes, that's right. I can mark up the stdlib.bash with some line comments to talk through what's going on, if you like?

aelsabbahy:
Making sure I didn't miss anything here.. this is still WIP, correct?

Which reminds me at some point I really want to take an axe to most of this and radically simplify the testing.

Many thanks for all the unit tests you've been adding. Honestly, I'll feel a lot better about the state of goss testing when:
1. dnstest.io goes away, should stand up a dns server as part of integration tests, dnstest.io is a ticking time-bomb, it's an external dependency (user contributed) that could go away any day.
2. Same for httpbin possibly, or any "external" url.
3. shell scripts are radically simplified.. the whole docker testing workflow should be pretty cookiecutter. This PR is a great step in that direction.
4. Need to migrate the travis-ci config to the new pipeline format or whatever it's called, so unit/integration tests are split properly. Or maybe.. look for something other than travis-ci?
5. A lot more unit tests (You have some PRs on this already, much appreciated).

Sorry.. Just thinking out loud.

petemounce:
@aelsabbahy I reckon this is no longer WIP, but ready to merge from my point of view. If you've got more feedback obviously I'll address that ahead of it being merged.

petemounce:
I've rebased.

aelsabbahy:
Let me know if you wanted this PR to be pushed ahead of some of the other ones.

Awesome job on all this by the way. I think it's all very close to being merged/released at this point.

petemounce:
@aelsabbahy merging this ahead of the others would be helpful, actually; I'd be able to rebase and get better log output to figure out the failure. So yes, if you wouldn't mind, please proceed with this one now?

> Awesome job on all this by the way. I think it's all very close to being merged/released at this point.

My pleasure. I'm just building on top of your existing great stuff here :)

Issue #447, opened 2019-05-22
Provide checksums for public binaries


taurus-forever:
Hi,

goss is available for downloading on https://github.com/aelsabbahy/goss/releases
Unfortunately checksums are missing there. Is it possible to publish them there as well?

Tnx!

aelsabbahy:
Checksums are available as of last release, but the path is a little off, see #590 and #591

PR #585, opened 2020-05-10
Multi-OS: Provide binaries

petemounce:
Fixes #551, #405, #406, #403, #448, #385, #380, #298 (or supercedes them)

* [x] See Travis make binaries and pass
    * [x] linux
        * [x] binary
        * [x] lint + short tests
        * [x] integration tests
    * [x] macOS
        * [x] binary
        * [x] lint + short tests
    * [x] Windows
        * [x] binary
        * [x] lint + short tests
* [x] Write OS-support matrix doc per #551
* [x] Link doc from readme

##### Checklist

- [x] `make test-all` (UNIX) passes. CI will also test this
- [x] unit and/or integration tests are included (if applicable)
- [x] documentation is changed or added (if applicable)

### Description of change

Adds binaries for macOS and Windows, plus feature-parity matrix doc / test-plan.

PR #590, opened 2020-05-27
Remove 'release/' from checksum file

diraol:
By having `<sha256> release/goss...` this prevents from automating the
check between the sha and the file, since sha256sum will try to find the
goss file within the `release` directory and usually, while automating goss
installation, we have them both in the same path and run the check from
this path, not from outside a release directory.

Relates to: #447

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,14 +46,12 @@ bench:
 	$(info INFO: Starting build $@)
 	go test -bench=.
 
-
-
 # Pattern rule for platform builds.
 # `subst` substitutes space for -, thus making an array
 # firstword, and word select indexes from said array.
 release/goss-%: $(GO_FILES)
 	CGO_ENABLED=0 GOOS=$(firstword $(subst -, ,$*)) GOARCH=$(word 2, $(subst -, ,$*)) go build -ldflags "-X main.version=$(TRAVIS_TAG) -s -w" -o $@ $(exe)
-	sha256sum $@ | sed 's/release\/goss/goss/g' > $@.sha256
+	cd $(@D) && sha256sum $(@F) > $(@F).sha256sum
 
 release:
 	$(MAKE) clean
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -53,7 +53,7 @@ bench:
 # firstword, and word select indexes from said array.
 release/goss-%: $(GO_FILES)
 	CGO_ENABLED=0 GOOS=$(firstword $(subst -, ,$*)) GOARCH=$(word 2, $(subst -, ,$*)) go build -ldflags "-X main.version=$(TRAVIS_TAG) -s -w" -o $@ $(exe)
-	sha256sum $@ > $@.sha256
+	sha256sum $@ | sed 's/release\/goss/goss/g' > $@.sha256
```
aelsabbahy, line 56: Wonder if this would work to avoid sed
```
cd $(@D) && sha256sum $(@F)
```

diraol: Well, to be really honest I have no clue about your suggestion! :joy: Lacking Make knowledge here.

But I totally understand (and agree) with removing the sed operation, so I'll use your suggestion ;)

diraol: So, just to understand, the "solution" would be:

```suggestion
cd $(@D) && sha256sum $(@F) > $(@D).sha256sum
```
?

aelsabbahy:
Travis-CI failed :( seems there's something in the makefile that's broken.

diraol:
Oh man, sorry, I didn't received Travis-ci failed build feedback.

I guess it was missing a Tab character, I've edited it, let's see. :)

diraol:
Ok, all good now, I guess :)

aelsabbahy:
#585 conflicted with this in a major way (re-wrote a lot of this code).

I ended up creating #592 to fix this.

Thank you for taking the time to implement this, sorry it ended up conflicting with a much bigger change.

diraol:
No problem at all, glad to help :)

PR #597, opened 2020-07-03
Migrate from go-ps to gopsutil for better process detection

### LINKED ISSUE #451: add process misreads process status
For several processes on my RHEL 7.3 server, goss seems to misrecord the statuses of processes... 

As a test, I pipe every process in ps into a goss add process command...
```
for p in $(ps -ef | grep -v '\[' | cut -b49- | cut -d" " -f1 | cut -d: -f1 | rev | cut -d/ -f1 | rev |sort -u); do 
  ../goss a process ${p}
done
```
Apart from the fluff (CMD from ps header, and ps, cut, rev, sort used in the command above), one would expect all these processes to be actual running processes.   Yet, goss reports several of them as **NOT** running.

If I take just systemd processes as an example:
```
ps -ef | grep systemd
root         1     0  0 May21 ?        00:00:51 /usr/lib/systemd/systemd --switched-root --system --deserialize 21
root       898     1  0 May21 ?        00:01:01 /usr/lib/systemd/systemd-journald
root       932     1  0 May21 ?        00:00:00 /usr/lib/systemd/systemd-udevd
dbus      2353     1  0 May21 ?        00:00:23 /usr/bin/dbus-daemon --system --address=systemd: --nofork --nopidfile --systemd-activation
root      2383     1  0 May21 ?        00:00:08 /usr/lib/systemd/systemd-logind
```
We see 4 processes all running since May 21.   If we examine the goss.yaml file we see this:
```
grep -A1 systemd goss.yaml
  systemd:            
    running: true     
  systemd-journald:   
    running: false    
  systemd-logind:     
    running: true     
  systemd-udevd:      
    running: true
```
3 of the 4 are listed as running, but even though systemd-journald is running, it is reported as false...

Whatever is happening is consistent, in the sense that validation will pass confirming "systemd-journald" is not running when goss re-executes.

There are a few other processes (excluding ps, cut, rev, sort from the command that fed goss) from a vendor product on my system that similarly report as running false when they are actually running...   

Any idea what is happening to cause this strange behavior?

[This is using v0.3.7 AMD64]

aelsabbahy:
##### Checklist

### Description of change

Fixes #451


petemounce:
LGTM!

ripienaar:
Hmm any chance this can be done to master as well? go-ps has some issues with newer go

aelsabbahy:
Reason I kept it v4 is because it introduces breaking changes and I'm bundling up all the breaking changes for a "major" release.

I can upgrade to new version of go as part of v4 too.

Curious, what's the go-ps issue with new go?

ripienaar:
I will ask again, someone who is packaging my software on FreeBSD where they use 1.15 already complained but he didnt give me details and is no afk, will see if I can get specifics.

do you have views on a v4 timeline?

ripienaar:
Ah, https://github.com/mitchellh/go-ps/blob/master/process_freebsd.go isnt limited to amd64 only, thats the issue.

ripienaar:
Basically just https://github.com/mitchellh/go-ps/commit/32d76597b274fe874d3e8e89be19315b7b713207#diff-bc4fae79b38e742342740544c00892f0, so I can do a PR against that to bring it up to same if possible, and a follow up here to update your master

aelsabbahy:
Yup sounds good. I can definitely have that in the next minor release.

Thanks!

ns-tghosh:
> Hmm any chance this can be done to master as well? go-ps has some issues with newer go

@aelsabbahy Would this change be merged to master? We have encountered the same issue of the process being misread.

```
# ps -p 13 -o comm,cmd
COMMAND         CMD
verdict_control /usr/bin/python3.8 /usr/local/bin/verdict_controller_server.py --port 5000

# cat process.yaml
process:
  verdict_control:
    running: true

# ../goss -g process.yaml validate
F

Failures/Skipped:

Process: verdict_control: running:
Expected
    false
to equal
    true

Total Duration: 0.027s
```

Goss version and linux distribution details:
```
# ../goss --version
goss version 0.13.18

# lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 16.04.7 LTS
Release:        16.04
Codename:       xenial
```

aelsabbahy:
`0.13.18` was this a copy/paste error or is the binary a custom local build?

I think this might be a new issue since the issue described by @ripienaar only affected freebsd.

Can you open a new issue with this, I assume `goss a process verdict_control` has the same issue?

aelsabbahy:
Out of curiosity, what does `cat /proc/13/comm` say? I wonder if ubuntu's `ps -o comm ..` command doesn't truncate the comm or does some enrichment to it.

PR #599, opened 2020-07-08
Validate checksums in installer

particleflux:
There exist checksums for both `goss` and `dgoss` since version 0.3.11,
so those should also be used to validate the downloads.
This only works since 0.3.13 though, as the filename in the checksum
file changed.

Also use the same version for dgoss than for goss, instead of always
latest master for dgoss.

##### Checklist

_The checklist does not apply, since this is installer code_

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -7,16 +7,15 @@ LATEST_URL="https://github.com/aelsabbahy/goss/releases/latest"
 LATEST_EFFECTIVE=$(curl -s -L -o /dev/null ${LATEST_URL} -w '%{url_effective}')
 LATEST=${LATEST_EFFECTIVE##*/}
 
-DGOSS_VER=$GOSS_VER
 
 if [ -z "$GOSS_VER" ]; then
     GOSS_VER=${GOSS_VER:-$LATEST}
-    DGOSS_VER='master'
 fi
 if [ -z "$GOSS_VER" ]; then
     echo "ERROR: Could not automatically detect latest version, set GOSS_VER env var and re-run"
     exit 1
 fi
+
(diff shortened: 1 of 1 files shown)


aelsabbahy:
This is an interesting pull request. I'll start with explaining the rationale behind the current way it works and we can discuss this further.

Since there's two changes here, I'll split the discussion into two parts.

### dgoss version

The reason the installer pulls the latest version when `$GOSS_VER` isn't specified is because I don't cut goss releases for dgoss changes (or anything in extras/). dgoss is currently around 120 lines of bash a lot of times there are minor changes done that aren't released for a month or two until goss is released.

However, if you specify an explicit `$GOSS_VER`, the installer honors that for dgoss.

### checksum

I'm curious on your reasoning for adding this to the installer. It seems like a good change, but I'm worried about the reason for the change.

To me there's two ways to install goss:
* The installer - not recommended for production boxes
* Manual - production or anywhere important

It feels to me, if a user cares enough about the checksum verification, they should not be using `curl | sh` as an installation method.

particleflux:
> It feels to me, if a user cares enough about the checksum verification, they should not be using curl | sh as an installation method.

Now that is an interesting point, as the installer itself is retrieved from a not really trusted source anyway. Which leads me to think that the verification could easily be patched out. Honestly, I didn't think that far :)
I personally am mostly using the installer for a bunch of CI builds - not the worst case if it went rogue, as it is a throwaway environment anyway. In that case, using the installer in a one-liner install seems simpler than manually installing and verifying.  

Aside from that, for a real verification, signatures would be required.

The only "pro" left for the verification would then be to circumvent download/server issues, which basically should also be catched by the `--version` call later.

PR #603, opened 2020-07-21
Make dgoss support gossfile directives

AlexisGoodfellow:
##### Checklist

- [ x ] `make test-all` (UNIX) passes. CI will also test this

### Description of change

`dgoss` does not support gossfile directives out of the box since it only copies `goss.yaml` and `goss-wait.yaml` into the container. It currently does _not_ copy in any of the other `.yaml` files that could be "imported" with the gossfile directive. 

We can fix this with very little changes by copying the entire directory's contents into the container so that all test files will be present in the container. This is useful to reduce code replication, since it allows test authors to logically partition test cases based on particular OSs/ tools/frameworks and then mix and match the smaller, compose-able test suites based on what each final image is supposed to contain. Besides, if `goss` allows it, `dgoss` should support it.

### DIFF
--- a/extras/dgoss/dgoss
+++ b/extras/dgoss/dgoss
@@ -30,9 +30,7 @@ run(){
     # Copy in goss
     cp "${GOSS_PATH}" "$tmp_dir/goss"
     chmod 755 "$tmp_dir/goss"
-    [[ -e "${GOSS_FILES_PATH}/${GOSS_FILE:-goss.yaml}" ]] && cp "${GOSS_FILES_PATH}/${GOSS_FILE:-goss.yaml}" "$tmp_dir/goss.yaml" && chmod 644 "$tmp_dir/goss.yaml"
-    [[ -e "${GOSS_FILES_PATH}/goss_wait.yaml" ]] && cp "${GOSS_FILES_PATH}/goss_wait.yaml" "$tmp_dir" && chmod 644 "$tmp_dir/goss_wait.yaml"
-    [[ ! -z "${GOSS_VARS}" ]] && [[ -e "${GOSS_FILES_PATH}/${GOSS_VARS}" ]] && cp "${GOSS_FILES_PATH}/${GOSS_VARS}" "$tmp_dir" && chmod 644 "$tmp_dir/${GOSS_VARS}"
+    cp -r "${GOSS_FILES_PATH}" "$tmp_dir/test_files"
 
     # Switch between mount or cp files strategy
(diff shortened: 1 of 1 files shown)


extras/dgoss/dgoss
```diff
@@ -30,9 +30,7 @@ run(){
     # Copy in goss
     cp "${GOSS_PATH}" "$tmp_dir/goss"
     chmod 755 "$tmp_dir/goss"
-    [[ -e "${GOSS_FILES_PATH}/${GOSS_FILE:-goss.yaml}" ]] && cp "${GOSS_FILES_PATH}/${GOSS_FILE:-goss.yaml}" "$tmp_dir/goss.yaml" && chmod 644 "$tmp_dir/goss.yaml"
-    [[ -e "${GOSS_FILES_PATH}/goss_wait.yaml" ]] && cp "${GOSS_FILES_PATH}/goss_wait.yaml" "$tmp_dir" && chmod 644 "$tmp_dir/goss_wait.yaml"
-    [[ ! -z "${GOSS_VARS}" ]] && [[ -e "${GOSS_FILES_PATH}/${GOSS_VARS}" ]] && cp "${GOSS_FILES_PATH}/${GOSS_VARS}" "$tmp_dir" && chmod 644 "$tmp_dir/${GOSS_VARS}"
+    cp -r "${GOSS_FILES_PATH}" "$tmp_dir/test_files"
```
aelsabbahy, line 33: The 644 is important, see: https://github.com/aelsabbahy/goss/pull/494

extras/dgoss/dgoss
```diff
@@ -101,18 +99,18 @@ case "$1" in
...
         fi
         ;;
     edit)
         run "$@"
         info "Run goss add/autoadd to add resources"
         docker exec -it "$id" sh -c 'cd /goss; PATH="/goss:$PATH" exec sh'
-        get_docker_file "/goss/goss.yaml"
-        get_docker_file "/goss/goss_wait.yaml"
-        [[ ! -z "${GOSS_VARS}" ]] && get_docker_file "/goss/${GOSS_VARS}"
+        get_docker_file "/goss/test_files/goss.yaml"
```
aelsabbahy, line 111: This behavior seems inconsistent, since the files copied in and the ones copied out are different.

extras/dgoss/dgoss
aelsabbahy: This also feels heavy handed as it's copying everything in the directory, not every user will only have goss files there.

PR #606, opened 2020-08-24
Add some test coverage to the goss serve command.

petemounce:
##### Checklist

  - [x] 200 OK on passing tests
  - [x] 503 Service Unavailable on failing tests
  - [x] response content-type == application/json when json is chosen as the process-level output format
  - [x] caching works as expected
  - n/a

### Description of change

I'm adding some test-coverage to the `goss serve` sub-command, to act as a safety net ahead of a change I'd like to make to it.

(I'd like to make the output format be possible to choose via the request's content-type header, because in the future I'd like to add a prometheus metrics endpoint in addition to a health-check endpoint. At the moment with output format being process-level rather than request-level, I'd need to run two goss serve processes to achieve that, and I'd prefer to avoid that)

### DIFF
--- a/outputs/structured.go
+++ b/outputs/structured.go
@@ -75,10 +75,6 @@ func (r Structured) Output(w io.Writer, results <-chan []resource.TestResult, st
 
 	fmt.Fprintln(w, string(j))
 
-	if result.Summary.Failed > 0 {
-		return 1
-	}
-
 	return 0
 }
 
--- a/outputs/structured.go
+++ b/outputs/structured.go
@@ -75,6 +75,10 @@ func (r Structured) Output(w io.Writer, results <-chan []resource.TestResult, st
 
 	fmt.Fprintln(w, string(j))
 
+	if result.Summary.Failed > 0 {
+		return 1
+	}
+
 	return 0
 }
 
--- a/serve.go
+++ b/serve.go
@@ -14,15 +14,15 @@ import (
 	"github.com/patrickmn/go-cache"
 )
 
-func Serve(c *util.Config) {
+func Serve(c *util.Config) error {
 	endpoint := c.Endpoint
 	health, err := newHealthHandler(c)
 	if err != nil {
-		log.Fatalf("Error: %v", err)
+		return err
 	}
 	http.Handle(endpoint, health)
 	log.Printf("Starting to listen on: %s", c.ListenAddress)
-	log.Fatal(http.ListenAndServe(c.ListenAddress, nil))
+	return http.ListenAndServe(c.ListenAddress, nil)
 }
 
 func newHealthHandler(c *util.Config) (*healthHandler, error) {
--- a/serve.go
+++ b/serve.go
@@ -2,10 +2,8 @@ package goss
 
 import (
 	"bytes"
-	"fmt"
 	"log"
 	"net/http"
-	"os"
 	"sync"
 	"time"
 
@@ -18,22 +16,30 @@ import (
 
 func Serve(c *util.Config) {
 	endpoint := c.Endpoint
+	health, err := newHealthHandler(c)
+	if err != nil {
+		log.Fatalf("Error: %v", err)
+	}
+	http.Handle(endpoint, health)
+	log.Printf("Starting to listen on: %s", c.ListenAddress)
+	log.Fatal(http.ListenAndServe(c.ListenAddress, nil))
+}
+
+func newHealthHandler(c *util.Config) (*healthHandler, error) {
 	color.NoColor = true
 	cache := cache.New(c.Cache, 30*time.Second)
 
 	cfg, err := getGossConfig(c.Vars, c.VarsInline, c.Spec)
 	if err != nil {
-		fmt.Printf("Error: %v\n", err)
-		os.Exit(1)
+		return nil, err
 	}
 
 	output, err := getOutputer(c.NoColor, c.OutputFormat)
 	if err != nil {
-		fmt.Printf("Error: %v\n", err)
-		os.Exit(1)
+		return nil, err
 	}
 
-	health := healthHandler{
+	health := &healthHandler{
 		c:             c,
 		gossConfig:    *cfg,
 		sys:           system.New(c.PackageManager),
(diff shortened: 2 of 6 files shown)


serve.go
```diff
@@ -18,22 +16,30 @@ import (
...
+	health, err := newHealthHandler(c)
+	if err != nil {
+		log.Fatalf("Error: %v", err)
+	}
+	http.Handle(endpoint, health)
+	log.Printf("Starting to listen on: %s", c.ListenAddress)
+	log.Fatal(http.ListenAndServe(c.ListenAddress, nil))
+}
+
+func newHealthHandler(c *util.Config) (*healthHandler, error) {
```
petemounce, line 28: I introduced this func to make the test-setup easier to do and more DRY.

serve.go
```diff
@@ -18,22 +16,30 @@ import (
...
 	if err != nil {
-		fmt.Printf("Error: %v\n", err)
-		os.Exit(1)
+		return nil, err
 	}
 
 	output, err := getOutputer(c.NoColor, c.OutputFormat)
 	if err != nil {
-		fmt.Printf("Error: %v\n", err)
-		os.Exit(1)
```
petemounce, line 33: I replaced these exits with the `log.Fatalf` inside the sub-command, since it makes this easier to write coverage for at some stage. If the system-level exits remain, that would kill the test-runner if it went down this branch.

serve_test.go
```diff
@@ -0,0 +1,57 @@
...
+			if err != nil {
+				require.NoError(t, err)
+			}
+			rr := httptest.NewRecorder()
+
+			handler := http.HandlerFunc(hh.ServeHTTP)
+
+			handler.ServeHTTP(rr, req)
+
+			t.Logf("testName %q log output:\n%s", testName, logOutput.String())
```
petemounce, line 54: If the test case fails, we'll see the log output inside the test's failure output now.

serve_test.go
```diff
@@ -0,0 +1,57 @@
...
+			require.NoError(t, err)
+
+			hh, err := newHealthHandler(config)
+			require.NoError(t, err)
+
+			req, err := http.NewRequest("GET", config.Endpoint, nil)
+			if err != nil {
+				require.NoError(t, err)
+			}
+			rr := httptest.NewRecorder()
```
petemounce, line 48: Inspired (read: stolen) from https://blog.questionable.services/article/testing-http-handlers-go/

outputs/structured.go
```diff
@@ -75,6 +75,10 @@ func (r Structured) Output(w io.Writer, results <-chan []resource.TestResult, st
 
 	fmt.Fprintln(w, string(j))
 
+	if result.Summary.Failed > 0 {
```
petemounce, line 78: Hope you don't mind, I fixed what I think is a bug, here. I think all outputs (except the nagios one?) behave like this change has made this one behave here.

serve.go
```diff
@@ -18,22 +16,30 @@ import (
 
 func Serve(c *util.Config) {
 	endpoint := c.Endpoint
+	health, err := newHealthHandler(c)
+	if err != nil {
+		log.Fatalf("Error: %v", err)
```
petemounce, line 21: This exits 1 after writing the output.

serve_test.go
```diff
@@ -0,0 +1,58 @@
...
+	}
+	for testName := range tests {
+		tc := tests[testName]
+		t.Run(testName, func(t *testing.T) {
+			var logOutput bytes.Buffer
+			log.SetOutput(&logOutput)
+
+			config, err := util.NewConfig(util.WithSpecFile(tc.specFile))
+			require.NoError(t, err)
+			t.Logf("Config: %v", config)
```
petemounce, line 39: It's handy to see (on test-failure only) the process-level goss configuration, even if it takes a little work to eyeball-parse.

serve_test.go
```diff
@@ -0,0 +1,73 @@
...
+			}
+			rr := httptest.NewRecorder()
+
+			handler := http.HandlerFunc(hh.ServeHTTP)
+
+			handler.ServeHTTP(rr, req)
+
+			t.Logf("testName %q log output:\n%s", testName, logOutput.String())
+			assert.Equal(t, tc.expectedHTTPStatus, rr.Code)
+			if tc.expectedContentType != "" {
```
petemounce, line 68: Current behaviour is to _only_ set the header when the output-format == json. I intend to change that in a follow-up PR, so adding some coverage here.

petemounce:
@aelsabbahy if you're around, I think this is now ready for review.

I think this adds about 7% coverage.

serve_test.go
```diff
@@ -0,0 +1,116 @@
...
+	if err != nil {
+		require.NoError(t, err)
+	}
+	rr := httptest.NewRecorder()
+
+	handler := http.HandlerFunc(hh.ServeHTTP)
+
+	t.Run("fresh cache", func(t *testing.T) {
+		handler.ServeHTTP(rr, req)
+		assert.Contains(t, logOutput.String(), "Stale cache")
```
petemounce, line 100: I wanted to assert on the http status code and distinguish a 200 OK from a 304 not modified, but figured that would be a functional change and so left it.

outputs/structured.go
petemounce: (If this was intended behaviour, I can choose a different output to test against easily enough.)

serve.go
aelsabbahy: Since this is being refactored, I would say any os.Exit() or log.Fatalf should be in `cmd/goss/goss.go`

@ripienaar did a refactor a while back and moved all exits up to the top layer where it belonged.

outputs/structured.go
petemounce: Having read the log of that file, I reckon this was by-design now. So, I've removed the change and aimed my test at the rspecish output format instead.

serve.go
petemounce: Done in fb6fa35.

petemounce:
@aelsabbahy friendly ping for further review or merge? I think I've addressed all your feedback so far.

PR #608, opened 2020-08-25
Add blackbox tests for serve command

petemounce:
##### Checklist

### Description of change

Added a cross-platform blackbox test for `goss serve` to give some outside in safety to #606 and incoming output-format-via-http-content-negotiation change.

Also integrated the alpha testing into the `Makefile` for coherence with the others. I'm _not_ a comfortable make-r, so please critique brutally if I've done things poorly?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,11 +41,11 @@ alpha-test-%: release/goss-%
 	$(info INFO: Starting build $@)
 	./integration-tests/run-tests-alpha.sh $*
 
-test-serve-%: release/goss-%
+test-int-serve-%: release/goss-%
 	$(info INFO: Starting build $@)
 	./integration-tests/run-serve-tests.sh $*
 # shim to account for linux being not in alpha
-test-int-serve-linux-amd64: test-serve-alpha-linux-amd64
+test-int-serve-linux-amd64: test-int-serve-alpha-linux-amd64
 
 release/goss-%: $(GO_FILES)
 	./release-build.sh $*
--- a/Makefile
+++ b/Makefile
@@ -36,6 +36,14 @@ bench:
 	$(info INFO: Starting build $@)
 	go test -bench=.
 
+alpha-test-%: release/goss-%
+	$(info INFO: Starting build $@)
+	./integration-tests/run-tests-alpha.sh $*
+
+test-serve-%: release/goss-%
+	$(info INFO: Starting build $@)
+	./integration-tests/run-serve-tests.sh $*
+
 release/goss-%: $(GO_FILES)
 	./release-build.sh $*
 
--- a/Makefile
+++ b/Makefile
@@ -72,9 +72,9 @@ push-images:
 	$(info INFO: Starting build $@)
 	development/push_images.sh
 
-test-int-64: centos7 wheezy precise alpine3 arch
+test-int-64: centos7 wheezy precise alpine3 arch test-int-serve-linux-amd64
 test-int-32: centos7-32 wheezy-32 precise-32 alpine3-32 arch-32
-test-int-all: test-int-32 test-int-64 test-int-serve-$(CURRENT_OS)-amd64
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -36,6 +36,14 @@ bench:
 	$(info INFO: Starting build $@)
 	go test -bench=.
 
+alpha-test-%: release/goss-%
```
aelsabbahy, line 39: When does this get executed?

Makefile
```diff
@@ -36,6 +36,14 @@ bench:
...
+	$(info INFO: Starting build $@)
+	./integration-tests/run-tests-alpha.sh $*
+
+test-serve-%: release/goss-%
```
aelsabbahy, line 43: make test should execute this as part of it's testing, so that it's part of CI.

I was able to test it locally with `make test-serve-linux-amd64` and it looks great.

aelsabbahy (commented):
Looks good, just needs to be hooked into CI

Makefile
petemounce: via `ci/build.sh` in the if darwin or windows bit.

(Honestly, Makefiles are not a strong suit for me, so I haven't wired there)

Makefile
petemounce: It's wired via `ci/build.sh` - or have I made a mistake?

petemounce:
> Looks good, just needs to be hooked into CI

@aelsabbahy I think that it already is via `ci/build.sh`? Is there something further?

aelsabbahy:
I'm fine with build.sh calling one make task, but I don't want build.sh to turn into an orchestration layer of multiple tasks or make calls.

Basically, for local development a user should be able to run `make X` and get the desired set of steps executed. I'm guessing this should be part of make test-int-all.

petemounce:
I've had a go, please take a look?

It's complicated by needing to filter to the current OS that is being executed on, so it sounds like `test-int-64/32/all` need restructuring to make room for the different OS'. `make` is really not my forte (I'm more familiar with https://bazel.build), I would deeply appreciate a detailed steer?

aelsabbahy:
I'll take a look over the weekend. Thanks for making the changes.

If you don't mind can you send me an order list of all the PRs that are ready to go and which order you prefer them merged in (if any).

I haven't used bazel (outside or demoing it), but I love the concepts behind it and nixos for hermetic builds.

integration-tests/run-serve-tests.sh
```diff
@@ -0,0 +1,63 @@
...
+  "serve"
+  "--listen-addr=127.0.0.1:${open_port}"
+)
+echo -e "\nTesting \`${GOSS_BINARY} ${args[*]}\` ...\n"
+
+"${GOSS_BINARY}" "${args[@]}" &
+if curl --silent "http://127.0.0.1:${open_port}/healthz" | grep 'Count: 2, Failed: 0, Skipped: 0' ; then
+  echo "passed"
+else
+  echo "failed, exit code $?"
```
aelsabbahy, line 62: Is this supposed to fail? It just echo's out a message and succeeds.

I added a:
ret=$?
exit $ret

but that didn't work because of the trap

maybe:
`ret=$?`
and a `exit "$ret"` at the bottom of cleanup?

petemounce: Yeah, that was supposed to fail. Fixed inside 23ce6f1. Good catch - sorry!

petemounce:
@aelsabbahy I've rebased and also fixed the lack of failure-on-tests-failing.

aelsabbahy:
Ci test failure appears related to this change.

I think this PR is really close.. but maybe some intermittent issues?

petemounce:
I've rebased and made no changes that I think could have fixed some problem, but it _is_ green now.

aelsabbahy:
Restarted the build 3 times and it was green all three times, so it might be intermittent.

Issue #605, opened 2020-08-19
Any special trick to get CNAME working in dns ? 


alainchiasson:
Figured I would ask, I'm pretty sure its not a bug - I did a tcpdump and saw the CNAME request / response. 

I cannot get the CNAME test to work as I think it should - testing that a CNAME target is correct. It always seems to compare against the end target (the IP). I saw it being called "flattened cname". 

Is there something I am missing ?

Example -----

```
[root@35db6826ec31 ~]# dig mail.chiasson.org

; <<>> DiG 9.11.13-RedHat-9.11.13-5.el8_2 <<>> mail.chiasson.org
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 48980
;; flags: qr rd ra; QUERY: 1, ANSWER: 2, AUTHORITY: 0, ADDITIONAL: 0

;; QUESTION SECTION:
;mail.chiasson.org.		IN	A

;; ANSWER SECTION:
mail.chiasson.org.	4502	IN	CNAME	chiasson.org.
chiasson.org.		4502	IN	A	143.95.235.64

;; Query time: 3 msec
;; SERVER: 192.168.65.1#53(192.168.65.1)
;; WHEN: Wed Aug 19 11:32:29 UTC 2020
;; MSG SIZE  rcvd: 65

[root@35db6826ec31 ~]# goss add dns CNAME:mail.chiasson.org
Adding DNS to './goss.yaml':

CNAME:mail.chiasson.org:
  resolvable: true
  addrs:
  - 143.95.235.64
  timeout: 500


[root@35db6826ec31 ~]# sed -ic '/143.95.235.64/s/143.95.235.64/chiasson.org/' goss.yaml
[root@35db6826ec31 ~]# cat goss.yaml
dns:
  CNAME:mail.chiasson.org:
    resolvable: true
    addrs:
    - chiasson.org
    timeout: 500
[root@35db6826ec31 ~]# goss v -f documentation
DNS: CNAME:mail.chiasson.org: resolvable: matches expectation: [true]
DNS: CNAME:mail.chiasson.org: addrs:
Expected
    <[]string | len:1, cap:1>: ["143.95.235.64"]
to contain element matching
    <string>: chiasson.org


Failures/Skipped:

DNS: CNAME:mail.chiasson.org: addrs:
Expected
    <[]string | len:1, cap:1>: ["143.95.235.64"]
to contain element matching
[...]

PR #610, opened 2020-08-26
Clarify the need for server for extra dns tests

alainchiasson:
Making it more obvious that the server entry is required. 

##### Checklist

### Description of change

Adding emphasis to server attribute requirements for the additional queries. I overlooked it in #605


aelsabbahy:
LGTM, thanks!

PR #631, opened 2020-10-09
Alpha test more quietly

### LINKED ISSUE #629: Removing alpha warning when flag is set
**Describe the feature:**

Goss alpha is too noisy.

Goss should not always display alpha warning, only when when use-alpha is not set.

petemounce:
Fixes #629.

##### Checklist


aelsabbahy:
What's this failure all about? ss command not found on windows?

petemounce:
Yeah, `ss` doesn't appear to exist on Windows, so the command falls back to a random port selection. :(

I didn't find a similarly-fast way on Windows to find a randomly selected _available_ port.

* `netstat -an` too slow, tens of seconds
* sysinternals `tcpvcon -ac` fast, doesn't include port like the GUI variant `tcpview` does 🤦 

So, I wrote a piece of powershell and included it into this PR - hope the change is ok within this PR, but if not, happy to slice it to a different one? 894b0a9

petemounce:
I enjoyed cb7c137.

PR #641, opened 2020-10-23
Allow log level to be set and filtering logs accordingly

obourdon:
##### Checklist

### Description of change

Some of the goss default logging messages might be seen as not very useful. This fix allow easy log level customisation and filtering


aelsabbahy:
I haven't had a chance to go through this PR in depth, but this was a much needed feature. Many thanks for taking the time to add this!

The main thing I'll be reviewing is making sure this PR and #576 don't have major conflicts.

obourdon:
@aelsabbahy no pb, many thanks to you for the great work

aelsabbahy:
One of these days I need to fork the go-lang code and run code climate on it. Very curious what the results would be.

aelsabbahy:
Which function was failing?

obourdon:
> Which function was failing?

strangely the error was triggered by my changes in Validate but seems like other functions also had this issue without failing [see this link](https://codeclimate.com/github/aelsabbahy/goss/pull/641)

aelsabbahy:
Took this for a spin, it's a very nice change that greatly improves debugging and can be built upon. The only thing that needs to change is that INFO is too noisy/chatty as a default.

aelsabbahy:
Oh, and I think a small writeup on the log-levels and when they should be used would be nice, but that can be a separate PR/improvement.

obourdon:
> Took this for a spin, it's a very nice change that greatly improves debugging and can be built upon. The only thing that needs to change is that INFO is too noisy/chatty as a default.

Understood and agreed will do the change soon (after rebased on v0.3.14)

obourdon:
> Oh, and I think a small writeup on the log-levels and when they should be used would be nice, but that can be a separate PR/improvement.

Will do too

obourdon:
Done with additional documentation. @aelsabbahy however, after doing so, I do not understand why you find INFO too verbose, may be just changing the default logging level to WARN instead of INFO would better match your thoughts

obourdon:
@aelsabbahy done rebasing on latest master
Do you think you'll merge this in some future version please ?

aelsabbahy:
```
$ goss v
2020/11/25 11:04:33 Setting log level to WARN
........

Total Duration: 0.002s
Count: 8, Failed: 0, Skipped: 0
```

obourdon:
@aelsabbahy got it, will have a look at this later this week. Many thanks for making this way clearer

aelsabbahy:
https://dave.cheney.net/2015/11/05/lets-talk-about-logging

This might not be a bad approach. Split logging into just two levels that are both disabled by default.. INFO and DEBUG.

Can use INFO to help users get more info regarding goss and DEBUG for more developer level debugging. But I'm open to suggestions.

obourdon:
@aelsabbahy finally got some time to rework on this. Could you please check that the new code fullfill your requirements
I also tried to understand why the CI is failing on OSX but I do not see why `brew install bash` is failing. Many thanks in advance

obourdon:
@aelsabbahy finally found a way to fix OSX CI by building bash homebrew formula from source. Currently trying another option

obourdon:
@aelsabbahy both solutions work, only keeping the binary HomeBrew formula version

obourdon:
@aelsabbahy any chance you could review and possibly integrate this soon please ?

sebinjohn:
Is there a plan to revive this PR. I'm interested in this feature.

obourdon:
@sebinjohn unfortunately looking at #699 and #713 the only possibility for you is to fork this PR to use it (BTW this is already what we are doing on our side for quite some time)
We were considering maintaining our own supported fork but this is completely non possible for our team at this point in time

ripienaar:
It's unfortunate, because now there is also a sever security vulnerablity in runc that this package depends on, it's likely we will have no choice but to maintain a fork :( @obourdon @sebinjohn

sebinjohn:
@obourdon  thanks for the info.  I will look into forking further.

obourdon:
@sebinjohn see PR [https://github.com/goss-org/goss/pull/802](https://github.com/goss-org/goss/pull/802)

Issue #616, opened 2020-09-15
Goss should be able to assert against Windows Registry state


DoomGerbil:
**Describe the feature:**

Goss should be able to assert against the state and contents of the Windows Registry.

**Describe the solution you'd like**

In order to properly support Windows configuration use cases, _lots_ of things will need to be able to assert against registry keys. On *nix systems, you'd generally use a `file.contains` assertion against a configuration file, but the Windows equivalent is generally keys stored in the registry.

**Describe alternatives you've considered**

The best I've got right now is that you can write a Powershell script that checks values in the registry and then outputs status, and then you wrap that script in a goss assertion. This is hacky, requires emplacing the script in the right place with the right permissions, etc, isolates _some_ of the configuration to be checked separate from the rest, and just generally feels dirty.

aelsabbahy:
@petemounce Any thoughts on this?

petemounce:
I'm definitely in favour of being able to assert against it.

The parallel approach works but due to its startup time injects about a second (unscientific approximate measure!) of overhead per assertion.

aelsabbahy:
Marked as approved. Windows and Mac are community supported, so I'll review/merge changes but I won't be the one making the change on this one.

PR #633, opened 2020-10-12
Add support for native windows commands

jsturtevant:
##### Checklist

- [x] `make test-all` (UNIX) passes. CI will also test this
- [ ] unit and/or integration tests are included (if applicable) 
- [ ] documentation is changed or added (if applicable)

### Description of change

Fixes #632 

Uses Native tools to execute commands on the host machine which enables better support for windows.   This requires the paths for the command files to be in correct format with `\` instead of `/`.

PR #651, opened 2020-11-18
In 0.3.14, command assertions that wrap powershell can no longer assert against stdout/stderr successfully

petemounce:
## Edit: Problem In Chair, Not In Computer on my part

see comment thread. Thanks @jsturtevant.

This is now a PR to add an integration test that illustrates how to assert via `command` using `powershell`

## Original content of PR description:

Assertion that fails in `0.3.14` that passed in `0.3.13`:

```yaml
command:
  Anonymous logins are disabled:
    exit-status: 0
    exec: powershell -noprofile -noninteractive -command "(get-itemproperty -path 'HKLM:/SYSTEM/CurrentControlSet/Control/Lsa/').restrictanonymous"
    stdout:
      - '1'
    timeout: 10000 # in milliseconds
    skip: false
```

## 0.3.13 passes

Screenshot 2020-11-18 at 11 16 59

## 0.3.14 fails

(Note goss.exe being referenced via `./` rather than taking the 0.3.13 as on PATH)

Screenshot 2020-11-18 at 11 15 12

##### Checklist

### Description of change

In 0.3.14, my command assertions that used powershell.exe to run powershell started to fail, and I suspect the new cmd /c wrapping, so I'm adding an integration test first. This also happens to be how I'm using goss to assert against the windows registry (note #616).

If this integration test fails I'll try to make changes to allow it to pass - I'm expecting to want to make it possible to use powershell as the shell within command assertions to achieve that.

cc @jsturtevant in case you'd like to take a stab at this?

### DIFF
--- a/integration-tests/goss/windows/tests/command.goss.yaml
+++ b/integration-tests/goss/windows/tests/command.goss.yaml
@@ -8,14 +8,14 @@ command:
     stderr: []
     timeout: 10000
   wrap a powershell - expect 1:
-    exec: powershell -noprofile -noninteractive -command "(get-itemproperty -path 'HKLM:/SYSTEM/CurrentControlSet/Control/Lsa/').restrictanonymous"
+    exec: powershell -noprofile -noninteractive -command (get-itemproperty -path 'HKLM:/SYSTEM/CurrentControlSet/Control/Lsa/').restrictanonymous
     exit-status: 0
     stdout:
     - 1
     stderr: []
     timeout: 10000
   wrap a powershell - expect 0:
-    exec: powershell -noprofile -noninteractive -command "(get-itemproperty -path 'HKLM:/SYSTEM/CurrentControlSet/Control/Lsa/').restrictanonymous"
+    exec: powershell -noprofile -noninteractive -command (get-itemproperty -path 'HKLM:/SYSTEM/CurrentControlSet/Control/Lsa/').restrictanonymous
     exit-status: 0
     stdout:
     - 0
--- a/integration-tests/goss/windows/tests/command.goss.yaml
+++ b/integration-tests/goss/windows/tests/command.goss.yaml
@@ -7,14 +7,7 @@ command:
     - hello world
     stderr: []
     timeout: 10000
-  wrap a powershell - expect 1:
-    exec: powershell -noprofile -noninteractive -command (get-itemproperty -path 'HKLM:/SYSTEM/CurrentControlSet/Control/Lsa/').restrictanonymous
-    exit-status: 0
-    stdout:
-    - 1
-    stderr: []
-    timeout: 10000
-  wrap a powershell - expect 0:
+  wrap a powershell - expect 0 because travis does not restrict anonymous logins:
     exec: powershell -noprofile -noninteractive -command (get-itemproperty -path 'HKLM:/SYSTEM/CurrentControlSet/Control/Lsa/').restrictanonymous
     exit-status: 0
     stdout:
--- a/integration-tests/goss/windows/tests/command.goss.yaml
+++ b/integration-tests/goss/windows/tests/command.goss.yaml
@@ -7,3 +7,17 @@ command:
     - hello world
     stderr: []
     timeout: 10000
+  wrap a powershell - expect 1:
+    exec: powershell -noprofile -noninteractive -command "(get-itemproperty -path 'HKLM:/SYSTEM/CurrentControlSet/Control/Lsa/').restrictanonymous"
+    exit-status: 0
+    stdout:
+    - 1
+    stderr: []
+    timeout: 10000
+  wrap a powershell - expect 0:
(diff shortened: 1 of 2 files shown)


integration-tests/goss/windows/tests/command.goss.yaml
```diff
@@ -7,3 +7,17 @@ command:
...
+    - 1
+    stderr: []
+    timeout: 10000
+  wrap a powershell - expect 0:
+    exec: powershell -noprofile -noninteractive -command "(get-itemproperty -path 'HKLM:/SYSTEM/CurrentControlSet/Control/Lsa/').restrictanonymous"
+    exit-status: 0
+    stdout:
+    - 0
+    stderr: []
+    timeout: 10000
```
petemounce, line 23: There are two tests because I don't know how Travis secures their build machines; _one_ of these should pass, though.

jsturtevant:
I gave some thought to making the command either be powershell or cmd but did some basic testing and found I could invoke powershell from the cmd wrapping.  Maybe we can revisit this if necessary.

I think in this case, it is the way the quotes are processed.  I removed the quotes:

```yaml
command:
  Anonymous logins are disabled:
    exit-status: 0
    exec: powershell -noprofile -noninteractive -command (get-itemproperty -path 'HKLM:/SYSTEM/CurrentControlSet/Control/Lsa/').restrictanonymous
    stdout:
      - '1'
    timeout: 10000 # in milliseconds
    skip: false
```

and it passed:

```
```

I had to remove some quotes from the commands in #633 as well.  Maybe there is something we can do to process the quotes properly?

petemounce:
@jsturtevant thank you! That allows me to work around this very handily.

integration-tests/goss/windows/tests/command.goss.yaml
```diff
@@ -7,3 +7,17 @@ command:
     - hello world
     stderr: []
     timeout: 10000
+  wrap a powershell - expect 1:
+    exec: powershell -noprofile -noninteractive -command (get-itemproperty -path 'HKLM:/SYSTEM/CurrentControlSet/Control/Lsa/').restrictanonymous
```
jsturtevant, line 11: looks like this command fails:

```
wrap a powershell - expect 1: stdout: patterns not found: [1]
```

Are both of these examples needed?

petemounce: No, definitely not; I added both because I have no idea what the expected value should be within Travis. Adding both was reconnaissance.

integration-tests/goss/windows/tests/command.goss.yaml
jsturtevant: makes sense :-)

integration-tests/goss/windows/tests/command.goss.yaml
jsturtevant: makes sense :-)

jsturtevant:
Always love more tests!  LGTM

aelsabbahy:
Appreciate the discussion here and the added test coverage. Many thanks to both of you for the continued support of windows on goss.

Given that this was non-intuitive initially for @petemounce is there an opportunity to improve the documentation here for goss windows users?

aelsabbahy:
Another question for you both. How stable is windows binary at this point?

Should we consider renaming the "alpha" to "community" or is it (and Mac) not ready for that yet?

petemounce:
In terms of stability, we haven't noticed anything obviously non-functional or non-performant. We've had 0.3.14 in production since a week after it was released, on Windows and macOS in addition to Linux.

There is a change I may want to do around how Windows process priorities are handled - we discovered that goss doesn't inherit onto sub-processes the `goss serve` process priority. This leads to resource starvation inside our build farm, since builds are trying to get as much of the computer's resources as possible; we've set up goss serve to have a high priority, but when it runs (for example) `command` assertions, that process priority is not inherited, and so we get timing out checks sometimes.

That's something we'll probably look at soonish, but not immediately. I don't think that needs to block renaming "alpha" to "community"; the integration-test coverage for both platforms is reasonably encompassing.

petemounce:
re: docs - I'm not sure. I'm actually still not sure _why_ this was a fix or a quoting issue, really.

aelsabbahy:
Cool, mostly asking because I'm trying to get v4 out by end of dec or early Jan. Hoping to rename mac and windows as part of that release.

jsturtevant:
> Cool, mostly asking because I'm trying to get v4 out by end of dec or early Jan. Hoping to rename mac and windows as part of that release.

@aelsabbahy just wanted to let you know we are now using this in image-builder as part for cluster api: https://github.com/kubernetes-sigs/image-builder/pull/563

Thanks for all your hard work!

petemounce:
@jsturtevant that's great! Thanks for checking back and saying so, that's given me a warm fuzzy feeling :)

PR #662, opened 2020-12-09
Expose service ports when running dcgoos

frederikbosch:
I'd like [gliderlabs/registrator ](https://github.com/gliderlabs/registrator) to pick up services from my `docker-compose.yml` file and register the service in consul. Then it is required to have ports exposed.

### DIFF
--- a/extras/dcgoss/dcgoss
+++ b/extras/dcgoss/dcgoss
@@ -38,11 +38,11 @@ run(){
     case "$GOSS_FILES_STRATEGY" in
       mount)
         info "Starting docker container"
-        docker-compose run -d -T --name "$service" -v "$tmp_dir:/goss" --rm "$service"
+        docker-compose run --service-ports -d -T --name "$service" -v "$tmp_dir:/goss" --rm "$service"
         ;;
       cp)
         info "Creating docker container"
-        docker-compose run -d -T --name "$service" --rm "$service" > /dev/null
+        docker-compose run --service-ports -d -T --name "$service" --rm "$service" > /dev/null
         info "Copy goss files into container"
         docker cp "$tmp_dir/." "$service:/goss"
         ;;
--- a/extras/dcgoss/dcgoss
+++ b/extras/dcgoss/dcgoss
@@ -38,11 +38,11 @@ run(){
     case "$GOSS_FILES_STRATEGY" in
       mount)
         info "Starting docker container"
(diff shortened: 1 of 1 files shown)


aelsabbahy:
So dcgoss is community maintained/supported. That said, why not follow the pattern that dgoss has and just allow users to pass docker-compose flags on the CLI when they run dcgoss, this way dcgoss is flexible and out of the way?

example:
https://github.com/aelsabbahy/goss/blob/620d32b003efedde8ee98e917005dd4642d88dc1/extras/dgoss/dgoss#L42

@fbartels, @officerJones, @frederikbosch would love your feedback.

frederikbosch:
Perfect, will update

fbartels:
My initial thought was "why export a container to the outside world, that only shortly exists during test runtime", but I agree with @aelsabbahy having a generic option to pass additional parameters seems to be the better choice.

frederikbosch:
> "why export a container to the outside world, that only shortly exists during test runtime"

My use case is not 'only shortly exists'. With docker-compose I boot a complete cluster with `docker-compose run a b c`, excluding the container that is under test (container d), then I run some things on those booted containers, and then I run `dcgoss run d`. Containers a b c will then communicate with container d. After `GOSS_SLEEP` I test the contents of the container under test and see if the communication between a b c resulted in the right placement of files container d.

frederikbosch:
PR updated

fbartels:
@frederikbosch an interesting case

frederikbosch:
Long story short, I use dcgoss for integration/system tests. It tests the end goal, via a fully booted cluster. A single test usually takes 1-5 minutes.

aelsabbahy:
Looks good to me, is this PR good to go? Any documentation changes needed?

aelsabbahy:
Cool use-case, I assume it's not public/open source.. but if it is, I'd love to look at the code/pattern.

I have plans to re-write the goss integration tests using docker compose to get rid of all external dependencies.

frederikbosch:
Every test I create is a single dir with a simple `test.sh` that executes `docker-compose run`, `curl`, `sleep` and `dcgoss`. Further the dir contains files that are used for booting/configuring the cluster. With `curl` I communicate with the booted cluster and post some resources on the APIs that are exposed by the cluster. The best part is that it works perfectly in my CI environment. The worst part is that it requires much more patience than a lib using simple unit testing.

aelsabbahy:
I assume this is good to go, if I misunderstood please let me know and I'll roll it back.

Awesome work, much appreciated!

PR #663, opened 2020-12-10
Rename "alpha" -> "community"

aelsabbahy:
##### Checklist

### Description of change

Rename alpha to community as discussed in #651 

@petemounce would love a quick review on this. It's just a naive `s/alpha/community/g` + file rename approach that I took and the build is green. Please let me know if there's anything I missed or if I renamed anything I wasn't supposed to =).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,15 +36,15 @@ bench:
 	$(info INFO: Starting build $@)
 	go test -bench=.
 
-alpha-test-%: release/goss-%
+community-test-%: release/goss-%
 	$(info INFO: Starting build $@)
-	./integration-tests/run-tests-alpha.sh $*
+	./integration-tests/run-tests-community.sh $*
 
 test-int-serve-%: release/goss-%
 	$(info INFO: Starting build $@)
 	./integration-tests/run-serve-tests.sh $*
-# shim to account for linux being not in alpha
-test-int-serve-linux-amd64: test-int-serve-alpha-linux-amd64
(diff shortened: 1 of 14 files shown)


petemounce:
Is there mention of GOSS_USE_ALPHA changing?

As far as I remember there is no automated coverage for the messaging part. I'll check that over here over the next few days.

What are your thoughts on:

* naming the binaries _without_ the suffix
* ... but keeping the messaging
* ... and keeping the must-use-flag-or-environment-variable awareness-need

?

Speaking for myself, dealing with a binary that is differently named across platforms requires an uncomfortable edge case in a few places that I would prefer to avoid.

Regardless of the precise nature of the change here, I'd like it if this were delayed until 0.4 because it is a breaking change.

aelsabbahy:
I'm fine with the changes you mentioned. Can you submit a PR with how you'd like this to work.

I won't merge the PR until v4 is merged (hopefully late dec, early Jan)

aelsabbahy:
Closing this in favor of @petemounce solution

aelsabbahy:
Forgot to mention, when creating a PR for this, please create against the v4 branch

aelsabbahy:
Do you happen to have an ARM macbook by any chance? Guessing the goss binary would need both arm and amd64

petemounce:
I'm gonna take a look at this over the next few days. Sadly, no, I don't have an ARM macbook. I'll branch from v4.

PR #665, opened 2020-12-15
Fix for invalid PR #662

frederikbosch:
The service name should be left out from the command as it is included in `"${@:2}"`. Sorry for the invalid PR before.


aelsabbahy:
No problem, I don't actually test the extras beyond dgoss. Thanks for submitting a fix so quickly!

PR #666, opened 2020-12-16
Fix windows cmd quoting

jsturtevant:
##### Checklist

- [X] `make test-all` (UNIX) passes. CI will also test this
- [X] unit and/or integration tests are included (if applicable)

### Description of change

In #651 we found that quoting was not working properly.  The following would fail:

```
 wrap a powershell with quotes - expect 0 because travis does not restrict anonymous logins:
    exec: powershell -noprofile -noninteractive -command "(get-itemproperty -path 'HKLM:/SYSTEM/CurrentControlSet/Control/Lsa/').restrictanonymous"
```

Not using quotes worked for that use case but for more complex scenarios quoting is necessary. Digging into it a bit more I found https://golang.org/pkg/os/exec/#Command:

> On Windows, processes receive the whole command line as a single string and do their own parsing. Command combines and quotes Args into a command line string with an algorithm compatible with applications using CommandLineToArgvW (which is the most common way). Notable exceptions are msiexec.exe and cmd.exe (and thus, all batch files), which have a different unquoting algorithm. In these or other similar cases, you can do the quoting yourself and provide the full command line in SysProcAttr.CmdLine, leaving Args empty.   

This PR uses the second approach.

cc: @petemounce

### DIFF
--- a/integration-tests/goss/windows/tests/command.goss.yaml
+++ b/integration-tests/goss/windows/tests/command.goss.yaml
@@ -14,3 +14,17 @@ command:
     - 0
     stderr: []
     timeout: 10000
+  wrap a powershell with quotes - expect 0 because travis does not restrict anonymous logins:
+    exec: powershell -noprofile -noninteractive -command "(get-itemproperty -path 'HKLM:/SYSTEM/CurrentControlSet/Control/Lsa/').restrictanonymous"
+    exit-status: 0
+    stdout:
+    - 0
+    stderr: []
+    timeout: 10000
+  powershell with quotes:
+    exec: powershell /c "(echo '{"b":2, "a":1}' | ConvertFrom-json).a"
+    exit-status: 0
+    stdout:
+    - "1"
+    stderr: []
+    timeout: 10000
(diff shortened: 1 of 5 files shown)


jsturtevant:
the codeclimate issue is: `exported function NewCommandForWindowsCmd should have comment or be unexported`

The `NewCommand` was missing a comment and I copied that style. Should I update the new function with a comment?

petemounce (approved):
Great find, and I really appreciate the revisit!

aelsabbahy:
@petemounce is this good to go? I might cut a release tomorrow.. so if this is ready I can merge it prior.

Awesome work tracking this down @jsturtevant !

petemounce:
@aelsabbahy yes, I think this looks good.

jsturtevant:
getting into the next release would be awesome.  Do I need to address the codeclimate error?

aelsabbahy:
If you want to address it, you can. Honestly, the code is littered with that error and I need to clean it up across the board one day.. I won't hold you to a standard I haven't implemented myself =)

aelsabbahy:
Merging, so I can cut a release with this and other changes

PR #675, opened 2021-01-18
add MacOS and Windows support to install.sh

undefynd:
I wanted to be able to install goss with the install.sh script, but it always installed the Linux binary. So I have updated the script to install goss also and MacOS and Windows.

##### Checklist

- [ ] `make test-all` (UNIX) passes. CI will also test this
- [ ] unit and/or integration tests are included (if applicable)
- [ ] documentation is changed or added (if applicable)

Issue #329, opened 2018-03-12
Parametrization of TMP_DIR needed?


comradefuzz:
I had problems with my CI integration. When dgoss runs my container it cannot access to mounted directory files.

I have this CI configuration:
```
image: docker:latest

services:
  - docker:dind
test:
  stage: test
  script:
    - apk add bash curl --no-cache
    - curl -fsSL https://goss.rocks/install | sh
    - docker pull "$CI_REGISTRY_IMAGE"
    - dgoss run "$CI_REGISTRY_IMAGE"
```

dgoss fails with output:
```
INFO: Starting docker container
INFO: Container ID: bbba20c7
INFO: Sleeping for 0.2
INFO: Running Tests
sh: /goss/goss: not found
ERROR: Job failed: exit code 127
```
Some painful debugging shown that directory mounted as `/goss` to container with `goss` binary file is empty while source temporary directory on host `/tmp/tmp.XXXXXXXXXX` contains correct files.
I put some patches in `dgoss` script that make temporary directory appear in CWD and problem seems to be fixed. I suppose that mounted directory in gitlab pipelines environment in `/tmp` is inaccessible in container context.

I suggest you can simply parametrize location of $tmp_dir to fix this issue.

iWader:
Also having this problem on CircleCI

moschlar:
The problems with `GOSS_FILES_STRATEGY=cp` are actually due to the docker-in-docker mechanism.
https://docs.gitlab.com/ee/ci/docker/using_docker_build.html#use-docker-in-docker-executor has a bullet point on these issues.
But I think for dgoss, just using the mentioned workaround is sufficient.

aelsabbahy:
@comradefuzz feel free to create a pull request that allows overriding the tmp_dir location.

aelsabbahy:
I believe this is fixed with #684, feel free to let me know if it doesn't.

Thank you for submitting this feature request.

PR #684, opened 2021-03-08
added DGOSS_TEMP_DIR environment variable.

ksaito1125:
##### Checklist

### Description of change

Location of the temporary directory used by dgoss.


ksaito1125:
#675

aelsabbahy:
Hello, thank you for taking the time to contribute this. I've removed the stale and I'm going through the old PRs that were impacted by it.

Can you provide a little bit of context on the use-case of this. I understand what it does on a technical level, just trying to better understand when it's needed by the user?

aelsabbahy:
I'm merging this as it seems like a good feature to have, and seems to be a solution for #329 

Thank you for the great contribution, apologies for the extremely long delay in merging this.

PR #730, opened 2022-01-27
bumped Go from 1.13.x to 1.17.x in the hopes of fixing CI

ekelali:
I are hoping that this change - namely, to bump Go from version 1.13.x to 1.17.x - will fix CI without having to exclude OSX

##### Checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ env:
 language: go
 
 go:
-  - 1.13.x
+  - 1.17.x
 
 os:
   - osx


mbainter:
I believe the only thing you're running into here is changes to the conditional build directives in 1.17.  If you go through resource/resource_list_genny.go, the system/* files and util/command_windows.go and update those directives that should work.

Just replace:
`// +build linux darwin !windows`

with:
`//go:build linux || darwin || !windows`

and you should be good to go.  With those changes and a `go mod tidy` the build seems to work well for me locally under 1.17.6.

PR #736, opened 2022-02-01
Remove curl dependency from install.sh

jkroepke:
##### Checklist

### Description of change

Remove curl dependency from install.sh by using github itself to gain the lastest version.

Additionally:
- Lint shell script with shellcheck
- Format shell script with shfmt


berney (approved):
LGTM

Issue #722, opened 2021-11-06
Support for  linux/arm64 architecture


tnaroska:
**Describe the feature:**
Goss currently provides 32bit arm binaries in the release but not arm64.

**Describe the solution you'd like**
- add linux arm64 to the supported architectures
- provide binary builds for arm64 in the GitHub release page

**Describe alternatives you've considered**
- sync codebase and build myself. But it’s more desirable to get all binaries out of the official build pipeline.

linki:
@tnaroska I created a PR for an arm64 release [here](https://github.com/aelsabbahy/goss/pull/737). I would like to hear your thoughts on this.

tnaroska:
Thanks for working on this @linki ! PR looks good to me.
My use case is AWS Graviton/Apple M1 only, so I cannot comment on the 32-bit arm part in install.sh.

smwltr:
@linki Is there anything I can do to help? Do you need some testing done?

linki:
@smwltr I just simplified https://github.com/aelsabbahy/goss/pull/737 and we can hopefully merge this soon. If you can test the resulting `goss-linux-arm` (should be 32-bit) and `goss-linux-arm64` (should be 64-bit) after we merged the PR, then that would be great.

If you have access to a 32-bit arm machine you could check its out for `uname -m` and we could add it to the `if` statement in https://github.com/aelsabbahy/goss/pull/756. Or if you're certain about the other possible values we could also just add them without confirming first.

smwltr:
@linki I only have access to ARM64 machines, which gives: `aarch64`

PR #737, opened 2022-02-02
Build binary for arm64

linki:
This is an attempt to build goss for `arm64` so we can run it on Graviton2 machines on AWS.

It fixes https://github.com/aelsabbahy/goss/issues/722.

It adds a Makefile target for `release/goss-linux-arm64` alongside `release/goss-linux-arm`, `release/goss-linux-amd64` etc.

Note, that this results in the [released assets](https://github.com/aelsabbahy/goss/releases/tag/v0.3.16) being changed slightly going forward: `goss-linux-arm` will be a 32-bit ARM binary and `goss-linux-arm64` will be a 64-bit ARM binary.

### Other changes

I also changed `install.sh` a little because the current implementation actually builds for the wrong architecture: when `uname -m` returns `aarch64` it should use `GOARCH=arm64` instead of the currently used `GOARCH=arm` which results in a 32-bit binary.

I also added `arm64` to the 64-bit if-statement because that's what is returned on the arm-based Macbooks.

In order to also support 32-bit ARM in `install.sh` I matched on `aarch32` but I don't know if that's correct. I believe that wouldn't work on, e.g. Raspberry Pis, because I think they return something like `armv7` or so. I didn't prioritize it here because the current implementation didn't do the right thing anyways.

/cc @tnaroska because you created the original issue
/cc @garyschulte because you [have built for arm64 as well](https://github.com/garyschulte/goss/commit/0700d5e2082390926941cf4f259ccd7020148c98)


ekelali:
Thank you for submitting this PR:

* Can you split the install.sh change to a separate PR. This way I can publish the new binaries prior to install.sh update.

* What are the implications of `aarch64` moving from `arm` -> `arm64` in the install script? Is it broken today, does this introduce a breaking change?

linki:
I'm still working on this feature on my end. I'll follow up here as soon as possible.

linki:
@ekelali Thanks for the suggestion to split the PR. I wasn't sure about the purpose of `install.sh` but it looks like it's not relevant to the release process.

It should be much easier to get this merged now since it will only add one more binary and shasum to each release, not touching anything else.

> Note, that this results in the [released assets](https://github.com/aelsabbahy/goss/releases/tag/v0.3.16) being changed slightly going forward: goss-linux-arm will be a 32-bit ARM binary and goss-linux-arm64 will be a 64-bit ARM binary.

This was actually a wrong comment by me. `goss-linux-arm` is already a 32-bit ARM binary (like it should) and this PR adds a `goss-linux-arm64` 64-bit ARM binary alongside it. So the releases are **backwards compatible**.

I'll clarify the changes to `install.sh` in a separate PR.

linki:
@aelsabbahy Would you mind creating a release that includes the `arm64` binaries?

linki:
@aelsabbahy Thanks! 🤗

aelsabbahy:
Thank you for the contribution, greatly appreciated!

PR #739, opened 2022-02-06
chore: first pass at dependency version updates

mbainter:
##### Checklist

### Description of change

Updates all of the modules that are easy low-hanging fruit with no breaking changes.  It's broken up by module right now so any associated dependency changes can be seen.  This can be squashed into a cleaner history if everything looks good.


ekelali:
@mbainter just wanted to make sure this was ready to be merged?

mbainter:
This should be good to go.  After I opened the PR and started going back through the remaining ones that looked sketchier I found a few more that were sufficiently straightforward they could be lumped into this, but I'm done with that now.  The rest should probably be reviewed and added independently.

PR #748, opened 2022-03-23
dgoss: add support for podman as the container backend

### LINKED ISSUE #507: Feature: Support for podman / buildah
Hi,

I am using podman for some time now and am starting to use goss as validator / verifier for ansible and my containers. For docker, you have the dgoss, dcgoss wrapper, which is already working fine.

I request to either:

- extend the existing wrapper to support podman via cli switch
- add pgos and bgoss for podman and buildah
- provide cgoss, which is a general container+goss wrapper with support for docker, podman, etc.

After a confirmation of one of the above, I can help to develop this wrapper in the way, you prefer.

thrix:
##### Checklist

### Description of change
If `podman` is available on the host, let's use it instead of `docker`. Podman
does not need a daemon to run and is common in RHEL-like OSes.

To be able to use `podman` as the container backend `podman run` needs to run
a command which does not finish. Commonly `sleep infinity` is used for that.

Provide enough documentation to understand the feature for the users.

I believe keeping just one command `dgoss` is a better approach, let's not complicate things for users.

Partially resolves #507

### DIFF
--- a/extras/dgoss/README.md
+++ b/extras/dgoss/README.md
@@ -101,3 +101,6 @@ Strategy used for copying goss files into the docker container. If set to `'moun
 
 ##### CONTAINER_LOG_OUTPUT
 Location of the file that contains tested container logs. Logs are retained only if the variable is set to a non-empty string. (Default `''`)
+
+##### CONTAINER_RUNTIME
+Container runtime to use - `docker` or `podman`. Defaults to `docker`. Note that `podman` requires a run command to keep the container running. This defaults to `sleep infinity` in case only an image is passed to `dgoss` commands.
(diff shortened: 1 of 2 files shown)


ekelali:
I would suggest the default continues to be docker. Those that want to use podman can set `CONTAINER_RUNTIME` to override.

Changing the default would be a breaking change for dgoss consumers that happen to have podman installed.

thrix:
@ekelali ack, let's mitigate breaking changes

thrix:
@ekelali it should be resolved

thrix:
@aelsabbahy hi, do you think we could get this in, or should I do some more changes?

aelsabbahy:
Sorry, the travis-ci OSS credits ran out again, I had to open up a support ticket. To get the job to trigger, I had to close/reopen the issue.

PR #757, opened 2022-05-04
update some dependencies

ripienaar:
##### Checklist

### Description of change

Fixes a security issue in runc and moves to
some later versions of docker

Signed-off-by: R.I.Pienaar <rip@devco.net>


aelsabbahy:
Tagging @petemounce  since he was one of the main drivers behind windows support.

petemounce:
I definitely believe goss support for Windows is valuable.

I haven't understood the reason for that coming up, though - CI shows macOS passing, and Linux and Windows failing for the same error
* Linux: `../../../../pkg/mod/github.com/aelsabbahy/!g!onetstat@v0.0.0-20160428114218-edf89f784e08/gonetstat.go:174:10: undefined: user.LookupUid`
* Windows: `..\..\..\..\pkg\mod\github.com\aelsabbahy\!g!onetstat@v0.0.0-20160428114218-edf89f784e08\gonetstat.go:174:10: undefined: user.LookupUid`

https://pkg.go.dev/os/user#LookupId suggests nothing about any Windows-ism...?

ripienaar:
The `user` in question is not the stdlib `user` it's https://github.com/aelsabbahy/GOnetstat/blob/edf89f784e0876818dc19f7744a16742a0a66f16/gonetstat.go#L22 which is why I pointed to the change in that repo.

petemounce:
Ah, I see; thanks. I think I eyeball parsed that away when I googled for the docs.

https://github.com/aelsabbahy/goss/blob/master/docs/platform-feature-parity.md#matrix---testsassertions - `mount` on there is currently not implemented for Windows. And, won't be - Windows doesn't really have the same concept (instead it has volumes, mapped differently than mount-points).

So - `mount` on Windows can be ignored; removing support will make it no more not-implemented than now :)

ripienaar:
Same code is in used in main goss code lines for group/users on windows. https://github.com/aelsabbahy/goss/blob/a2153db055d1127fcc2045e1c48cf276fb387697/system/user.go

So options I see are 1) to find an alternative a new multi OS user/group membership lib 2) vendor the old runc code wrt users/groups 3) drop support for users/groups on windows

ripienaar:
https://github.com/aelsabbahy/GOnetstat/pull/2 should get rid of runc there, if we can merge that I'll update here and look at removing it here too if possible.

ripienaar:
I took a rough stab at removing runc from main goss code, so if https://github.com/aelsabbahy/GOnetstat/pull/2 is merged and I update the dependency we can remove the runc dependency entirely.

ripienaar:
OK, I finally made it through this trial.

Where I needed noncgo things I essentially implemented basic versions of what runc did, so this should be functionally equivelant now to what was before minus the runc dependency

Also udpated a few other things - the only exception is yaml from v2 to v3 which changed a few types

petemounce (approved):
LGTM!

Does this mean more implementation coverage on Windows as a side effect?

ripienaar:
thank you

PR #609, opened 2020-08-25
goss serve can now negotiate response's content-type via accept request header

petemounce:
##### Checklist

- [x] `make test-all` (UNIX) passes. CI will also test this
- [x] unit and/or integration tests are included (if applicable)
- [x] documentation is changed or added (if applicable)

### Description of change

Fixes #615.

`goss serve` will now respond to the `Accept` http header on incoming requests, and choose the `Content-Type` for the response if possible based on that.

This allows `goss serve` more flexibility at run-time; previously, if more than one output format was wanted, one endpoint process+port per output format would need to be set up.

This PR builds on top of #606 and #608, and is aimed at allowing a single `goss serve` to satisfy both of
* I want to ask about health in a variety of formats (e.g. my cloud automated health check doesn't care about the body, just the http status code, but when a person asks for health they want to see human-readable output)
* I want to scrape metrics to ingest into prometheus (see #607).
    * (The most common pattern here is to serve prometheus metrics via a separate path, so `http://{host}:{port}/metrics`)

This could be extended so that the response content-type can be driven from a querystring variable as well as a header, so it's more accessible to people in browsers compared to `curl` and machines.

PR #771, opened 2022-07-07
Prometheus output format

mhuysamen:
##### Checklist

### Description of change

Add a **prometheus** output format.

The format was copied from [DracoBlue](https://github.com/DracoBlue/goss-metrics-exporter).

Additions include:
**meta** are added to labels on the testoutput metric
**goss_test_count** metric added

There is no special negotiation (Accept Header) logic, so goss serve must be started with _-f prometheus_, or scraper must be configured to send Accept header as _application/vnd.goss-prometheus_.

Endpoint will respond with _text/plain; version=0.0.4_ as per Prometheus specification.

### DIFF
--- a/outputs/outputs.go
+++ b/outputs/outputs.go
@@ -30,6 +30,7 @@ var (
 		"json":          &Json{},
 		"junit":         &JUnit{},
 		"nagios":        &Nagios{},
+		"prometheus":    &Prometheus{},
 		"rspecish":      &Rspecish{},
 		"structured":    &Structured{},
 		"tap":           &Tap{},
--- a/outputs/prometheus.go
+++ b/outputs/prometheus.go
@@ -1,0 +1,115 @@
+package outputs
+
+import (
+	"fmt"
+	"io"
+	"regexp"
+	"strings"
+	"time"
+
+	"github.com/aelsabbahy/goss/resource"
+	"github.com/aelsabbahy/goss/util"
+	"github.com/fatih/color"
+)
+
+type Prometheus struct{}
+
+func (r Prometheus) ValidOptions() []*formatOption {
+	return []*formatOption{}
+}
+
+func (r Prometheus) Output(w io.Writer, results <-chan []resource.TestResult,
+	startTime time.Time, outConfig util.OutputConfig) (exitCode int) {
+
+	color.NoColor = true
+	testCount := 0
+	var skipped, failed int
+
+	const resultName string = "goss_result"
+	const durationName string = "goss_duration_seconds"
+	const promType string = "gauge"
+
+	fmt.Fprintf(w, "# HELP %s goss result\n", resultName)
+	fmt.Fprintf(w, "# TYPE %s %s\n", resultName, promType)
+	fmt.Fprintf(w, "# HELP %s goss duration\n", durationName)
+	fmt.Fprintf(w, "# TYPE %s %s\n", durationName, promType)
+
+	for resultGroup := range results {
+		for _, testResult := range resultGroup {
+			labels := make(map[string]string)
+			labels["property"] = testResult.Property
+			labels["resource_id"] = testResult.ResourceId
+			labels["resource_type"] = testResult.ResourceType
+			switch testResult.Result {
+			case resource.SUCCESS:
+				labels["result"] = "success"
+			case resource.SKIP:
+				labels["result"] = "skipped"
+				skipped++
+			case resource.FAIL:
+				labels["result"] = "failure"
+				failed++
+			}
+			testCount++
+
+			if testResult.Title != "" {
+				labels["title"] = testResult.Title
+			}
+			if len(testResult.Meta) > 0 {
+				for k, v := range testResult.Meta {
+					labels[ToSnakeCase(k)] = v.(string)
+				}
+			}
+
+			fmt.Fprintf(w, "%s{%s} %d\n", resultName, KeysString(labels), testResult.Result)
+			fmt.Fprintf(w, "%s{%s} %f\n", durationName, KeysString(labels), float64(testResult.Duration.Nanoseconds())/1000000000)
+		}
+	}
+
+	fmt.Fprint(w, "\n\n")
+	fmt.Fprint(w, "# HELP goss_failed_total goss failed tests\n")
+	fmt.Fprint(w, "# TYPE goss_failed_total gauge\n")
+	fmt.Fprintf(w, "goss_failed_total %d\n", failed)
+	fmt.Fprint(w, "# HELP goss_successful_total goss successful tests\n")
+	fmt.Fprint(w, "# TYPE goss_successful_total gauge\n")
+	fmt.Fprintf(w, "goss_successful_total %d\n", testCount-failed-skipped)
+	fmt.Fprint(w, "# HELP goss_skipped_total goss skipped tests\n")
+	fmt.Fprint(w, "# TYPE goss_skipped_total gauge\n")
+	fmt.Fprintf(w, "goss_skipped_total %d\n", skipped)
+	fmt.Fprint(w, "# HELP goss_test_count goss test count\n")
+	fmt.Fprint(w, "# TYPE goss_test_count gauge\n")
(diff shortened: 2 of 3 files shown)


aelsabbahy:
Can you explain at a high level the difference between this PR and #607?

I'm going to cross reference this PR with the open issue to hopefully get some reviews on this.

Apologies for missing this earlier, got it mixed up with #607 and didn't realize it was a new PR/approach.

mhuysamen:
This PR adds a new output format for goss, to present the results in Prometheus format.  This is not prometheus metrics for how well goss is running, but this is meant to be triggered via `serve`, to trigger a test run and output the test results in detail.

Unlike #607 this PR doesn't add a new endpoint.  goss already allows the endpoint to be changed with configuration, which serves the purpose to run as a prometheus collector.

Also, no additional external libraries are added, creating the prometheus output by hand.

aelsabbahy:
@petemounce would love your review on this since you spent a lot of time thinking about this problem.

I'm open to merging whatever the community agrees on.

t2d:
I would be happy to see this merged. It's basically what I asked for in https://github.com/aelsabbahy/goss/pull/607#issuecomment-1112366423

aelsabbahy:
Thanks for the feedback @t2d !

Tagging a few of the active participants from the previous chat to get a better gauge on consensus on this PR. Glad to see movement on this again. 

@harre-orz @Smithx10 @petemounce @timeu any thoughts?

Many thanks for doing this work and clear explanation of differences @mhuysamen

timeu:
@aelsabbahy : I skimmed through the code and from my point if view it looks good. The change is relatively small and the added labels would help us a lot with the prometheus alerting because it would make it easier to track flapping goss tests which happens sometimes in our environment.

petemounce:
Hey, long time! I've been dealing with some pretty hefty personal stuff.

I made some time to compare the two approaches in what I hope is an objective way:

| row | thing | 771 | 607 | comments | references |
|--|-------|-----|-----|----------|---|
| 1 | function | different | different | 771 uses gauges per iteration; 607 uses continuously incrementing counters. 607 means that functions like `rate` etc work. 771 means alerts would be (I think) harder to maintain since they'd be forced to be threshold-based and also they'd need to have their thresholds move in lock-step with the test-suite. (My assumption is that metrics exist to be alerted against, but that might not be everyone's starting point) |
| 2 | endpoint | n | y | It's a prometheus-ism to host a /metrics endpoint - so I elected to do the same. If this is unwanted (it was in [2020](https://github.com/aelsabbahy/goss/pull/607#issuecomment-716993446)), it's [1 line to remove in 607](https://github.com/aelsabbahy/goss/pull/607/files#diff-e17324709f13f4622759cc409a925ec69b1195ab5192cd5ad6156b00f900569eR28) | [Relevant discussion](https://github.com/aelsabbahy/goss/pull/607#issuecomment-717089483) |
| 3 | "deals with new concept of 'metrics' as a new feature/paradigm" | n | n | Was @aelsabbahy's feedback in 2020. If 771 is mergeable without this, then I think 607 is too? | [Relevant discussion](https://github.com/aelsabbahy/goss/pull/607#issuecomment-721302850) |
| 4 | includes integration tests | n | y | 771 doesn't, but easily could steal them from 607! | [Relevant diff](https://github.com/aelsabbahy/goss/pull/607/files#diff-5c297f761921b9217528855917c89263456bb6091464f9ee401f297693561162R95-R103) |
| 5 | includes unit tests | n | y | 771 doesn't. Some could be readily added via `prometheus.Output` asserting the contents of a buffer. (607 arguably needs fewer since it uses the upstream dependency and chooses to trust its own coverage - so only testing at the dependency-seam) | [607 relevant diff](https://github.com/aelsabbahy/goss/pull/607/files#diff-f6722a25b82fd84cd295dd8f5b535dc09524d4755a3a11ef4d30c9d62c2602ffR1-R76) |
| 6 | avoids missing metrics | n | y | (note: 771 is "I _think_ 'no' here; I could be mistaken?) Missing metrics happen when they're not initialised to 0; 771 is reliant on the inputs to be "complete", here. | [607 relevant diff](https://github.com/aelsabbahy/goss/pull/607/files#diff-849da76e77be2451c5dc14580961f6b50f12d6e7ed269fd7421922103ecd39c5R57-R67) |
| 7 | upstream dependency | n | y | The upstream dependency on the prometheus client means a) the endpoint content type; b) the output text format; should be "correct" (as in, need no maintenance beyond dependency bumps) | This _should_ mean less maintenance required over time due to upstream changes (though tbh prometheus is pretty damn settled at this point w.r.t. contract type things!) |
| 8 | process health metrics | n | y | via the upstream dependency, `goss serve` also includes golang process health metrics | [relevant discussion](https://github.com/aelsabbahy/goss/pull/607#issuecomment-717089483) |

My thoughts:
- #607 was blocked on 2, above. That was explained, I thought, and the explanation accepted as ok-not-a-blocker during [oct 2020 explanation](https://github.com/aelsabbahy/goss/pull/607#issuecomment-717089483) and [nov 2020 followup](https://github.com/aelsabbahy/goss/pull/607#issuecomment-721302850)
- #607 was blocked on 3, above. That remains. However - if 771 is mergeable without doing that, then I think 607 should be unblocked too.
    - FWIW, I think the block would be resolveable via follow-up changes if any users gave feedback that indicated the need. Looking back, and as a user - I don't think I had (or have) that need.

So I think - if 771 is considered mergeable today, I think 607 should also be - and if both are mergeable, I would prefer 607 to be based on the comparison above.

timeu:
@petemounce: Thanks for the detailed comparison. 
@aelsabbahy: Based on the above table and comparison I would also prefer #607 especially because it avoids missing metrics because I have run into issues when previous metrics didn't exist in regards to alerting. 
Also I think that a counter would make more sense than a guage for the reasons @petemounce mentioned.

timeu:
@petemounce: One difference I noticed between your PR and this one is that this one additonally adds the `Property` and `ResourceId` as a label to the metrics. I remember we discussed this and considered of making it configurable.  I guess this will be left for a potential followup ?

petemounce:
@timeu yes, that would be my intention; more scope later.

aelsabbahy:
@mhuysamen and @t2d do you agree with the above?

I know @petemounce time is limited, but if #607 is the way to go, it's a minor merge conflict away from being mergeable. I'm willing to drop the whole metrics part of the chat, it can be a future iteration/improvement.

Hopefully we're very close on this

mhuysamen:
I don't understand the gauge vs counter comments.  Counters imply that there is state stored between subsequent checks, which means it's only of use when goss is running in serve mode.

My patch only adds an output format for goss output.  It's not meant to serve goss metrics (how many checks were done, total failed/succeeded, etc).  This would have to be a specific implementation, exposing those generic metrics on a /metrics endpoint, and only useful in serve mode again, where counters do make sense.

Yes, I don't have tests, but as noted these can be added.

The missing values argument doesn't make sense either, since the output depends on which checks have been configured.  We cannot determine which checks will be performed/configured.  Dealing with missing metrics is more relevant in summaries, where counters for all the different types of checks are predetermined.

The gauge/counter selection could also be configurable, but adds complexity.

I don't particularly care which implementation is accepted, as long as someone is able to run goss and request a prometheus output.  Either via serve, or via validate.

petemounce:
My use-case is goss running in `serve` mode. It's used to implement health/readiness check endpoints for machines - the poller hits the endpoint every few seconds and the tests run. If tests fail 5x in a row, the machine is automatically killed and replaced.

607 returns the prometheus output for both `serve` (where it is available via the content-type negotiation that #609 implemented) and the one-shot `validate` command.

mhuysamen:
Then #607 being more complete, should be the one to be merged.

aelsabbahy:
#607 has been merged, so I am closing this for now.

Next goss release will have #607 in it.

Follow up PRs welcome, glad to see this finally get completed =)

Thank you all for your patience and contribution.. this was a pretty large effort and lots of opinions on the matter.

Thank you again!

Issue #770, opened 2022-07-06
kgoss - error: open .: is a directory


gaganyaan2:
**Describe the bug**

kgoss it does not copy the `goss.yaml` back to local directory

`error: open .: is a directory`

**How To Reproduce**

```bash
[root@home ~]# kgoss edit -i nginx
[INFO]: Parsing command line
[INFO]: using image: nginx
[INFO]: going to execute (may be blank): -- 
[INFO]: Preparing files to copy into container
[INFO]: Creating Kubernetes pod/container to test
++ /usr/local/bin/kubectl run kgoss-tester-17296 --image-pull-policy=Always --restart=Never --labels=app=kgoss-test '--output=jsonpath={.metadata.name}' --image=nginx --
+ id=kgoss-tester-17296
+ set +x
[INFO]: Waiting for container to be ready
pod/kgoss-tester-17296 condition met
[INFO]: Copying goss files into pod/container
[INFO]: Marking copied files as executable
[INFO]: Using pod/container: kgoss-tester-17296
[INFO]: When prompt appears you can run `goss add` to add resources
# goss a port 80
Adding Port to './goss.yaml':

tcp:80:
  listening: true
  ip:
  - 0.0.0.0


# 
Copying goss.yaml and goss_wait.yaml files back to local dir
[INFO]: Copied 'goss.yaml' from pod/container to '.'
tar: Removing leading `/' from member names
error: open .: is a directory
[INFO]: Deleting pod/container
```
**Expected Behavior**

Should copy the `goss.yaml` back to local directory as it does in dgoss

**Actual Behavior**

**Environment:**
 - Version of goss : goss version v0.3.18
 - OS/Distribution version (if applicable) : Fedora Linux 35

PR #773, opened 2022-07-10
Fixed kgoss kubectl cp issue

koolwithk:
Fixed https://github.com/aelsabbahy/goss/issues/770 issue

`kubectl cp <file-spec-src> <file-spec-dest>` **file-spec-dest** should be filename with path(optional) but It cannot be directory.

Reference for `kubectl cp`
-  https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#cp
- https://stackoverflow.com/questions/52407277/how-to-copy-files-from-kubernetes-pods-to-local-system


aelsabbahy:
Tagging @joshgav , if this looks good to you, I'll merge.

joshgav:
LGTM, the `get_pod_file` function only copies one file at a time anyway, and this makes that more resilient.

Thanks @koolwithk.

aelsabbahy:
Thank you both for your contributions,  greatly appreciate it!

PR #783, opened 2022-09-11
docs: backport gomega matchers from v4

### LINKED ISSUE #782: Support multiple valid values for the `file.mode` check
**Describe the feature:**

Support multiple valid values instead of just one for a check.

Eg:

```yaml
file:
  /foo:
    exists: true
    mode: "0644"
```

but if on a host, the file has the mode `0444`, I want to consider it as `good` also.

Instead of supporting only 1 value for `mode`, it could support a list of valid `modes`.

```yaml
file:
  /foo:
    exists: true
    mode:
      - "0644"
      - "0444"
```

I don't know any workaround except dropping the `file.mode` check.

samuel-phan:
Backport documentation on gomega matchers from v4.

##### Checklist

### Description of change

Add more documentation about the available gomega matchers so the end-users don't need to look at the Golang file https://github.com/aelsabbahy/goss/blob/master/resource/gomega_test.go .

If the end-user doesn't know Golang, it can be harder to understand them.

fix #782

### DIFF
--- a/docs/manual.md
+++ b/docs/manual.md
@@ -860,44 +860,116 @@ Count: 2, Failed: 0
 ```
 
 ## Advanced Matchers
-Goss supports advanced matchers by converting json input to [gomega](https://onsi.github.io/gomega/) matchers.
 
-### Examples
+Goss supports advanced matchers by converting YAML input to [gomega](https://onsi.github.io/gomega/) matchers.
 
-Validate that user `nobody` has a `uid` that is less than `500` and that they are **only** a member of the `nobody` group.
+### String Matchers
 
+These will convert the system attribute to a string prior to matching.
+
+* `'55'` - Checks that the numeric is "55" when converted to string
+* `have-prefix: pre` - Checks if string starts with "pre"
+* `have-suffix: suf` - Checks if string ends with "suf"
+* `match-regexp: '.*'` - Checks if string matches regexp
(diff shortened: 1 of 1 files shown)


aelsabbahy (changes requested):
Hello, thank you for taking the time to do this. However, I don't believe this documentation is accurate for the current version of goss. I highlighted the first two issues I saw, but from there it was clear that the rest of the documentation is most likely incorrect for the same reasons.

1. There are no type conversions, that is a new concept introduced in v4.
2. Some of the examples don't work because they reference matchers that don't exist in the current version of goss.

Unfortunately, updating the existing documentation will be a bit more involved process, backporting from v4 will be inaccurate in a lot of cases since a lot of the v4 work involves changes to the matcher logic.

If I was doing it, I would probably start with `gomega_test.go` as the source of truth and work from there.

Alternatively, I'm hoping that v4 gets released in a few months.. not the most ideal, but eventually will be rolled out. :man_shrugging:

samuel-phan:
I'll see if I have time to dig further, but truth is: I don't know Golang, so my reading can only be superficial.

aelsabbahy:
Closing this, as v0.4.0 is already in release-candidate: https://github.com/goss-org/goss/releases/tag/v0.4.0-rc.1

Thank you for the contribution though. Please feel free to open up new issues if the new documentation/goss version is not behaving as expected.

PR #785, opened 2022-09-20
Skip by type

ripienaar:
##### Checklist

### Description of change

When using Goss as a package it might be desirable to disable a certain resource type - mainly Command - so this adds a new utility `util.WithDisabledResourceTypes()` that can be used to disable a resource type.

Might not be desirable to change the `Validate()` signature but was best I could come up with, let me know if you have alternatives in mind.

Also deals with `ioutil` deprecation.

### DIFF
--- a/resource/validate.go
+++ b/resource/validate.go
@@ -50,6 +50,7 @@ const (
 
 type TestResult struct {
 	Successful   bool          `json:"successful" yaml:"successful"`
+	Skipped      bool          `json:"skipped" yaml:"skipped"`
 	ResourceId   string        `json:"resource-id" yaml:"resource-id"`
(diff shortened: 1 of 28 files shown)


resource/validate.go
```diff
@@ -31,6 +31,7 @@ const (
 
 type TestResult struct {
 	Successful   bool          `json:"successful" yaml:"successful"`
+	Skipped      bool          `json:"skipped" yaml:"skipped"`
```
ripienaar, line 34: This was an acidental add from some left over previous work, however I think it's useful as atm only way to find out if it was skipped is by the int value from `Result` and thats quite hard to use when looking at JSON representations of results.

aelsabbahy:
Thank you!

Issue #649, opened 2020-11-17
Research: Migrating to github actions or circleCI


aelsabbahy:
**Describe the feature:**

Find a new CI provider other than travis-ci.org. Queue times have been 45m-1hr as of late.

**Describe the solution you'd like**

A CI provider with reasonable queue time

Added bonus if easy to reproduce locally, Ex: Github Actions + https://github.com/nektos/act ?

**Describe alternatives you've considered**

Turning it off and back on again

petemounce:
Can I recommend https://cirrus-ci.org/ ? 

I'll happily do the work to experiment so we can compare side-by-side. 

I'm motivated to suggest it because I think it looks more capable than travis and github.

I'm also happy to be disagreed with and the experiment discarded if once I've wired it, it doesn't look like the best choice for some reason.

aelsabbahy:
Never heard of cirrus-ci to be honest. But the cirrus-cli looks interesting.

There are many layers to this issue.. and maybe it needs to be broken up into multiple tickets.

High-level thoughts on things that need changing:
1. Simplification of build/test - This needs a big re-write... and maybe deserves its own ticket
2. Ability to isolate 1 in a way that provides consistency between local and CI (as much as possible)
3. Finding the best CI provider to run 1,2 above


So For 2 and 3, my high-level thoughts:
The way goss is today doesn't reflect how I typically like to do CI/CD. I have a strong bias towards single responsibility as much as possible. So:
* CI engine should be nothing more than a generic orchestrator of things.
  * Basically nothing more than a list of "what to run" and "when to run", ideally those things are one-liners.
  * Support secrets management
  * The only other requirement is that it's fast, so caching & short queue time.
* CI build/deploy tools - This should be a set of tools that work consistently between local and CI. This is what the CI engine will call

For 1 - the current build/test/release scripts are a mess.. Very verbose, lots of bash code stringing it all together. High-level thoughts on changes:
* Need a simple target for all lint related stuff (maybe get rid of codeclimate for now.. not sure, I don't like that it's "remote"/SAAS)
* Simple target for building binaries
* Docker image tests should be limited to OS-specific tests, there's currently lots of duplication here that make this brittle. Ex: rpm vs deb, systemd vs initv
* Remote dependencies need to go. So DNS, httpbin, tinyproxy, etc.. maybe docker-compose is a good solution here.


Some tools of interest so far:
* https://github.com/nektos/act
* https://github.com/cirruslabs/cirrus-cli
[...]

PR #786, opened 2022-09-25
GitHub Actions based CI

petemounce:
First commit is just a very skeleton workflow.

This PR is really starting as just an experiment. I'm pretty familiar with GHA, and the experiment part will really be figuring out how to interact with the Make-based build in a nice UX way but also get nice, fast, modern, all-platforms-we-want, not-as-rate-limited-as-travis CI.

If this turns out to be quite easy, then my approach would be to get to parity with all the things Travis does today, _before_ doing anything that Travis does _not_ do that we might _also_ want. I'd want to minimise the amount of time there are 2 CIs in play.

That means changing the in-repo build-automation as little as possible, and if there are things necessary to change, then Travis must stay green. At all points, Travis outcome should agree with GHA outcome. If GHA differs, GHA is probably not quite there yet.

##### Checklist

Related: #649

(I don't think I will aim at everything that #649 wants, like researching other platforms - just getting off Travis).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,29 @@
+---
+name: ci
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+
+# Each of these can be read|write|none
+permissions:
+  actions: read
+  checks: read
+  contents: write
+  deployments: read
+  issues: read
+  packages: read
+  pull-requests: read
+  repository-projects: read
+  security-events: read
+  statuses: read
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@v3.0.2
+      - run: ci/build.sh


petemounce:
@aelsabbahy I think for this to start to work, you may need to [do some settings changes within the repository](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository).

I think for a contribution from a fork to run, you'd need to allow the (dangerous) `fork pull request workflows` permission. I advise against that; it's safer to only allow trusted contributors run CI for a tonne of reasons.

I think I can probably iterate on this in a not-fork repository that I own, then bring it back here.

petemounce:
Note nuance to integration tests being run. https://github.com/goss-org/goss/pull/856#issuecomment-1817913102

dklimpel:
I have been working on this a bit and experimenting. The biggest problem seems to be getting the centos7 images to run in github. These use systemd and the runability seems to be blocked by github.

I noticed that the jobs do not run ideally. IMHO the linitng should be removed from the test and run before the tests. But the linting is also defective and is not executed at all.

dklimpel:
IMHO this PR is superseded by a lot of other PR and can be closed.

aelsabbahy:
Yup, thank you both for the contributions so far. This has been a ton of work and great to see it progressing

PR #791, opened 2022-12-06
Change project name to goss-org

ripienaar:
Signed-off-by: R.I.Pienaar <rip@devco.net>

##### Checklist

### Description of change

Renaming the package and related files, also update dependencies.


ripienaar:
We should probably move the `go-ps` and `GOnetstat` dependencies also.

ripienaar:
Also worth noting that dependencies are starting to move to go1.18 syntax (gomega) we should consider bumping the minimum in go mod up to that level and update testing

ripienaar:
Also need to move the docker containers like `aelsabbahy/goss_centos7`

aelsabbahy (approved):

PR #793, opened 2022-12-23
Json schema to provide coding assistance for goss.yaml in IDEs

gberche-orange:
The json schema when configured in IDEs (see updated README.md for Intellij instructions) providers coding assistance for editing and authoring goss.yaml files

including: 
* completion
image
* inline doc in text or html format for long examples
image
* inspection such as required properties
image

##### Checklist

### Description of change

* adds a json schema file in yaml format
* addsa a sample goss.yaml file gathering most currently documented examples. Used for testing the json schema

Limitations:
- coverage for matchers is still low. 
- advanced matchers are not yet modeled
- templates likely can't be supported by json schema, and the source document isn't valid yaml

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,7 +172,13 @@ curl -H "Accept: application/vnd.goss-rspecish" localhost:8080/healthz
 
 ### Manually editing Goss files
 
-Goss files can be manually edited to use:
+Goss files can be manually edited to improve readability and expresssiveneess of tests.
+
+A [Json draft 7 schema](https://github.com/json-schema-org/json-schema-spec/blob/draft-07/schema.json) available in [docs/goss-json-schema.yaml](./docs/goss-json-schema.yaml) makes it easier to edit simple goss.yaml files in IDEs, providing usual coding assistance such as inline documentation, completion and static analysis.
+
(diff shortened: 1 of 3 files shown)


ripienaar:
This is amazing, I've wanted to do something like this myself but always seemed too much work.  Do you generate it somehow or hand built?

gberche-orange:
Thanks, this is hand built but with coding assistance from intellij it goes pretty fast.

aelsabbahy:
Reran Travis, if was an intermittent network issue that caused the failure.

aelsabbahy:
Does the auto complete work with all the gomega conditions?

ripienaar (approved):
LGTM

aelsabbahy (approved):
Thank you for taking the time to contribute this.

gberche-orange:
@aelsabbahy 
> This looks very cool. Out of curiosity, when doing templates or advanced matchers does it just fail completely or is it somewhat graceful?

Advanced matchers are gracefully ignored. I think it would be possible to refine the Json schema to include gomega supported matchers 

For templates, unfortunately the goss documents with templates are not valid json/yaml, so the plain json can't do much yet.

image

A useful workaround would be to find a way to actually render the templates into a valid goss rendered yaml document. This way, the json validation step could be applied after a template pre-rendering step, still providing useful feedback.

I'll submit a new PR with few fixes soon, and possibly would explore those ideas. 

Thanks for your quick feedback and merge (I'm just getting back from Christmas breaks)

gberche-orange:
Submitted https://github.com/goss-org/goss/pull/795 with rendering preprocessing step using the `render` command

PR #802, opened 2023-02-13
Allow log level to be set and filtering logs accordingly

obourdon:
##### Checklist

### Description of change

Some of the goss default logging messages might be seen as not very useful. This fix allow easy log level customisation and filtering

Reopen of PR [https://github.com/goss-org/goss/pull/641](https://github.com/goss-org/goss/pull/641)


util/config.go
```diff
@@ -73,6 +74,7 @@ func NewConfig(opts ...ConfigOption) (rc *Config, err error) {
 		DisabledResourceTypes: []string{},
 		ListenAddress:         ":8080",
 		LocalAddress:          "",
+		LogLevel:              "FATAL",
```
ripienaar, line 77: would it be possible to add a few helpers like `WithLogLevel(l string) and `WithLogWriter(w io.Writer)` or maybe WithLogger(c *Config)` to allow the case of embedding the code to be able to modify these new config items programatically?

ripienaar (commented):
generally pretty happy with this, `"log"` is a pretty bad logger lets hope the work to make a real logger in go standard gets up to speed. This is an improvement though.

Only real comment is if we can expose these new config items in the embedded scenarios

obourdon:
> Only real comment is if we can expose these new config items in the embedded scenarios

not sure I fully understand what you meant here

util/config.go
obourdon: Indeed, I also think this could be a very nice to have in the future but I will need some more time to work on this

Could it be possible to merge this as is for now and I'll work on a next more complete PR to address this

ripienaar: I am ok with merging as is. I am the main user of the embedded behaviours (that this enables) so can also propose a follow up PR for that

aelsabbahy:
> I am ok with merging as is. I am the main user of the embedded behaviours (that this enables) so can also propose a follow up PR for that

Thank you for the review.

@obourdon thank you for taking the time to submit this PR.. twice (apologies for that)

Issue #760, opened 2022-05-13
HTTP Header X-XSS-Protection does not match 


seffparker:
**Describe the bug**
The HTTP header `X-XSS-Protection` which present does not match, and we have to either match the string `X-Xss-Protection` or use case-insensitive regex pattern like `"/(?i:X-XSS-Protection: 1;mode=block)/"`

**How To Reproduce**
Apache conf:
```
  Header set X-XSS-Protection 1;mode=block
```
HTTP headers:
```
# curl -I localhost/index
HTTP/1.1 200 OK
Date: Fri, 13 May 2022 10:48:01 GMT
Server: Apache/2.4.41 (Ubuntu)
X-Powered-By: PHP/7.4.29
X-Content-Type-Options: nosniff
Content-Security-Policy: frame-ancestors 'self'
Strict-Transport-Security: max-age=31536000; includeSubDomains; preload
Vary: Origin
X-Frame-Options: SAMEORIGIN
X-XSS-Protection: 1;mode=block
Content-Type: application/json
```
Test rule:
```
http:
  http://localhost:{{ .Vars.http_port }}/index:
    status: 200
    timeout: 2000
    #method: GET
    headers:
    - "X-XSS-Protection: 1;mode=block"
```
**Expected Behavior**
Test passed.

**Actual Behavior**
```
.F..

Failures/Skipped:

HTTP: http://localhost:80/index: Headers: patterns not found: [X-XSS-Protection: 1;mode=block]

Total Duration: 0.044s
Count: 4, Failed: 1, Skipped: 0
```
**Environment:**
 - goss v0.3.16
 - Ubuntu 20

aelsabbahy:
Leaving this open. Making everything lowercase does fix the issue to some degree, but it is a breaking change for all existing users.

The challenge is go mutates the headers. I wonder if there's a way to skip that.. :thinking:

PR #814, opened 2023-06-25
V4 new

### LINKED ISSUE #220: Clarify kernel-param value as string

### LINKED ISSUE #262: Either drop file/size from add/autoadd OR add md5 and sha to it.
Need to make a decision on which one of these attributes are included in add and which are not:
* Size
* md5
* sha256

### LINKED ISSUE #317: Separate exit code for test parsing errors?
Thanks for all the hard work on this utility, I've found it really easy to use.

I use a simple bash script to run tests depending on different facter facts on hosts. However, I'd like to be able to tell failing tests from synctactically incorrect tests, e.g.:

```
#============ The test which should fail:
lan1:~ # cat goss_incorrect.yaml
port:
  tcp:22:
listening: false

lan1:~ # goss --gossfile goss_incorrect.yaml validate -f tap
Error: Could not parse resource Port:tcp:22

lan1:~ # echo $?
1

#============ The incorrect test:
lan1:~ # cat goss_failing.yaml 
port:
  tcp:22:
    listening: false

lan1:~ # goss --gossfile goss_failing.yaml validate -f tap
1..1
not ok 1 - Port: tcp:22: listening: doesn't match, expect: [false] found: [true]

lan1:~ # echo $?
1
```

I believe having a separate exit code for YAML parsing/template rendering would make scripting easier.

### LINKED ISSUE #416: Sort output in Documentation format
As I can see - documentation format is displaying different order for checks. This is making work with this format quite hard especially if results presents many checks / results. Having the same order allows quicker to see output when refresing web brwoser (to re-check status of check)

(As I do understand - I can be wrong - Goss displays results based on dictionary - which is not sorted by default - can we add some OrderDict function on Sort? - I am familar with python and this suggestion is based on python experience).

### LINKED ISSUE #443: Check for mount options doesn't work
I'm trying to check my mount points with goss. It reports an error about missing mount options but the options are there.
```
$ cat mounts.yml
mount:
  /zpool/media/download:
    exists: true
    opts:
      - rw
      - relatime
      - posixacl
    source: zpool/media/download
    filesystem: zfs
  /home:
    exists: true
    opts:
      - rw
      - noatime
      - compress=zstd
      - ssd
      - discard
      - subvol=@home
    source: /dev/mapper/cryptroot
    filesystem: btrfs
$ goss -g mounts.yml validate -f tap
1..8
ok 1 - Mount: /home: exists: matches expectation: [true]
not ok 2 - Mount: /home: opts: doesn't match, expect: [["rw","noatime","compress=zstd","ssd","discard","subvol=@home"]] found: [["rw","noatime"]]
ok 3 - Mount: /home: source: matches expectation: ["/dev/mapper/cryptroot"]
ok 4 - Mount: /home: filesystem: matches expectation: ["btrfs"]
ok 5 - Mount: /zpool/media/download: exists: matches expectation: [true]
not ok 6 - Mount: /zpool/media/download: opts: doesn't match, expect: [["rw","relatime","posixacl"]] found: [["rw","relatime"]]
ok 7 - Mount: /zpool/media/download: source: matches expectation: ["zpool/media/download"]
ok 8 - Mount: /zpool/media/download: filesystem: matches expectation: ["zfs"]
$ goss -v
goss version v0.3.7
```

Any idea what I'm doing wrong?

### LINKED ISSUE #483: Expose actual command stdout / stderr on the goss test result

### LINKED ISSUE #518: DNS checks should have hostname as attribute, not key
**Describe the feature:**
Currently, DNS checks specify the hostname to resolve as a key.  The server name to use for resolving is an attribute.  However, if you want to test the same item resolving on multiple DNS servers you cannot do it as the resolving hostname is a unique key in the heira.

As previously done with the exec resource, I would like to add a new attribute 'resolve' which optionally holds the hostname to resolve, overriding the key.   This will allow structures such as:

```
dns:
  primaryserver:
    resolve: myhost.company.com
    server: ns0.company.com
    resolveable: yes
  secondaryserver:
    resolve: myhost.company.com
    server: ns1.company.com
    resolveable: yes
```

which will allow you to check correct name resolution from multiple DNS hosts, which is not currently possible.

I also believe that keeping the keys as symbolic only is a better way to work, with all the test definition held in the attributes, as this allows better control when merging.

It would be possible to use different resolution targets on different DNS servers, but this seems untidy, and may cause automation or scaling problems.

### LINKED ISSUE #538: Feature Request: Improve gomega matchers

### LINKED ISSUE #578: add plugin support for goss
**Describe the feature:**

Sometimes, we want to use some 3rd tools (like ethotool and demicode) to do some test. On current stage we can use *Command* to test. But I think it's better to implement plugins so that user can define their own test method.

1. a plugin (in this situtation is binary file) should follow rules as below
   1. `./plugin name` will output plugin's name
   2. `./plugin options` will output a list with plugin's options for goss to call. Like a ethtool plugin should output `speed, is-bond, is-virtual` etc
   3. `./plugin test-method test-target` will output the test result. E.g: ethtool plugin, `./ethotool_plugin.sh is-virtual etho0` will output true
2. the plugin pass should be passed via cli args or config file, like: ./goss --plugins /usr/local/bin/ethotool_plugins then when goss start, it will create struct via `./plugin name` and `./plugin options` output.

For the implement of plugins, I have an idea. I use python to explain the code logic. We can talk about the implement and then I can create a new PR for this. 

```yaml
plugin:
  /tmp/test.sh:
      eth0:
        is-virtual: false
        is-bond: false
```

```python
import yaml
import pprint
import json
import get_command_output
with open("test.yml") as f:
        data = yaml.load(f)

def get_options(executable_file):
        return get_command_output("%s options" % executable_file).strip().split(",")

def get_output_by_option(executable_file, option, target):
        return get_command_output("%s %s %s" % (executable_file, option, target)).strip()

for executable_file in  data["plugin"]:
        for execute_target in data["plugin"][executable_file]:
                for k,v in data["plugin"][executable_file][execute_target].items():
                        print "k=%s, v=%s, result=%s" % (k,v,json.loads(get_output_by_option(executable_file, k, execute_target)))
                        print(get_output_by_option(executable_file, k, execute_target) == v)
```

```bash
check_virtual() {
        echo "false"
}

check_bond() {
        echo "false"
}

case $1 in
        is-virtual)
                check_virtual $2
                exit 0
                ;;
        is-bond)
                check_bond $3
                exit 0
                ;;
        options)
                echo "is-virtual, is-bond"
                exit 0
                ;;
        name)
                echo "ethtool-plugin"
                exit 0
                ;;
        *)
                echo "Usage: $0 {options|is-virtual|is-bond}"
                exit 0
                ;;
esac
```

### LINKED ISSUE #612: goss serve cache should separate results from output

### LINKED ISSUE #742: Ability to isolate tests
**Describe the feature:**
Running industry benchmarks testing these can mean ensuring that different lines may occur within the same file each with a unique test. These will have differing titles, meta data dependent on the test that takes place.

I am sure this type of testing that may occur is not unique to this but the is an example i am working with.

In the example below
It appears the filename used by the file test is the unique identifier and therefore only take the output of the final test that runs. These benchmarks for each line could be in the same file or different files. Only the last benchmark that runs using the goss test reports the output.

e.g.
file:
  /etc/somefilename:
    {{ if .Vars.test1 }}
    title: test1 
     exists: true
     contains:
     - test1_line
     meta:
       test1_ID
       other_data
   {{ end }}
   {{ if .Vars.test2 }}
    title: test2 
     exists: true
     contains:
     - test2_line
     meta:
       test2_ID
       other_data
   {{ end }}

If both are enabled the output would only show test2 information.

I have tested with different files, filenames etc.

Is there a better way that i can achieve this?

Ability to run the different requirements on using a like goss test and control that is able to report back all the  output and the associated data

Rewrite each control with the command test and rename that command title to make them unique. But using the actual goss test is more efficient and less calls than running the command test.

### LINKED ISSUE #762: ps output is not trimmed before checking for running process
**Describe the bug**
Goss doesn't trim the output of ps command which can cause false alarm.

**How To Reproduce**
```
root@sp1:~# ps auxww | grep ngcp-fax
faxserv+ 1539901  0.1  2.9 126932 90308 ?        Ss   09:54   0:02 ngcp-faxserver --cfg-file /etc/ngcp-faxserver/faxserver.conf
faxserv+ 1539959  0.0  2.2  82880 69800 ?        S    09:54   0:00 ngcp-faxserver --cfg-file /etc/ngcp-faxserver/faxserver.conf
root@sp1:~# 

root@sp1:~# ps -p 1539901 -o comm
COMMAND
ngcp-faxserver 
root@sp1:~# 

root@sp1:~# ps -p 1539901 -o comm | tail -1| hexdump -C
00000000  6e 67 63 70 2d 66 61 78  73 65 72 76 65 72 20 0a  |ngcp-faxserver .|
00000010
root@sp1:~# 

#### Steps to reproduce: 

root@sp1:~# cat /tmp/failure.yaml

process:
  ngcp-faxserver:
    running: true

root@sp1:~# goss -g /tmp/failure.yaml validate
F

Failures/Skipped:

Process: ngcp-faxserver: running:
Expected
    <bool>: false
to equal
    <bool>: true

Total Duration: 0.004s
root@sp1:~# 


root@sp1:~# cat /tmp/workaround.yaml

process:
  "ngcp-faxserver ":
    running: true

root@sp1:~# goss -g /tmp/workaround.yaml validate
.

Total Duration: 0.004s
root@sp1:~# 

```

**Expected Behavior**
Goss should trim ps output befome matching it to the value from YAML file.

**Actual Behavior**
Goss is not removing trailing space from ps output:
> 00000000  6e 67 63 70 2d 66 61 78  73 65 72 76 65 72 20 0a  |ngcp-faxserver .|
>........................................................................................................................ ^

**Environment:**
 - Version of goss: v0.3.14
 - OS/Distribution version (if applicable): Debian 11 (bullseye)

aelsabbahy:
##### Checklist

### Description of change

# Overview

This PR is a combination of v4 changes + a bunch of misc fixes/additions. My hope is to get this released sooner rather than later so future PRs/changes can be smaller and incremental.

## The major changes with v4 is the ability to compare different types and more advanced string matching

**note:** Manual has been updated to document most (if not all) new matchers, also `testdata/` folder has ~80% coverage of the matcher language.

For example:

```yaml
command:
  echo_test:
    # command that outputs JSON, but this could be anything, http response, command output, etc
    exec: |
      echo '{"string_value": "15"}'
    exit-status: 0
    stdout:
      # advanced string parsing
      gjson:
        # extract "string_value"
        string_value:
          and:
            # The value is numerically <= 20 (auto type conversion)
            - le: 20
            # The value is numerically 15
            - 15
            # Equal does a strict check, the types have to match, hence string_value is "15" but not 15
            - equal: "15"
            - not: {equal: 15}
```

This conversion also allows treating an io.Reader (memory efficient line-by-line parsing) as a whole string (when compared with a string). This has the benefit of showing the full command output, which was a long-requested feature. For example:

```
$ cat goss.yaml
command:
  echo_test:
    exec: |
      echo 'hello world'
    exit-status: 0
    stdout: |
      goodbye world

$ goss v
.F

Failures/Skipped:

Command: echo_test: stdout:
Expected
    "hello world\n"
to equal
    "goodbye world\n"

Total Duration: 0.001s
Count: 2, Failed: 1, Skipped: 0

```

# Changes

* Added: transforms support (closes #538)
  * Allows comparison of string to int, ex: ensuring a systctl output below/above a specific number (closes #220)
  * Allows JSON parsing/validation support (closes #578)
  * Allows comparing command output to a string, this will display the failed output on failure (closes #483)
  * `-o include_raw` will show the non-transformed value
* Added: Some new matchers
  * `equal` matcher added to do strict type comparison (e.x. string to string)
  * `contain-substring`
  * `have-patterns` - This mimics the existing default behavior of file contents and command output
  * `contain-elements` - checks that an array value matches a set of matchers
* Added: runlevels support for service
* Added: VfsOpts to mount check, (closes #443)
* Added: option to sort output (closes #416)
* Added: All resources now support key override (closes #518, closes #742)
* Added: Use exit code 78 if test file is unparseable (closes #317)
* Changed: removed alpha from non-linux binary names
* Changed: StartTime is no longer calculated from goss start. Calculated from first test start, this allows accurate reporting when showing a cached result
* Changed: file.contains to file.contents
* Changed: Completely re-worked matchers and shared output logic
* Changed: Cache test results in serve instead of output (closes #612)
* Changed: RPM Version is now full EVR (`-qf "%|EPOCH?{%{EPOCH}:}:{}|%{VERSION}-%{RELEASE}`), this opens the door for proper rpm version comparison in the future
* Added: Significantly improved test coverage of matcher, output code
* Added: bench output format
* Changed: Doc to reference `cat -E /proc/<PID>/comm` to avoid whitespace issues (closes #762)
* Changed: Removed file.Size from autoadd (closes #262)
* Added: `matcher.as-reader` this allows the matcher string to behave as an io.Reader output. Can be useful for testing
* Removed: `json_online` output format, use: `goss v -f json -o oneline`

# Decisions

##  Should `-o include_raw` be enabled by default, and the option becomes `-o exclude_raw`

```
$  cat goss.yaml
command:
  echo_test:
    exec: |
      echo '{"string_value": "15"}'
    exit-status: 0
    stdout:
      gjson:
        string_value:
          le: 5
```

Without include_raw:

```
$ goss v
.F

Failures/Skipped:

Command: echo_test: stdout:
Expected
    15
to be numerically le
    5
the transform chain was
    [{"gjson":{"Path":"string_value"}},{"to-numeric":{}}]

Total Duration: 0.001s
Count: 2, Failed: 1, Skipped: 0
```

With `-o include_raw`:

```
$ goss v -o include_raw
.F

Failures/Skipped:

Command: echo_test: stdout:
Expected
    15
to be numerically le
    5
the transform chain was
    [{"gjson":{"Path":"string_value"}},{"to-numeric":{}}]
the raw value was
    "{\"string_value\": \"15\"}\n"

Total Duration: 0.001s
Count: 2, Failed: 1, Skipped: 0

```

# reverted (defer to later)

These were changes living on the v4 branch. However, they are either bad fixes, or need more polish before release

* Changed: http.headers from io.reader -> array
    * This is a good discussion point on whether headers should be treated like an io.Reader or array. There are pros/cons to this.. however, I'm avoiding the breaking change for now.
* Changed: Convert headers to lowercase. #760
    * Not sure how I like this solution, it breaks existing tests.. and not sure it provides too much value. Will discuss on #760 prior to rolling out the change
* Changed: Migrate from go-ps to gopsutil for better process detection (#597)
    * I was hoping this woulud fix some of the issues with go-ps.. and it does. However, it reduces the ability to discover the process name for the user. The logic for gopsutil is [here](https://github.com/shirou/gopsutil/blob/fb83fb22d68f468adf2a01a9ed4963e7d980cfd7/process/process_linux.go#L837-L854). Deferring decision for now.


resource/command.go
```diff
@@ -12,15 +12,15 @@ import (
...
-	Command    string   `json:"-" yaml:"-"`
-	Exec       string   `json:"exec,omitempty" yaml:"exec,omitempty"`
-	ExitStatus matcher  `json:"exit-status" yaml:"exit-status"`
-	Stdout     []string `json:"stdout" yaml:"stdout"`
-	Stderr     []string `json:"stderr" yaml:"stderr"`
-	Timeout    int      `json:"timeout" yaml:"timeout"`
-	Skip       bool     `json:"skip,omitempty" yaml:"skip,omitempty"`
+	Title      string  `json:"title,omitempty" yaml:"title,omitempty"`
+	Meta       meta    `json:"meta,omitempty" yaml:"meta,omitempty"`
+	id         string  `json:"-" yaml:"-"`
```
ripienaar, line 17: no harm, so just a comment - the `json` and `yaml` tags here not needed as the marshallers wont include private struct members

ripienaar (dismissed):
Too big to realistically give a thorough review but overall this looks really great

.travis.yml
```diff
@@ -41,8 +41,8 @@ deploy:
   api_key:
     secure: ijNltjw/mIHIOx8vLZ6asUun3SbY7D+XZbs5NX8vcIv0jvOiwaaT1hqny7SQBHfGZzqHsYUSS/GYAYJdBqKFFfGmTZsl90hFT6D0RGdz9C71UVxNFX4wQ5KQ/WVvdMT2SrLymGvu9TvoU0VG8OWqWVdxSlUPf6qOTGAagrzg+Tbsbb6czeiG67mlBBL23XSlfMG1p45UxzvI41SZj2R3ElUb0hym1CrFaoC36PBGrb0x41TXzvd8J7cu6xDzgczYhnYQQZpS6f2YcqNV1z0f+P67EQqQiDWIIcK2jE/YG+RgM8cbpLMiMec8CDiwNCsejBA5EbVMlGJlODvBXT5NmMBeugueqfSHEfkl5qZTQG4AOAT7UsqbnM7r0NqzmaE5Lj90igvJK6rNsH1ZRe79WfSsTtuzlkkouHGvyoz0M8gnMSzpbbwoyIy+UT0hhPMoZvIpXfr43en5WkbkPKfop0p4Vjc8NGg0iD45q1JAvIVTtz/WvWTknM1P8e3u+TiDTaZkcJJmFaBqgaeLoWktOGfi54p9nhgQnSyBYt4PyvhWDQs7QFmX0BdKlqJCESvUOJTe1t6zJJsV7Gn/3sGCN7JUEwbnXTsCoMjjFFUvQdm0Ur7t7/2xU3kO+dyfqcdM/5SYFeppQcjHI0ckhI51mIoBTsJsGvaVwKKL1I4cyBU=
   file:
-    - release/goss-alpha-darwin-amd64
-    - release/goss-alpha-darwin-amd64.sha256
+    - release/goss-darwin-amd64
+    - release/goss-darwin-amd64.sha256
```
petemounce, line 45: Same for windows...?

ci/go-test.sh
```diff
@@ -3,7 +3,7 @@ set -euo pipefail
 
 command -v go
 
-go test -coverprofile="c.out" "${1}"
+go test -coverpkg=./... ./... -skip '^TestPrometheus' -coverprofile="c.out"
```
petemounce, line 6: Mind commenting the reason for the skip?

cmd/goss/goss.go
```diff
@@ -416,14 +415,14 @@ func addAlphaFlagIfNeeded(app *cli.App) {
 	if runtime.GOOS == "darwin" || runtime.GOOS == "windows" {
 		app.Flags = append(app.Flags, cli.StringFlag{
 			Name:   "use-alpha",
-			Usage:  fmt.Sprintf("goss is alpha-quality. Set to 1 to use anyway."),
+			Usage:  fmt.Sprintf("goss on macOS/Windows is alpha-quality. Set to 1 to use anyway."),
```
petemounce, line 418: Contradicts expectation from binary name? https://github.com/goss-org/goss/pull/814/files#diff-6ac3f79fc25d95cd1e3d51da53a4b21b939437392578a35ae8cd6d5366ca5485R44-R45

matcher_test.go
```diff
@@ -0,0 +1,84 @@
...
+	"strconv"
+	"strings"
+	"testing"
+
+	"github.com/goss-org/goss/util"
+	"github.com/stretchr/testify/assert"
+)
+
+var (
+	update = flag.Bool("update", false, "update the golden files of this test")
```
petemounce, line 21: Mind adding a comment for how and when to use this?

matcher_test.go
```diff
@@ -0,0 +1,84 @@
...
+func TestMatchers(t *testing.T) {
+	files, err := filepath.Glob(filepath.Join("testdata", "out_matching_*"))
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	for _, outFile := range files {
+		outFile := outFile
+		parts := strings.Split(outFile, ".")
+		tfName := fmt.Sprintf("%s.yaml", strings.TrimPrefix(parts[0], "testdata/out_"))
```
petemounce, line 38: What's tf in this context? Please expand or comment?

serve_test.go
```diff
@@ -268,14 +268,14 @@ func TestServeCacheNegotiatingContent(t *testing.T) {
 		logOutput.Reset()
 	})
 
-	t.Run("immediately re-request but different accept header, cache should be cold", func(t *testing.T) {
+	t.Run("immediately re-request but different accept header, cache should be warm", func(t *testing.T) {
```
petemounce, line 271: 🤕 on my part!

petemounce (commented):
I've reviewed all the parts that are _not_ the matchers; 58/135. I'll review the rest in a follow-up today.

petemounce (approved):
LGTM

petemounce:
> Should -o include_raw be enabled by default, and the option becomes -o exclude_raw

I would prefer `-o include_raw` to be the default, because I think that's a higher UX. I would be frustrated if I got a failure, had to re-run with `-o include_raw` and the 2nd run passed; indicating the test is flaky somehow. I think that it's best to give the data up-front, because in the context of machine-assertions (and especially when some can be remote-assertions like `http`, `addr`) the chance of flakes is higher than in "normal" test frameworks.

petemounce:
(Also - this is great work; thank you!)

aelsabbahy:
Awesome, thanks for the feedback. I'll address it over the coming week and hopefully this gets merged/released after that.

> I would prefer -o include_raw to be the default

This was my leaning as well, for all the reasons you mentioned.

ci/go-test.sh
aelsabbahy: Ooh, this one is easy.. it fails :smile: 

To be fair, I did run it on the main branch and it fails there too. Basically, the way that shell script was written it wasn't actually running those tests.. so they may have never worked.. not sure.. didn't try to check out the original commit to see.

I can take a look as a follow-up PR, or since you're familiar with the code, maybe give it a second look? Maybe I broke something a while back and not noticed.

cmd/goss/goss.go
aelsabbahy: I guess at this point I can remove the alpha flag altogether and just document that it's not as featureful/stable as Linux..

matcher_test.go
aelsabbahy: Added comment

matcher_test.go
aelsabbahy: "test file" renamed to specFile

matchers/be_numerically_matcher.go
```diff
@@ -0,0 +1,69 @@
...
+	"fmt"
+
+	"github.com/onsi/gomega/matchers"
+)
+
+type BeNumericallyMatcher struct {
+	fakeOmegaMatcher
+	Comparator string
+	CompareTo  []interface{}
+	//matchers.BeNumericallyMatcher
```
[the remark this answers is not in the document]
aelsabbahy: done

outputs/bench.go
```diff
@@ -0,0 +1,53 @@
...
+import (
+	"fmt"
+	"io"
+	"time"
+
+	"github.com/goss-org/goss/resource"
+	"github.com/goss-org/goss/util"
+)
+
+type Bench struct{}
```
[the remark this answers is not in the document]
aelsabbahy: Mentioned in docs, also mentioned it may not be stable in future releases.. Was using it for debugging and got tired of it not being part of goss =)

Logging may replace it, and it may get deleted in the future.

Thinking through it, part of me wonders if maybe I should remove it from this PR. Leaning towards that, but will sleep on it.

outputs/prometheus.go
```diff
@@ -73,10 +73,14 @@ func (r Prometheus) ValidOptions() []*formatOption {
...
 func (r Prometheus) Output(w io.Writer, results <-chan []resource.TestResult,
-	startTime time.Time, outConfig util.OutputConfig) (exitCode int) {
+	outConfig util.OutputConfig) (exitCode int) {
 	overallOutcome := resource.OutcomeUnknown
+	var startTime time.Time
 	for resultGroup := range results {
 		for _, tr := range resultGroup {
+			if startTime.IsZero() || tr.StartTime.Before(startTime) {
+				startTime = tr.StartTime
+			}
```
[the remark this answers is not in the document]
aelsabbahy: It's a loop finding the earliest start time. Before, startTime was calculated at program start, and endTime was when the output was printing. Now, it's calculated from the start time of the first test and endtime of the last test.

The reason for this is when caching testResults and rendering them in different formats, the start/end time needs to be stable. Without this, different output formats would show different times depending on when they were run.

serve_test.go
aelsabbahy: Nah, those tests were great for implementing this feature. Made my life so much easier.

aelsabbahy (commented):
Added some comments, will address more over the week.

cmd/goss/goss.go
aelsabbahy: Looking back at the history of this, I'll put it back on your plate.. What are your thoughts on alpha vs non-alpha.

Your PR renamed the files.. seems they're fairly stable from your comment 2-3 years back.


https://github.com/goss-org/goss/pull/671

ci/go-test.sh
petemounce: Hahaha, ok. Reasonably sure it ran and passed when introduced, but I also haven't gone back and checked.

Once this is merged, I propose to spend some time enhancing our build & test pipeline.

cmd/goss/goss.go
petemounce: I'd be fine with renaming each binary to remove the `alpha` segment. We still have the flag/env-var & message, and we have integration tests that run and pass for a lot of the functionality; missing or broken things can be iterated incrementally at need.

outputs/prometheus.go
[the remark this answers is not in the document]
petemounce: If that's not present as a comment someplace, please could it become present? I don't think I would have inferred that without the explanation.

outputs/bench.go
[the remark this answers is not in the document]
petemounce: 🤷 . I'd say leave it in if it was useful.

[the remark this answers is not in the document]
aelsabbahy: Slept on it, removing it for now. Initial logging support was added as part of PR a few months back. I have some thoughts on expanding logging to be a lot more useful. Once that story concludes, perhaps the bench output will not be useful.

I'll open up some issues after this PR is merged.

resource/file.go
```diff
@@ -65,8 +81,12 @@ func (f *File) Validate(sys *system.System) []TestResult {
 	if f.Filetype != nil {
 		results = append(results, ValidateValue(f, "filetype", f.Filetype, sysFile.Filetype, skip))
 	}
-	if len(f.Contains) > 0 {
-		results = append(results, ValidateContains(f, "contains", f.Contains, sysFile.Contains, skip))
+	if isSet(f.Contains) {
+		fmt.Fprintf(os.Stderr, "DEPRECATION WARNING: file.contains has been renamed to file.contents\n")
```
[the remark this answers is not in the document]
aelsabbahy: I'll add it to the upgrade notes

aelsabbahy:
* Removed bench output
* Temp solution to try to improve test flakiness

Addressed some (most/all?) of the feedback here.

@petemounce let me know if all looks good, if I don't hear back in a week, I'll merge it.

petemounce (dismissed):

aelsabbahy:
Modified install.sh (locked down version) to avoid breaking anyone. Merging, will create a release candidate release.

PR #825, opened 2023-07-28
Address a few shellcheck complaints about dgoss

mdb:
👋 Hello - this addresses a few `shellcheck` complaints with `dgoss`.

### Description of change

* double quote variables to prevent globbing and wordsplitting: https://www.shellcheck.net/wiki/SC2086
* use `-n` instead of `! -z`: https://www.shellcheck.net/wiki/SC2236


aelsabbahy:
Eyeball check looks good. Will run validate it locally and merge soon.

Thank you for this contribution.

PR #836, opened 2023-09-09
Added ctx object to all system struct's methods

aelsabbahy:
The idea behind this commit is to set the stage for the following features:

* logging - by adding context Values for logging purposes
* timeout/cancellation - move this up the stack and slowly add support for all resources where this makes sense

##### Checklist

### Description of change

Busy PR, lots of boring work to wire this up.. but hopefully this opens the door for some future features/flexibility.

One thing I wasn't 100% sure on is whether to have Validate() take ctx, or it instantiates it based on the yaml config.

My high level thoughts are:
* `Validate()` - creates a new context and sets the timeout based on the value in yaml (or default)
* `ValidateValue(..)` - adds some values to the ctx object for logging purposes
* `command.Stdout(ctx)` - the `system/` struct will then be able to log with the values from ValidateValue, and will timeout when the context is cancelled.

### Possible alternatives

Pass ctx only to the constructors, that would bring a lot of simplicity to the codebase as most of it would remain unchanged. Also, don't think any of the system resources have methods that require different timeouts.

I'll create another simpler PR with that approach and we can discuss how they compare.


aelsabbahy:
went with #837 instead

PR #837, opened 2023-09-09
Added ctx object to all system struct's constructors

aelsabbahy:
The idea behind this commit is to set the stage for adding additional logging metadata.

Added logging support for command as an example.

##### Checklist

### Description of change

This is the simpler version of #836 it goes against the guidance [here](https://go.dev/blog/context-and-structs) but it is a much simpler (and cleaner?) implementation.

Also, this approach didn't break any of the unit tests, like #836 did.


resource/command.go
```diff
@@ -48,14 +49,15 @@ func (c *Command) GetExec() string {
 }
 
 func (c *Command) Validate(sys *system.System) []TestResult {
+	ctx := context.WithValue(context.Background(), "id", c.ID())
```
ripienaar, line 52: Using context values for this is ok, but probably more idiomatic these days to use a structured logger like the (new) slog or something like logrus and then pass that down with key-vals added.

aelsabbahy: I'm very interested to see what ends up as idiomatic with the introduction of slog.

Seems like it's been a heavily debated topic: https://github.com/golang/go/issues/58243

I'm okay with pivoting to whatever becomes the norm once the dust settles a bit on slog and patterns emerge.

aelsabbahy:
Seems a bit of work piled up, going to start merging it.. since it's internal to the app we can revisit/change in a future PR.

PR #843, opened 2023-09-16
Added: warning when duplicate keys cause tests to be overwritten

### LINKED ISSUE #743: render silently drops overlapping directives
**Describe the bug**
Since the goss file format is a dictionary, overlapping keys do not work. This is [described in the manual](https://github.com/aelsabbahy/goss/blob/master/docs/manual.md#important-note-about-goss-file-format). However, when using multiple goss files, it is less obvious when this happens. Additionally I have a use case where it is helpful to break checks out into multiple logical categories that need to use the same directives to assert different things.

When `render` is used to combine files that have these types of overlaps, it does produce a valid file. But it silently drops values that appear first.

**How To Reproduce**
Below is a minimal example that reproduces the issue. Checking environment variable _can_ be worked around using `exec` and providing distinct names. But splitting multiple checks for a specific file across files can never work.

```shell
$ cat a.yaml 
command:
  env:
    exit-status: 0
    stdout:
      - HOME=
file:
  /etc/passwd:
    exists: true
    owner: root

$ cat b.yaml 
command:
  env:
    exit-status: 0
    stdout:
      - PWD=
file:
  /etc/passwd:
    exists: true
    group: root

$ cat combined.yaml 
gossfile:
  a.yaml: {}
  b.yaml: {}

$ ./goss-linux-amd64 -g combined.yaml render
file:
  /etc/passwd:
    exists: true
    group: root
    contains: []
command:
  env:
    exit-status: 0
    stdout:
    - PWD=
    stderr: []
    timeout: 0

$ echo $?
0
```

**Expected Behavior**
In an idea world, I'd like the checks to be merged if they do not conflict (which is true in the example & my actual use case). However, I understand that is much more complicated to implement and might be surprising behavior to others.

Being more realistic, I'd expect `render` to exit with an error because it can't produce a valid goss file that accurately represents all of the files being combined.

**Actual Behavior**
`render` produces a valid goss file that does **not** include all of the checks from the combined files with no indication that this happened.

**Environment:**
 - Version of goss: v0.3.16

aelsabbahy:
##### Checklist

### Description of change

Prints out warning messages when overlapping keys are detected, for example:

goss.yaml:
```yaml
command:
  example_test:
    exec: echo "hi"
    exit-status: 0
    stdout:
    - hi
    stderr: ""
    timeout: 10000
gossfile:
  goss_embed.yaml: {}
```

goss_embed.yaml:
```yaml
command:
  example_test:
    exec: echo "bye"
    exit-status: 0
    stdout:
    - bye
    timeout: 10000
```

```
$ goss -l warn v
2023-09-16T20:36:48Z [WARN] Duplicate key detected: 'command: example_test'. The value from a later-loaded goss file has overwritten the previous value.
..

Total Duration: 0.002s
```

Same with render:
```
$ goss -l warn render > /dev/null
2023/09/16 13:37:07 [WARN] Duplicate key detected: 'command: example_test'. The value from a later-loaded goss file has overwritten the previous value.
```

This warning is printed for both failing and passing tests, essentially it is informing the user that the tests were overwritten.. which may or may not be desired.

As a consequence of this change, I moved some existing log levels around to match what I would expect as a user:

  * `TRACE` - Detailed internal system activities useful for goss developers to debug.
  * `DEBUG` - Information useful for the goss user to debug.
  * `INFO` - General operational messages, useful for tasks where a more structured output is needed (e.g. `goss serve`).
  * `WARN` - Non-critical issues that may require attention, such as overwritten keys or deprecated features.

closes #743

### Questions

1. Given the above classification, should the default log level be set to INFO? Users who find it noisy can do `-l warn` or `-l error`

### DIFF
--- a/cmd/goss/goss.go
+++ b/cmd/goss/goss.go
@@ -75,7 +75,7 @@ func main() {
 	app.Flags = []cli.Flag{
 		cli.StringFlag{
 			Name:   "log-level, loglevel, L, l",
-			Value:  "FATAL",
+			Value:  "INFO",
 			Usage:  "Goss log verbosity level",
 			EnvVar: "GOSS_LOGLEVEL",
 		},
--- a/docs/manual.md
+++ b/docs/manual.md
@@ -229,13 +229,12 @@ service:
 * `--endpoint <value>`, `-e <value>` - Endpoint to expose (default: `/healthz`)
 * `--format`, `-f` - output format, same as [validate](#validate-v---validate-the-system)
 * `--listen-addr [ip]:port`, `-l [ip]:port` - Address to listen on (default: `:8080`)
-* `--loglevel level`, `-L level` - Goss logging verbosity level (default: `INFO`). `level` can be one of `TRACE | DEBUG | INFO | WARN | ERROR | FATAL`. Lower levels of tracing include all upper levels traces also (ie. INFO include WARN, ERROR and FATAL outputs).
-  * `TRACE` - Print details for each check, successful or not and all incoming healthchecks
-  * `DEBUG` - Print details of summary response to healthchecks including remote IP address, return code and full body
-  * `INFO` - Print summary when all checks run OK
-  * `WARN` - Print summary and corresponding checks when encountering some failures
-  * `ERROR` - Not used for now (will not print anything)
-  * `FATAL` - Not used for now (will not print anything)
+* `--loglevel level`, `-L level` - Goss logging verbosity level (default: `INFO`). `level` can be one of `TRACE | DEBUG | INFO | WARN | ERROR`. Lower levels of tracing include all upper levels traces also (ie. INFO include WARN and ERROR).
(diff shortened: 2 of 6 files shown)


aelsabbahy:
CC: @obourdon, @ripienaar,@plannigan, @GhuulCZ

Since you all have been involved with discussions around this.

plannigan:
Seems reasonable to me. :+1:

My preference would have been to actually error, this is much better than silently overwriting.

aelsabbahy:
@plannigan how about a layered approach

1. Default, user gets a warning if they're not aware of this behavior. (This pr)
2. User can pass `--log-level ERROR` to silence warnings.. perhaps they're intentionally using overrides. (This pr)
3. User wants to explicitly block this behavior, they pass `--strict` (new pr)

plannigan:
Yeah, that makes sense. Default to showing the warning. People can silence it if they want. People can also elevate it to an error if they want.

PR #850, opened 2023-10-11
make have-len more strict, only accept int

aelsabbahy:
##### Checklist

### Description of change

Follow up to #849, merged the PR too soon.


aelsabbahy:
nvm, json unmarshals numbers into floats. The merged code is the correct one.

PR #858, opened 2023-11-23
"socket" as one more valid filetype for "file" testing

fruit:
##### Checklist

### Description of change

Some files can be sockets. Goss detects already socket files correctly with filetype `socket`.


aelsabbahy:
Great catch, thanks!

PR #864, opened 2024-01-13
add more labels to prometheus output

### LINKED ISSUE #862: Add more labels to prometheus output format
**Describe the feature:**

The prometheus output format should include references to test cases in order to allow more fine grained alerting. The current implementation only contains a summary of test cases for each resource type and a summary across all tests. Since test cases can have different criticality levels, neither of those summaries are sufficient to decide whether a person should be called right now (potentially in the middle of the night) or not.

This has been previously discussed in #607 and my understanding is that it was postponed, but not rejected. There is a valid concern for storage cost, so I don't think this should/must be enabled by default, but rather should be an format option for the prometheus output.

The metric `goss_tests_outcomes_total` should contain additional labels to uniquely identify a single test case or a different metric should be introduced that does exactly that.

My current workaround is an annotation I placed on the prometheus alert for goss tests that links to a runbook which tells the poor soul who has to deal with this to do:

```shell
$ ssh ...
$ goss --gossfile ... validate
```

This returns the failing tests and they can decide whether to fix it right now or go back to bed.

sebhoss:
##### Checklist

### Description of change

This adds a `resource_id` label to each test-specific metric if the `verbose` format option was enabled. This assumes that the combination of resource ID and resource type is unique.

Since we only have access to format options in the `Output` method, we are creating metrics during the first run of that method.

Metrics can only be registered once for any given set of labels, so we create a custom registry and use the existence of that as a check whether we need to create/register metrics. This has the added benefit of simplifying the test a tiny bit, and it exposes even less metrics on the /healthz endpoint.

fixes #862

Pinging @petemounce @aelsabbahy

### DIFF
--- a/docs/manual.md
+++ b/docs/manual.md
@@ -270,7 +270,7 @@ The `application/vnd.goss-{output format}` media type can be used in the `Accept
   * `silent` - No output. Avoids exposing system information (e.g. when serving tests as a healthcheck endpoint)
 * `--format-options`, `-o` (output format option)
   * `perfdata` - Outputs Nagios "performance data". Applies to `nagios` output
-  * `verbose`  - Gives verbose output. Applies to `nagios` output
+  * `verbose`  - Gives verbose output. Applies to `nagios` and `prometheus` output
   * `pretty`   - Pretty printing for the `json` output
   * `sort`     - Sorts the results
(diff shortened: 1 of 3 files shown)


petemounce (commented):
This LGTM; thanks. I'll await @aelsabbahy for approve.

I'll take care of adding documentation when #856 merges; I don't want to cause @noirbizarre a conflict.

aelsabbahy (approved):
Approved, merging.

Issue #602, opened 2020-07-18
Allow setting environment variables for command tests


kasimon:
**Describe the feature:**
Add the ability to specify environment variables for command checks.

**Describe the solution you'd like**
```
command:
  vault status:
    exit-status: 0
    environment:
      VAULT_ADDR: "https://....:8200"
      VAULT_CACERT: "/etc/ssl/my-internal-ca.pem"
    timeout: 10000
```

**Describe alternatives you've considered**
Create a wrapper script that sets the environment variables and is called by goss. Possible but not as elegant as if this feature was provided by goss itself.

aelsabbahy:
Another way to do it today if you want to avoid a wrapper script is.

```yaml
exec: "VAULT_ADDR=.. VAULT_CACERT=.. vault status"
```

That said, this sounds like a good enhancement to have, marking it as approved.

PR #871, opened 2024-01-27
Command optional exec style

### LINKED ISSUE #870: Cannot write command tests without shell (sh)
**Describe the bug**

Goss command test will execute the command wrapped with a `sh -c "<the command>"`.
This assumes `sh` executable exists.
When testing distroless containers with a single binary, goss cannot execute the binary.

*Dockerfile*

```Dockerfile
FROM scratch
ADD figlet /
ENTRYPOINT ["figlet"]
```

*goss.yaml*
```
Command: figlet: exit-status:
Error
    exec: "sh": executable file not found in $PATH
```

**How To Reproduce**

Create a single statically compiled binary, or a binary with the needed libraries.
For me it is `figlet`.

```yaml
command:
  figlet:
    exec: figlet test
```

**Expected Behavior**

Goss to directly execute (without a shell in the middle) `figlet` (arg 0) with single argument `test` (arg 1).

I think that goss giving a shell, should be configurable - and optional.
Perhaps a new attribute `shell` that defaults to `sh`, and if explicitly set to nothing `shell:`, then goss will exec directly the command, rather than running it in a shell.

**Actual Behavior**

Goss attempts to execute `sh`, which doesn't exist, so fails.

**Environment:**
 - Version of goss: goss version v0.4.4
 - OS/Distribution version (if applicable): Alpine Linux

berney:
Fixes #870

##### Checklist

- [x] `make test-linux-all` (UNIX) passes. CI will also test this
- [ ] Windows platform
- [x] Update goss yaml schema
- [x] Update goss JSON schema

### Description of change

This change allows specifying a command to be executed exec style without a shell.
This change is backwards compatible, and existing goss configuration files should continue to work as expected.
This change is needed for operating in environments where a shell doesn't exist, such as testing a distroless docker image.
The syntax is inspired by the Dockerfile `RUN`, `ENTRYPOINT`, `CMD` instructions that can take either a string (shell style), or a list (exec style).  This allows specifying arguments that contain spaces, etc, as the argument list is explicit.

Shell syntax (unchanged from before):

```yaml
command:
  figlet:
    exit-status: 0
    exec: figlet test
```

This would be wrapped and executed as `sh -c "figlet test"`

Exec syntax (this change introduces):

```yaml
command:
  figlet:
    exit-status: 0
    exec:
      - figlet
      - test
```

This would be executed directly, without a shell, as `figlet test`, where `figlet` is arg 0 and `test` is arg 1.

The above could use alternative yaml syntax:

```yaml
command:
  figlet:
    exit-status: 0
    exec: [figlet, test]
```

### DIFF
--- a/resource/command_test.go
+++ b/resource/command_test.go
@@ -4,7 +4,7 @@ import (
 	"reflect"
 	"testing"
 
-	"gopkg.in/yaml.v3"
+	"go.yaml.in/yaml/v3"
 )
 
 func TestCommandGetExec(t *testing.T) {
--- a/resource/command_test.go
+++ b/resource/command_test.go
@@ -1,0 +1,62 @@
+package resource
+
+import (
+	"reflect"
+	"testing"
+
+	"gopkg.in/yaml.v3"
+)
+
+func TestCommandGetExec(t *testing.T) {
+	tests := []struct {
+		name  string
+		input string
+		want  any
+	}{
+		{
+			name:  "no exec falls back to the id",
+			input: "foo:\n  exit-status: 0\n",
+			want:  "foo",
+		},
+		{
+			name:  "string exec is shell style",
+			input: "foo:\n  exit-status: 0\n  exec: echo hi\n",
+			want:  "echo hi",
+		},
+		{
+			name:  "list exec is exec style",
+			input: "foo:\n  exit-status: 0\n  exec: [/bin/echo, hello world]\n",
+			want:  []string{"/bin/echo", "hello world"},
+		},
+		{
+			name:  "bool scalar exec keeps legacy string behavior",
+			input: "foo:\n  exit-status: 0\n  exec: true\n",
+			want:  "true",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			var m CommandMap
+			if err := yaml.Unmarshal([]byte(tt.input), &m); err != nil {
+				t.Fatalf("unexpected error: %v", err)
+			}
+			cmd, ok := m["foo"]
+			if !ok {
+				t.Fatalf("expected foo resource, got %+v", m)
+			}
+			got := cmd.GetExec()
+			if !reflect.DeepEqual(got, tt.want) {
+				t.Fatalf("GetExec() = %#v, want %#v", got, tt.want)
+			}
+		})
+	}
+}
+
+func TestCommandInvalidExec(t *testing.T) {
+	input := "foo:\n  exit-status: 0\n  exec: {a: b}\n"
+	var m CommandMap
+	if err := yaml.Unmarshal([]byte(input), &m); err == nil {
+		t.Fatalf("expected error for mapping exec, got %+v", m)
+	}
+}
--- a/system/command.go
+++ b/system/command.go
@@ -18,7 +18,7 @@ type ContextKey struct{}
 var CommandIDKey = ContextKey{}
 
 type Command interface {
-	Command() string
+	Command() util.ExecCommand
 	Exists() (bool, error)
 	ExitStatus() (int, error)
 	Stdout() (io.Reader, error)
@@ -27,7 +27,7 @@ type Command interface {
 
 type DefCommand struct {
 	Ctx        context.Context
-	command    string
+	command    util.ExecCommand
 	exitStatus int
 	stdout     io.Reader
 	stderr     io.Reader
@@ -36,12 +36,27 @@ type DefCommand struct {
 	err        error
 }
 
-func NewDefCommand(ctx context.Context, command string, system *System, config util.Config) Command {
-	return &DefCommand{
+// NewDefCommand accepts a command specified either shell style (string) or exec
+// style ([]string). The concrete type is validated here; an unexpected type is
+// recorded as an error on the returned Command so it surfaces as a failing test
+// result rather than a panic. Values loaded from a gossfile are already
+// validated by util.ExecCommand's unmarshalers, so this only guards against
+// programming errors.
+func NewDefCommand(ctx context.Context, command any, system *System, config util.Config) Command {
+	c := &DefCommand{
 		Ctx:     ctx,
-		command: command,
 		Timeout: config.TimeOutMilliSeconds(),
 	}
+	switch cmd := command.(type) {
+	case string:
+		c.command = util.ExecCommand{CmdStr: cmd}
+	case []string:
+		c.command = util.ExecCommand{CmdSlice: cmd}
+	default:
+		c.err = fmt.Errorf("command type must be either a string or a list of strings, got %T", command)
+		c.loaded = true
+	}
+	return c
 }
 
 func (c *DefCommand) setup() error {
@@ -50,7 +65,16 @@ func (c *DefCommand) setup() error {
 	}
 	c.loaded = true
 
-	cmd := commandWrapper(c.command)
+	var cmd *util.Command
+	switch {
+	case c.command.CmdStr != "":
+		cmd = commandWrapper(c.command.CmdStr)
+	case len(c.command.CmdSlice) > 0:
+		cmd = util.NewCommand(c.command.CmdSlice[0], c.command.CmdSlice[1:]...)
+	default:
+		c.err = fmt.Errorf("empty command")
+		return c.err
+	}
 	err := runCommand(cmd, c.Timeout)
 
 	// We don't care about ExitError since it's covered by status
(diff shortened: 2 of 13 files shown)


berney:
@aelsabbahy I've got this working on Linux for shell and exec style commands. I've made it draft as I'm unsure of a few things. If you have the time I'd appreciate some feedback.

Once I have this change in acceptable form, I want to add:

1. Ability to specify more os/exec attributes like environment variables, re #602
2. Setting current working directory

This would expose more of https://pkg.go.dev/os/exec#Cmd to goss Command resource.
I could open separate PRs or bundle the changes in the one. Do you have a preference?

So far, I've just worked on posix platform.
I will work on Windows when I'm on the right track.

The existing code use generic code, and can only pass a single string from resource to system.
In the change I made, I have it pass a struct, which should either have a string or a slice.
I added code to unmarshal this struct from json/yaml.

I couldn't work out how to properly using `genny`, so I have manually updated `resource/resource_list.go`.
I know this is wrong way to do it.
Any tips on proper way to get this to work?

Later when I expand on this work, I'll want to be passing more data, env vars and working directory.

Also, in system/system.go it imports util as util2, I don't understand why it needs to alias it, why can't it default to `util`?

berney:
I'm looking into the ci/build.sh and test failures with the unmarshal errors.

berney:
Fixed, so passes tests now.

aelsabbahy:
I'm a bit mixed on this UX. I understand it, it makes sense, is very elegant, which makes me love it. However, it does mean that command is a string or an array of strings.. which no other resource does, and it can't be used with `goss add ..`, not sure what other implications there are if any.

The `resource/resource_list.go` part does need to be figured out since that contract is the same across all resources, with all of them taking a string.

I don't have an easy fix for you to be honest. Is a solution possible if NewCommand took an interface and then type asserted it to a string or []string? I'm on my phone, so didn't dig deep into the code to see if it's viable or not.

Assuming we can get the codebase to corporate without any odd foot guns, this does seem to be an extremely elegant solution.

berney:
> However, it does mean that command is a string or an array of strings.. which no other resource does, and it can't be used with goss add .., not sure what other implications there are if any.

`goss add  command ...`  still works in that it has the same old behaviour, it will be the shell style.
At the moment the, only way to get exec style is to manually edit the yaml.
Do you think this is acceptable? Or do you think there should be a way to add exec style commands with `goss add command` as well?

> Is a solution possible if NewCommand took an interface and then type asserted it to a string or []string?
At the moment I made it take a struct instead of a string:

[`func NewDefCommand(ctx context.Context, command util.ExecCommand, system *System, config util.Config) Command {`](https://github.com/goss-org/goss/pull/871/files#diff-830f39b86d4a5fba941767c7f228921d023957504bb9896759493f66c7bbf200R33)

and the code looks at the attributes of the struct.
I then did a dirty temporary workaround for the autogenerated code.

Perhaps an interface and type assertions like you suggest would work too, to make all the code use the same function prototype.

aelsabbahy:
> Or do you think there should be a way to add exec style commands with goss add command as well?

Reflecting on this.. I don't think it's necessary. It's fine if adding by cli only uses shell.. and only way to handle exec is by editing the YAML.

Let's try the interface change or any approach that won't require manual edits to resource_list.go as all the code in there is generated and similar across all resource types.

berney:
@aelsabbahy I've implemented the `interface{}` idea. This fixes the problem I had with `genny` and the `resource/resource_list.go`.

The `interface{}` means that any type can be passed, such as string, or a string slice `[]string`, or any other type (which would be unexpected). At run time I check the types. So, I changed the function prototype to return an error. But in most cases I don't actually check the error (I'll come back to this).

All goss checks uses only strings, except command which can be either a string or a string slice (for the exec style, the point of this PR). In the `resource/command.go` I do check if the `sys.NewCommand` had an error. I wasn't sure of best way to handle the error, I create a `TestResult` with `Result: FAIL`. I'm not sure if there's a better or more goss idiomatic way to do this.

I can update the rest of the resources to check if there was an error, rather than ignore it.

I think the genny code could be replaced with native Go generics introduces in Go v1.18, and possibly other parts of the code could use generics too. Thus far I think this is the only bit of code using them: https://github.com/goss-org/goss/blob/e73553f9c3065ac297499dafb4f8abef6acb24ad/goss_config.go#L114

I think generics would be a better solution because IIUC the type checking would happen at build time, based on type constraints, rather than at runtime. Because the type checking would happen at build time, runtime checks shouldn't be needed, and so checking for returned errors wouldn't be needed either.

With the `interface{}` and runtime checking, programming mistakes could result in successful builds, but bugs at runtime, and these might not be covered by automated testing. For example, I had a bug at first where I returned the wrong type in the `c.id` case, but luckily the tests covered it: https://app.travis-ci.com/github/goss-org/goss/jobs/617952958. I fixed this in c129fd41c0f759e8e4a557c878c623ceaabf4195.

1. What do you think about rewriting `resource_list.go` to use Go generics?
2. Should I do that in a separate PR or in this one?
3. Do it first or get this merged and then do the Go generics refactor? (I'm thinking generics first).

After the implementation is solidified, I'll work on the remaining tasks (Error handling for all resource types (if needed), Unit/Integration Tests, Docs, json and yaml schemas).

aelsabbahy:
Looping back on this, any update?

javierfreire:
Any news on this feature? I'm interested. I can try to finish it if needed.

aelsabbahy:
If we don't hear back from the original author, feel free to finish this PR. I don't think anything has changed since my last comment on here providing feedback.

berney:
I've picked this back up again. I've made changes based off #1016 thanks to @javierfreire.
In the commit I gave him co-credit.
I believe this resolves all the feedback from this PR and #1016

resource/command_test.go
```diff
@@ -0,0 +1,62 @@
+package resource
+
+import (
+	"reflect"
+	"testing"
+
+	"gopkg.in/yaml.v3"
```
kgaughan: Could you use this instead?
```suggestion
	"go.yaml.in/yaml/v3"
```
I have #1084 prepped to replace it elsewhere, but I just need to get a review for it. I can review and merge other people's PRs, just not my own. I'd like to get goss moved to a maintained YAML library generally.

system/command.go
```diff
@@ -50,7 +65,16 @@ func (c *DefCommand) setup() error {
...
 
-	cmd := commandWrapper(c.command)
+	var cmd *util.Command
+	switch {
+	case c.command.CmdStr != "":
+		cmd = commandWrapper(c.command.CmdStr)
+	case len(c.command.CmdSlice) > 0:
+		cmd = util.NewCommand(c.command.CmdSlice[0], c.command.CmdSlice[1:]...)
+	default:
+		c.err = fmt.Errorf("empty command")
```
kgaughan: Could you declare a static error at the top-level of the file with `errors.New` and use it here instead?

util/command.go
```diff
@@ -2,12 +2,77 @@ package util
...
+// strings (exec style). Anything else is an error.
+func (e *ExecCommand) UnmarshalJSON(data []byte) error {
+	// Try shell style first.
+	if err := json.Unmarshal(data, &e.CmdStr); err == nil {
+		return nil
+	}
+	// Fall back to exec style.
+	e.CmdStr = ""
+	if err := json.Unmarshal(data, &e.CmdSlice); err != nil {
+		return errors.New("command must be a string or a list of strings")
```
kgaughan: Could you move this to the top-level of the file and make it static?

util/command.go
```diff
@@ -2,12 +2,77 @@ package util
...
+func (e *ExecCommand) UnmarshalYAML(unmarshal func(any) error) error {
+	// Try shell style first. A bool/int scalar decodes into the string, which
+	// preserves the long-standing behavior of `exec: true` meaning `exec: "true"`.
+	if err := unmarshal(&e.CmdStr); err == nil {
+		return nil
+	}
+	// Fall back to exec style.
+	e.CmdStr = ""
+	if err := unmarshal(&e.CmdSlice); err != nil {
+		return errors.New("command must be a string or a list of strings")
```
kgaughan: Same here: could you reuse a top-level declaration of this error?

util/command_test.go
```diff
@@ -0,0 +1,168 @@
+package util
+
+import (
+	"encoding/json"
+	"reflect"
+	"testing"
+
+	"gopkg.in/yaml.v3"
```
kgaughan: Ditto:
```suggestion
	"go.yaml.in/yaml/v3"
```

kgaughan (commented):
Nice work! There are a few fix-ups I'd like to get made regarding making some errors static and the code switched to a maintained YAML library.

berney:
I've also updated docs/platforms.md to include the exec field for command, and listed Linux, MacOS, and Windows and fully supported. I've tested this PR on all 3 platforms and both shell style and exec style work.

kgaughan (approved):

PR #873, opened 2024-02-07
documentation: dgoss on mac needs DGOSS_TEMP_DIR

rasschaert:
##### Checklist

### Description of change

On macOS, dgoss needs a different temporary directory than /tmp. This has been reported and fixed already in https://github.com/goss-org/goss/issues/329 and https://github.com/goss-org/goss/issues/624.
The workaround of setting `DGOSS_TEMP_DIR=~/tmp` works well, but could be better documented.
I did lose some time figuring this out today because the installation instructions don't tell you that this is needed.


aelsabbahy (approved):
Thanks for the documentation update.

If this is a consistent problem on macos, perhaps dgoss should detect when it's running on macos and choose a different temp dir by default?

aelsabbahy:
Cc: @petemounce

PR #874, opened 2024-02-07
IDE json schema: skip is not mandatory for process tests

rasschaert:
### Description of change

The json schema used by IDEs claims that "skip" is a mandatory property of a process test.

image

Adding the skip property causes goss to fail with this error message:
```
Error: Invalid Attribute for Process:caddy: skip
```

I can only presume that at some point the skip attribute was supported and required, but was later removed. It seems the json spec file wasn't modified with this change.

### DIFF
--- a/docs/goss-json-schema.yaml
+++ b/docs/goss-json-schema.yaml
@@ -385,16 +385,12 @@ definitions:
   processTest:
     required:
       - running
-      - skip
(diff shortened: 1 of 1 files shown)


aelsabbahy (approved):
Thanks for this, skip was never mandatory, the json schema was wrong.

PR #892, opened 2024-03-28
add a default user-agent header

### LINKED ISSUE #882: Set a custom user agent string for http checks
**Describe the feature:**

- currently when goss performs a http check like below, `"Go-http-client/1.1"` will be sent to the server as the user agent

```
http:
  https://www.google.com:
    status: 200
    allow-insecure: false
    no-follow-redirects: false
    timeout: 1000
    username: ""
    password: ""
    ca-file: ""
    cert-file: ""
    key-file: ""
    proxy: ""
    skip: false
    method: GET
```

- either goss should set its own user agent (like goss/0.4.4) or allow overriding the user agent on a per check basis

- none

catdevman:
##### Checklist

### Description of change
Close #882. Adding default user-agent header by detecting if a user-agent header was given in the `request_headers` field in the goss file and defaulting to `goss/x.x.x`. Any guidance that can be given on how to get the current  version of goss within the codebase would be appreciated. I saw app.Version but that seemed to be related to urfave/cli so wasn't sure I could get that from that far down. The easiest way is probably adding it to the context but I didn't want to bloat that since it would go to all checks not just http.

### DIFF
--- a/integration-tests/goss/goss-shared.yaml
+++ b/integration-tests/goss/goss-shared.yaml
@@ -204,9 +204,7 @@ http:
     request-headers:
       - "Foo: bar"
     headers: ["Content-Type: application/json"]
-    body:
-      - '"Foo": "bar"'
-      - '"user-agent": "goss/0.0.0"'
+    body: ['"Foo": "bar"', '"user-agent": "goss/0.0.0"']
   http://httpbin/headers?host:
     status: 200
     timeout: 60000
--- a/integration-tests/goss/goss-shared.yaml
+++ b/integration-tests/goss/goss-shared.yaml
@@ -205,6 +205,11 @@ http:
       - "Foo: bar"
     headers: ["Content-Type: application/json"]
     body: ['"Foo": "bar"']
+  http://httpbin/headers:
+    status: 200
+    timeout: 60000
+    headers: ["Content-Type: application/json"]
+    body: ['"User-Agent": "goss']
   http://httpbin/headers?host:
     status: 200
     timeout: 60000
--- a/integration-tests/goss/goss-shared.yaml
+++ b/integration-tests/goss/goss-shared.yaml
@@ -204,7 +204,7 @@ http:
     request-headers:
       - "Foo: bar"
     headers: ["Content-Type: application/json"]
-    body: ['"Foo": "bar"', '"user-agent": "goss/0.0.0"']
+    body: ['"Foo": "bar"']
   http://httpbin/headers?host:
     status: 200
     timeout: 60000
--- a/integration-tests/goss/goss-shared.yaml
+++ b/integration-tests/goss/goss-shared.yaml
@@ -204,7 +204,9 @@ http:
     request-headers:
       - "Foo: bar"
     headers: ["Content-Type: application/json"]
-    body: ['"Foo": "bar"']
+    body:
+      - '"Foo": "bar"'
+      - '"User-Agent": "goss/0.0.0"'
   http://httpbin/headers?host:
     status: 200
     timeout: 60000
--- a/integration-tests/goss/goss-shared.yaml
+++ b/integration-tests/goss/goss-shared.yaml
@@ -204,12 +204,9 @@ http:
     request-headers:
       - "Foo: bar"
     headers: ["Content-Type: application/json"]
-    body: ['"Foo": "bar"']
-  http://httpbin/headers:
-    status: 200
-    timeout: 60000
-    headers: ["Content-Type: application/json"]
-    body: ['"User-Agent": "goss']
+    body:
+      - '"Foo": "bar"'
+      - '"user-agent": "goss/0.0.0"'
   http://httpbin/headers?host:
     status: 200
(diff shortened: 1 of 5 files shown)


catdevman:
@aelsabbahy any suggestions on how to get the current version of goss down into the NewDefHTTP function?

aelsabbahy:
Apologies, somehow I've completely missed this.

I will review over the weekend and get back to you.

aelsabbahy:
To answer your question on version, it's currently here:

https://github.com/goss-org/goss/blob/ab2f627b385d3e4414490f9e13954004aff03651/cmd/goss/goss.go#L21

https://github.com/goss-org/goss/blob/62a9743aa04e950b205d742c6024c7ab727fd6f1/release-build.sh#L25

Perhaps moving it to a shared version of and having both main and this import the shared pkg/var. If there's a better way, feel free to suggest it.

catdevman:
@aelsabbahy I thought about this a little bit and since the util package is already included in the http code perhaps it might make sense to add it there.  I didn't change how main.version was getting set, instead I just added to that process. I tested by making a goss file that called an "echo server" that spat out whatever headers got sent.  I am not sure of the best way to test this... maybe integrating with a service like https://echo.free.beeceptor.com/ or spinning up an "echo server" to test against within the pipeline? or maybe this doesn't need tested? Let me know what you're thinking.

system/http.go
```diff
@@ -209,3 +217,12 @@ func (u *DefHTTP) Body() (io.Reader, error) {
 
 	return u.resp.Body, nil
 }
+
+func hasUserAgentHeader(headers []string) bool {
+	for _, header := range headers {
+		if strings.HasPrefix(header, USER_AGENT_HEADER_PREFIX) {
```
aelsabbahy, line 223: Do we need to lowercase the input to guarantee match?

catdevman: Yes we do need that :)

integration-tests/goss/goss-shared.yaml
```diff
@@ -205,6 +205,11 @@ http:
       - "Foo: bar"
     headers: ["Content-Type: application/json"]
     body: ['"Foo": "bar"']
+  http://httpbin/headers:
```
aelsabbahy, line 208: Yaml will merge this incorrectly with the one above it. since it's a map key the last one will win.

Perhaps just combine this one with the one above it, something like this:

```yaml
http:
  http://httpbin/headers:
    status: 200
    timeout: 60000
    request-headers:
      - "Foo: bar"
    headers: ["Content-Type: application/json"]
    body:
      - '"Foo": "bar"'
      - '"User-Agent": "goss/0.0.0"'
```

aelsabbahy (commented):
Looks good, minor comment. Also, travis-ci seems to be not working, I'll take a look at that.

integration-tests/goss/goss-shared.yaml
catdevman: I did do lower case user agent because that is how it is in the code

aelsabbahy:
CI failed due to linting issues:

```
invalid gofmt:
cmd/goss/goss.go
make: *** [Makefile:49: fmt] Error 1
```

https://app.travis-ci.com/github/goss-org/goss/jobs/621547789#L307

catdevman:
Yeah I shouldn't be surprised I tried editing it through GitHub because I wasn't at a computer. 🤦‍♂️

catdevman:
@aelsabbahy I know it was just because of whatever character was in there but it does crack me up a little bit that in that last run Windows was like "looks good to me brother" and mac and linux are saying "hold on a moment something is not quite right" :laughing:

catdevman:
Well I was trying to figure out why Linux was failing but perhaps that is expected behavior in the pipeline?

aelsabbahy:
No problem, I added the test back. The issue is httpbin returns the headers as pascal case but the goss.yaml was lowercase so it wasn't matching.

aelsabbahy (approved):
Thank you for taking the time and effort to add this. Great addition!

catdevman:
@aelsabbahy ah I believe that's a different error then I was getting locally or perhaps what I was getting just didn't make sense for it to be this error. 
Either way glad to see it merged and happy to have been of service. I've never gotten to use goss professionally but I always thought it was a brilliant bit of tech.

aelsabbahy:
Curious, do you use it for personal projects? If so, what's your primary use-case?

Always fun learning how others are using Goss.

Fun note, after initially releasing Goss, I didn't use it personally or professionally for 1-2 years.

catdevman:
I've not found a personal use. I worked in DevOps for a while and I was going to use it for health checks and maybe even regression test for a rest API but they hired a test engineer and he scrapped that for his own thing that did exactly what goss does but it took him 6 months to just get started in it. Then I left right as the whole software division closed up. That was the closest I got. But I have some new ideas and I think eventually these projects could benefit from goss. Mostly as a way of regression testing. Essentially boot up a golden DB and test if the responses have that is expected in them.

PR #893, opened 2024-04-18
s/during during/during/

alexandregz:

### Description of change

Just replace "during during" from documentation.


aelsabbahy:
Oops, great catch, thank you!

PR #895, opened 2024-04-21
regexFirstGroup return first group from regexp, useful with .env file…

alexandregz:
…s and another config files.

From my server I need to access config.inc.php file to know user, password and BD to execute a mysql command to check data using "command" resource. Also can use with .env files or similar.

Example:

{{ $regexBDRC := "\\'mysql:\\/\\/[a-z0-9]+:[a-z0-9]+@localhost\\/(roundcube[a-z0-9]+)\\';"}}
{{ $BDRC := readFile $fileRCConfig | regexFirstGroup $regexBDRC}}

Extract BD (first group from regexp) from Roundcube config, to check table "users" with something like that:

command:
  'mysql RC':
    exit-status: 0
    exec: {{ print "mysql -u " $UserBDRC " -p" $PassBDRC " " $BDRC " -e 'select user_id from users limit 1' --skip-column-names" }}
    stdout:
      and:
      - 1
    stderr: []
    timeout: 10000 # in milliseconds
    skip: false

##### Checklist

### Description of change

I just modified template to add func regexFirstGroup only. I checked on my own docker image with success.

### DIFF
--- a/docs/gossfile.md
+++ b/docs/gossfile.md
@@ -918,10 +918,29 @@ Available functions:
     `toUpper`
     :   Changes piped input to UPPERCASE
 
+       
     `findStringSubmatch regex string`
-    :   Returns the n group from regex. For example:
-        `{{ $findStr := findStringSubmatch "([a-z0-9+]):([a-z0-9+])@localhost\/([a-z0-9+])" "my $conf_bd = 'mysql://userBD01:passBD01@localhost/db01'" }}`
-        `{{ $USER := index $findStr 1 }}{{ $PASS := index $findStr 2 }}  {{ $BD := index $findStr 3 }}`
+    :   Returns map[string]interface{} with the names of the parenthesized subexpressions, like `(?P<first>[a-z])`
+        
+        {{ $regexDBrc := "\\'mysql:\\/\\/(?P<login>[a-z0-9]+):(?P<password>[a-z0-9]+)@localhost\\/(?P<database>roundcube_[a-z0-9]+)\\';"}}
+        
+        {{ $rcConf := readFile /home/user/roundcube/config.inc.php | findStringSubmatch $regexDBrc }}
+        {{ $UserDBrc := get $rcConf "login" }}
+        {{ $PassDBrc  := get $rcConf "password" }}
+        {{ $DBrc := get $rcConf "database" }}
+
+    If not exists named parenthesized subexps, returns stringfied array string:
+
+        {{ $regexDBrc := "\\'mysql:\\/\\/([a-z0-9]+):([a-z0-9]+)@localhost\\/(roundcube_[a-z0-9]+)\\';"}}
+        
+        {{ $rcConf := readFile /home/user/roundcube/config.inc.php | findStringSubmatch $regexDBrc }}
+        {{ $UserDBrc := get $rcConf "1" }}
+        {{ $PassDBrc  := get $rcConf "2" }}
+        {{ $DBrc := get $rcConf "3" }}
+
+    NOTE: stringfied string array begins with "1" ("0" is all the string matched)
+
+        
 
 
 !!! warning
--- a/docs/gossfile.md
+++ b/docs/gossfile.md
@@ -918,6 +918,15 @@ Available functions:
     `toUpper`
     :   Changes piped input to UPPERCASE
 
+    `regexFirstGroup "my $conf_bd = '(mysql://user:pass@localhost/db)'"`
+    :   Returns the first group from regexp, for example config to connect mysql. Useful with `readFile FILE | regexFirstGroup REGEXP`
+
+```go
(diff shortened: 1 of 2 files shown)


aelsabbahy:
With all the options above, I think care should be given to how we handle cases where the group isn't matched. Empty string, error, something else? 🤔

alexandregz:
> With all the options above, I think care should be given to how we handle cases where the group isn't matched. Empty string, error, something else? 🤔

Personally, I prefer error 👍 

If you want, I can put changes to commit, basically use your function "func regexFindSubmatch(regex, input string, index int) (string, error)" and documentation change to show regexFindSubmatch use.

ripienaar:
I also prefer `{{ findStringSubmatch "(\d+)-(\d+)-(\d+)" . | get 2 "Group not found" }}` which for the specific case of first can be `{{ findStringSubmatch "(\d+)-(\d+)-(\d+)" . | first }}` which would return empty string but if you needed an error would be  `{{ findStringSubmatch "(\d+)-(\d+)-(\d+)" . | mustFirst }}`

So I think we dont need to decide how missing data are handled in this function the downstream access into the resulting array has all the options we need

alexandregz:
> I also prefer `{{ findStringSubmatch "(\d+)-(\d+)-(\d+)" . | get 2 "Group not found" }}` which for the specific case of first can be `{{ findStringSubmatch "(\d+)-(\d+)-(\d+)" . | first }}` which would return empty string but if you needed an error would be `{{ findStringSubmatch "(\d+)-(\d+)-(\d+)" . | mustFirst }}`
> 
> So I think we dont need to decide how missing data are handled in this function the downstream access into the resulting array has all the options we need

Something like that?:

```go
func findStringSubmatch(pattern, input string) map[string]interface{} {
	re := regexp.MustCompile(pattern)
	els := re.FindStringSubmatch(input)

	elsMap := make(map[string]interface{})
	for i := 0; i < len(els); i++ {
		elsMap[strconv.Itoa(i)] = els[i]
	}
	return elsMap
}
```

because Sprig `get()` needs map[string]interface{} ( https://github.com/Masterminds/sprig/blob/master/dict.go#L8-L13 )

ripienaar:
Yeah indeed, `get` wont work - `index` probably will. I would not turn it into a map.

alexandregz:
Yep, `index` works but without pipe: string is the first param, not last, you can't use something like `{{ findStringSubmatch "(\d+)-(\d+)-(\d+)" $str | index 2 }}`.

For me, works with:

{{ $findStr := readFile "/home/user/roundcube/config.in.php" | findStringSubmatch "([a-z0-9+]):([a-z0-9+])@localhost\/([a-z0-9+])"}}
{{ $USER := index $findStr 1 }}

aelsabbahy:
So it seems like the only options are:

Function takes index:

1. 
```
func regexFindSubmatch(regex, input string, index int) (string, error)
```

2. Function returns map + `get`

3. Same as 2 but maybe also create keys for named captures, ex: https://stackoverflow.com/a/20751656

4. Array + index, but clunky user experience.


It seems to me option three allows for the most flexibility and will allow using named captures which can be helpful for making a regex self documenting. Feel free to push back if you disagree.

alexandregz:
Hi. I modified  `findStringSubmatch` to returns:
-  keys with named captures `(?P<first>,...)`: `elsMapsNamed: [map[string]interface {}{"user":"my_user"}]`
- or stringfied array string (``): `elsMap:[map[string]interface {}{"0":"all string with my_user", "1":"all string with", "2":"my_user"}]`

both to use with `get`


If don't exists named captures, function returns the stringfied array (because user don't introduce something with the (?P<name>...) syntax.

aelsabbahy:
This looks great! Any reason to not just have one map and always return it?

This way:

1. You can get the first group using either `get "1"` or `get "name"`
2. You can mix and match named groups and non named groups. Not saying this is a good idea, but not sure we should actively prevent it
3. Consistency.. adding one named groups doesn't completely change the map

Let me know if there's a danger or something that would break with the single map approach.

alexandregz:
Always return a `map[string]interface{}` to use with `get`: when code have named captures, returns this; if not, returns same kind of map, but with keys "1", "2", "3", etc.

But I think mix both (named captures and "standard" map with numbered keys) is complicated to manage and not useful.

alexandregz:
Because if returns only named captures, you force to use named captures and maybe sometimes you want don't use that and use a "standard" regexp, it's much more flexible

aelsabbahy:
Since the behavior is well documented, I'll merge this. If in the future it causes issues we can look into either splitting them into separate functions or combining them in a more merged way.

Sorry for the late response, thought I had addressed this PR, but apparently I didn't.

aelsabbahy:
Many thanks for this great addition, it'll be in the next Goss release! 🎉

PR #902, opened 2024-05-07
Add Debian Bookworm to integration-tests

dklimpel:
##### Checklist

### Description of change

The current used images / Dockerfiles are EoL. Starting add a supported Debian Bookworm image.

I did a few tests. Since Debian Jessy the [check for run level](https://github.com/goss-org/goss/blob/e73553f9c3065ac297499dafb4f8abef6acb24ad/docs/manual.md#service) does not work anymore.

```yaml
bookworm:
  services:
    apache2: []
```


dklimpel:
A few results for the ticket.

The new container image runs in podman without any problems. In docker the container terminates immediately.
I have the same result as the CI on my local machine.

The old cent os image does not run on my local machine as well. It behaves in the same way as in the github pipeline (my tests). The container freezes without starting the services it contains (webserver/proxy).

The images all contain more or less systemd/init. This is not really an architecture or a use case for containers. Starting multiple processes could also be done with the help of e.g. supervisord.

The real reason why init/systemd is included is probably to check services via goss. This would certainly not be possible in standard containers.

aelsabbahy:
Ok, so I think we need to be careful with upgrading different Linux versions. The purpose of the integration tests is to test against real world examples of services/package managers that goss supports.

https://github.com/goss-org/goss/blob/aed56336c3e8ff683e9540065b502f423dd6760d/system/system.go#L162-L179

So if upgrading a distro will move it from say sysV init to systemd then the integration tests now have a gap in coverage. Also, it probably means they are potentially redundant (e.g. Debian may no longer be providing any value over ubuntu tests).

dklimpel:
I understand the arguments. And that is also correct.

On the other hand "test against real world examples" is not possible with the current versions that are EOL.

| version | EOL |
| -------- | ---- |
| centos7 | 30 Jun 2024 |
| wheezy | 31 May 2018 |
| trusty | 02 Apr 2019 |

aelsabbahy:
sure, I wonder if that part of the codebase should be deprecated or if I need to see if any non-EOL linux distributions still use sysV init.

aelsabbahy:
> sure, I wonder if that part of the codebase should be deprecated or if I need to see if any non-EOL linux distributions still use sysV init.

Hmm.. doesn't seem like sysv init is around much anymore.

Let me think through the best approach for CI, it may need to be rethought a bit. Currently the docker tests allow for testing different package managers and init styles. Perhaps it's time to start deprecating support for some of them and removing some of the tests.

kgaughan:
#1061 does this out of necessity.

I think the integration tests need to be chopped up differently: there are a whole bunch of tests (most of the networking tests) that could be done on any distro and get the same results. I think it'd be best to limit the tests that actually need a distro to that distro. I'd even consider separating out DNS tests from HTTP tests so they can be run in parallel.

kgaughan:
I think this can be closed out now that #1063 is merged.

PR #907, opened 2024-05-10
bump alpine linux to 3.19

dklimpel:
##### Checklist

### Description of change

Bump Alpine Linux in integration test.

### DIFF
--- a/integration-tests/Dockerfile_alpine3
+++ b/integration-tests/Dockerfile_alpine3
@@ -1,9 +1,9 @@
-FROM alpine:3.12
+FROM alpine:3.19
 MAINTAINER Ahmed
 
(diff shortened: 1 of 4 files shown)


aelsabbahy:
There's some logic to publish the docker images with a checksum to speed up future CI builds.

I wonder if those would perform faster in ghcr when this is all running as GitHub actions. 🤔

aelsabbahy:
Agree with your assessment, 

1. I'll manually update the image + checksum for now. The logic just falls back to slower builds, so nothing is broken.
2. When pipelines are ready, they will be responsible for updating the images after successful CI. This way whenever we make image changes, they're slow at first, but speed up once CI passes in master branch.

PR #908, opened 2024-05-10
Add Ubuntu Noble to integration-tests

dklimpel:
##### Checklist

### Description of change

The current used images / Dockerfiles are EoL. Add a supported Ubuntu Noble image.

[Check for run level](https://github.com/goss-org/goss/blob/e73553f9c3065ac297499dafb4f8abef6acb24ad/docs/manual.md#service) does not work anymore.

```yaml
noble:
  services:
    apache2: []
```


aelsabbahy:
I think the new Ubuntu version doesn't use one of the supported service types for run levels:

https://github.com/goss-org/goss/blob/e73553f9c3065ac297499dafb4f8abef6acb24ad/docs/manual.md#service

> runlevels is only supported on Alpine init, sysv init, and upstart

kgaughan:
I would say that jammy is a better target for now. I've done something similar in #1061.

kgaughan:
You can close this out: I got it updated to jammy in #1063.

PR #901, opened 2024-05-07
Build own docker images for integration tests

dklimpel:
##### Checklist

- [x] `make test-all` (UNIX) passes. CI will also test this
- [ ] unit and/or integration tests are included (if applicable)
- [ ] documentation is changed or added (if applicable)

### Description of change

Add own github workflow to build the docker images from the test and store them in a repo (ghcr.io). This can resolve the dependency on external repos in the future and ensure more up-to-date images or tests.

There are no significant changes at the Dockerfiles. I have therefore updated the md5 files so that the tests continue to run unchanged.

- wheezy has expired repo signatures and needs a `--force-yes`
- `archlinux:base` has moved
- problems with `pacman`, but `systemd-sysvcompat` seems up to date

```
#6 [2/4] RUN pacman -Sy --noconfirm systemd-sysvcompat
#6 0.061 :: Synchronizing package databases...
#6 2.666  core downloading...
#6 2.666  extra downloading...
#6 2.671 resolving dependencies...
#6 2.838 :: There are 2 providers available for dbus-units:
#6 2.838 :: Repository core
#6 2.838    1) dbus-broker-units  2) dbus-daemon-units
#6 2.838 
#6 2.838 Enter a number (default=1): 
#6 2.838 looking for conflicting packages...
#6 2.839 
#6 2.839 Package (5)              Old Version  New Version  Net Change  Download Size
#6 2.839 
#6 2.839 core/dbus-broker                      36-2           0.35 MiB       0.15 MiB
#6 2.839 core/dbus-broker-units                36-2           0.00 MiB       0.00 MiB
#6 2.839 core/systemd             255.2-1      255.5-4       -0.47 MiB       8.00 MiB
#6 2.839 core/systemd-libs        255.2-1      255.5-4       -0.06 MiB       1.07 MiB
#6 2.839 core/systemd-sysvcompat  255.2-1      255.5-4        0.00 MiB       0.01 MiB
#6 2.839 
#6 2.839 Total Download Size:    9.22 MiB
#6 2.839 Total Installed Size:  33.84 MiB
#6 2.839 Net Upgrade Size:      -0.18 MiB
#6 2.839 
#6 2.839 :: Proceed with installation? [Y/n] 
#6 2.839 :: Retrieving packages...
#6 3.718  systemd-255.5-4-x86_64 downloading...
#6 3.718  systemd-libs-255.5-4-x86_64 downloading...
#6 3.718  dbus-broker-36-2-x86_64 downloading...
[...]

PR #909, opened 2024-05-10
Add pipeline for build goss docker image

dklimpel:
##### Checklist

### Description of change

Add building own docker image and upload to ghcr.io. Realted image: https://github.com/aelsabbahy/goss-docker/blob/master/Dockerfile

Build image from
- master branch with goss version: `<latest-tag>-master+<short commit sha>` (semantic version)
- tag with version `<tag>`

Add CVE scan to github  security page

Related to:
- #901

### DIFF
--- a/.github/workflows/docker-goss.yaml
+++ b/.github/workflows/docker-goss.yaml
@@ -1,0 +1,93 @@
+name: Docker image for Goss
+
+on:
+  push:
+    branches:
+      - master
+    tags:
+      - "v*"
+  workflow_dispatch:
+
+env:
+  PLATFORMS: "linux/amd64,linux/arm64"
+
+jobs:
+  goss:
+    name: Build and push Docker image
+    runs-on: ubuntu-latest
+    permissions:
+      packages: write
+      contents: read
+      security-events: write # To upload Trivy sarif files
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Login to GHCR
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Extract metadata (tags, labels) for Docker
+        id: meta
+        uses: docker/metadata-action@v5
+        with:
+          images: |
+            ghcr.io/${{ github.repository_owner }}/goss
+
+      - name: Get latest git tag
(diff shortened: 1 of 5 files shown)


aelsabbahy:
So is this essentially a replacement for: https://github.com/aelsabbahy/goss-docker

Btw.. many thanks for the awesome work you've submitted! These are all great improvements.

.github/workflows/docker-goss.yaml
```diff
@@ -0,0 +1,98 @@
...
+            GOSS_VERSION=${{ github.ref_name }}
+          context: .
+          push: true
+          tags: |
+            ghcr.io/${{ github.repository_owner }}/goss:latest
+            ghcr.io/${{ github.repository_owner }}/goss:${{ github.ref_name }}
+          labels: ${{ steps.meta.outputs.labels }}
+          platforms: ${{ env.PLATFORMS }}
+
+      - name: Run Trivy vulnerability scanner
```
aelsabbahy, line 88: Is this some thing that typically runs before or after the container is pushed to the repository?

Also, when doing a goss release, this will run against the master branch, but not the released version?

aelsabbahy (commented):
Added a small question, and might be helpful to add some documentation to this PR (can be minimal). Otherwise, it seems ready to merge

dklimpel:
> Added a small question, and might be helpful to add some documentation to this PR (can be minimal). Otherwise, it seems ready to merge

Did you mean this kind of documentation? Or something to the GH pipeline?

.github/workflows/docker-goss.yaml
dklimpel: This is a small add on and runs whenever an image (goss:master) is pushed into the repo. It also scans the image goss:master. This is the smallest step at the time of creating a tag. The step should not trigger a failing, but serves to keep an overview of the image. It scans not only the goss binary, but the whole image including the operating system (base image).
This gives you an overview of the status of the CVEs at the time the image was published. In my opinion, this cannot replace a regular scan of your own code.

aelsabbahy:
> Did you mean this kind of documentation? Or something to the GH pipeline?

Yup, this looks great!

dklimpel:
I will have a look why it is failing. I had tested it successfully.

dklimpel:
In my repo I have not problems and it works. But I cannot see the package in goss Repo: https://github.com/orgs/goss-org/packages?repo_name=goss

Perhaps GH needs some time to activate.

dklimpel:
It seems that permissions have to set there for the package: https://docs.github.com/en/packages/learn-github-packages/configuring-a-packages-access-control-and-visibility

aelsabbahy:
Yeah it's set to private, but I can't set it to public due to org settings. I'll take a look at it sometime in the next couple of days.

aelsabbahy:
https://github.com/goss-org/goss/pkgs/container/goss

Done, was pretty simple actually.

PR #899, opened 2024-05-06
Add github golang lint CI pipeline

dklimpel:
##### Checklist

- [x] `make test-all` (UNIX) passes. CI will also test this
- [ ] unit and/or integration tests are included (if applicable)
- [ ] documentation is changed or added (if applicable)

### Description of change

Related to:
- #786 
- #649 

It is only a start for using gh pipelines. Add a linting pipeline for go with github actions.

Linting is currently broken in Travis CI:

```
INFO: Starting build lint
golint . ./cmd/... ./matchers/... ./outputs/... ./resource/... ./system/... ./util/... || true
/bin/sh: golint: command not found
```

Do not lint for `latest-windows`. There is a failure if nothing was changed:

```
only new issues on pull_request: C:\Users\RUNNER~1\AppData\Local\Temp\tmp-6348-vzusUQlR5XIf\pull.patch
  Running [D:\a\_temp\b2f1ab8f-0666-4c9f-8876-3984c905d141\golangci-lint-1.58.0-windows-amd64\golangci-lint run --out-format=github-actions --new-from-patch=C:\Users\RUNNER~1\AppData\Local\Temp\tmp-6348-vzusUQlR5XIf\pull.patch --new=false --new-from-rev=] in [D:\a\goss\goss] ...
  level=error msg="Running error: context loading failed: no go files to analyze: running `go mod tidy` may solve the problem"
  level=error msg="Timeout exceeded: try increasing it by passing --timeout option"
```

Only lint code changes (`only-new-issues: true`) and not all code.
Reason is that there are some failures at the code:

```
Running [/Users/runner/golangci-lint-1.58.0-darwin-arm64/golangci-lint run --out-format=github-actions] in [/Users/runner/work/goss/goss] ...
  Error: error	util/config.go:99:15:	Error return value is not checked (errcheck)
  Error: error	system/service_init.go:75:9:	Error return value of `cmd.Run` is not checked (errcheck)
  Error: error	system/service_systemd.go:38:9:	Error return value of `cmd.Run` is not checked (errcheck)
  Error: error	system/service_systemd.go:57:9:	Error return value of `cmd.Run` is not checked (errcheck)
  Error: error	outputs/json.go:102:16:	Error return value of `json.Unmarshal` is not checked (errcheck)
  Error: error	outputs/junit.go:99:16:	Error return value of `xml.EscapeText` is not checked (errcheck)
  Error: error	serve.go:94:19:	Error return value of `resp.body.WriteTo` is not checked (errcheck)
[...]

PR #922, opened 2024-05-22
fix some linting failures and configure golangci-lint

dklimpel:
##### Checklist

### Description of change

Fix some linting failures and add new go linting to makefile.

I am not a go expert. Please have a deeper look inside.

See a list of lint failures at:
- #899


system/service_systemd.go
```diff
@@ -73,7 +77,9 @@ func (s *ServiceSystemd) Running() (bool, error) {
 		return false, nil
 	}
 	cmd := util.NewCommand("systemctl", "-q", "is-active", s.service)
-	cmd.Run()
+	if err := cmd.Run(); err != nil {
+		return false, nil
```
aelsabbahy, line 81: these should probably bubble up the error:

```
return false, err
```

Makefile
```diff
@@ -35,6 +35,11 @@ htmlcov:
 	go test -v -coverpkg=./... -coverprofile=c.out ./...
 	go tool cover -html ./c.out
 
+.PHONY: golangci-lint
+golangci-lint:
+	$(info INFO: Starting build $@)
+	go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.58.2
+	golangci-lint run --timeout 5m $(pkgs)
 
 lint:
```
aelsabbahy, line 44: Maybe we change this to also run golangci-lint.

```
lint: golangci-lint
```

aelsabbahy (commented):
Thanks for adding this, l wonder if CI should run `make lint` to be consistent with local dev story.

For example, this PR is currently passing the golangci-lint, but failing Travis do to the gofmt check. Ideally, GitHub actions would catch both.

Makefile
dklimpel: I think `golint` does not exist anymore and `golint` can be replaced by `golangci-lint`.

dklimpel:
> For example, this PR is currently passing the golangci-lint, but failing Travis do to the gofmt check. Ideally, GitHub actions would catch both.

I have added fmt to pipeline.
Github currently does not find it because it only looks at modified code, and this line is currently not included in the diff. If you check everything, the pipeline fails always until all errors are fixed. A bit of a dilemma.

https://github.com/goss-org/goss/blob/7ca545977ec1c9c9c3de6e612b92940940be849d/.github/workflows/golangci-lint.yaml#L27

dklimpel:
I do not not knwo why it is failing.

image

Perhaps a Travis CI problem :(

Makefile
aelsabbahy: Hah, didn't even notice that, yeah.. we can replace `golint` in the `lint` target with `golangci-lint` that makes more sense.

So much has changed in the go ecosystem since Goss was originally written.

aelsabbahy (changes requested):
Looks great, small comment on golint and make targets.

The travis-ci is failing because cmd.Run() will return an error on non-zero exit status. So would need to either:
1. Ignore the errors and skip linting those calls
2. Add a check for ExitError and ignore it
3. Refactor cmd.Run() to ignore ExitError and the consumer code has to do a check for Status if they care.

aelsabbahy (commented):
I think we can promote this out of draft.

At some point, I'll run a `golangci-lint run` and fix all the findings.

curious, what are the pros/cons of using golangci-lint-action vs just running `make lint` in a github action?

I assume the former provides some unique benefits/flags, but requires making sure the local `make lint` and the job are in sync, (same version, same flags, etc.).

Anyways, correct me if I'm wrong, but putting everything behind make should make it easier to move off of github if that ends up being another transition in the future?

resource/addr.go
```diff
@@ -51,7 +53,7 @@ func (a *Addr) GetAddress() string {
 }
 
 func (a *Addr) Validate(sys *system.System) []TestResult {
-	ctx := context.WithValue(context.Background(), "id", a.ID())
+	ctx := context.WithValue(context.Background(), idKey{}, a.ID())
```
dklimpel, line 56: The issue was:

```
resource/addr.go:54:49: SA1029: should not use built-in type string as key for value; define your own type to avoid collisions (staticcheck)
	ctx := context.WithValue(context.Background(), "id", a.ID())
```

dklimpel:
> Looks great, small comment on golint and make targets.
> 
> The travis-ci is failing because cmd.Run() will return an error on non-zero exit status. So would need to either:
> 
> 1. Ignore the errors and skip linting those calls
> 2. Add a check for ExitError and ignore it
> 3. Refactor cmd.Run() to ignore ExitError and the consumer code has to do a check for Status if they care.

I have decided to "Ignore the errors and skip linting those calls"

dklimpel:
> curious, what are the pros/cons of using golangci-lint-action vs just running `make lint` in a github action?
> 
> I assume the former provides some unique benefits/flags, but requires making sure the local `make lint` and the job are in sync, (same version, same flags, etc.).
> 
> Anyways, correct me if I'm wrong, but putting everything behind make should make it easier to move off of github if that ends up being another transition in the future?

I think the github actions are actively developed further and certainly optimised for performance and caching. For this you have to maintain makefile and action twice. The github action also runs without problems on forked branches. With the makefile, each developer has to make sure that it works, which hopefully it does out of the box.

> Anyways, correct me if I'm wrong, but putting everything behind make should make it easier to move off of github if that ends up being another transition in the future?

I think so, too.
The dual maintenance is currently limited to the golint version

dklimpel:
Sorry, the PR has become a little more complicated than planned. In the end, the currently configured linting has now been fully implemented and the pipeline no longer only needs to check the diff (`only-new-issues: true`)

aelsabbahy (approved):
This looks good to me. I assume this is ready to merge.

I can enable some of the commented out lintes and fix some of the issues as follow up PRs.

marked this as approved, if you're ready for this to be merged just post a comment here and I'll merge it.

aelsabbahy:
Minor question:

>The github action also runs without problems on forked branches. With the makefile, each developer has to make sure that it works, which hopefully it does out of the box.

Why does `make lint` not work on forked branches in github?

dklimpel:
> marked this as approved, if you're ready for this to be merged just post a comment here and I'll merge it.

Yes, is ready for merging.

dklimpel:
> Why does make lint not work on forked branches in github?

I think I was thinking too much. The crucial point is that the Travis CI only runs when you create a PR. Otherwise you would have to have your own Travis account. But for development it would also be good to run the CI beforehand creating a PR. For example, I sometimes develop directly in GitHub without a computer that has `make` and co installed. It's good to get direct feedback.

PR #927, opened 2024-06-08
disable deprecated codeclimate plugin golint

dklimpel:
##### Checklist

### Description of change

- The plugin is deprecated: https://docs.codeclimate.com/docs/golint
- And it reports errors: https://codeclimate.com/github/aelsabbahy/goss/builds/693

I have not seen an alternative for this plugin.

Reported error/line:
https://github.com/goss-org/goss/blob/6a8fc9623d846e4b9093fde7f9685e6abcc21457/goss_config.go#L114

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -6,8 +6,6 @@ checks:
 plugins:
   gofmt:
     enabled: true
-  golint:
-    enabled: true
   govet:
     enabled: true
 


aelsabbahy:
Same question as other PR, does code climate offer any additional value at this point?

I feel it was added many years back and then ignored 😅

dklimpel:
> I feel it was added many years back and then ignored

If you do not used this, it can be removed in feature.

PR #936, opened 2024-06-19
Revert alpine

aelsabbahy:
##### Checklist

### Description of change

Alpine integration test is failing consistently as of late. Reverting to old alpine version to see if that resolves the issue.

### DIFF
--- a/integration-tests/Dockerfile_alpine3
+++ b/integration-tests/Dockerfile_alpine3
@@ -1,9 +1,9 @@
-FROM alpine:3.19
+FROM alpine:3.12
 MAINTAINER Ahmed
 
 # install apache2 and remove un-needed services
 RUN apk update && \
-  apk add --no-cache openrc apache2=2.4.59-r0 bash ca-certificates tinyproxy && \
+  apk add openrc apache2 bash ca-certificates tinyproxy && \
   rc-update add apache2 && \
   rc-update add tinyproxy && \
   rm -rf /etc/init.d/networking /etc/init.d/hwdrivers /var/cache/apk/* /tmp/*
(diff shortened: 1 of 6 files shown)


aelsabbahy:
Seems re-using :latest is a bad idea since travis-ci seems to cache it now.

Next up will be modifying this PR to use pinned docker tags. image_name:$(sum of dockerfile)

Will also investigate what's causing alpine to behave differently. I wonder if it's as simple as tcp6 :thinking:

dklimpel:
```diff
-opts=(--env OS=$os --cap-add SYS_ADMIN -v "$PWD/goss:/goss" -d --name "$container_name" --security-opt seccomp:unconfined --security-opt label:disable)
+opts=(--env OS=$os --cap-add SYS_ADMIN -v "$PWD/goss:/goss" -d --name "$container_name" --security-opt seccomp:unconfined --security-opt label:disable --privileged)
```

I did a lot of tests. Mostly I had localy trouble with this line. Perhaps Travis CI has changed this rights management.

For example: https://github.com/goss-org/goss/pull/908/files#diff-42ccbd89e05a79ca47745ab8cf461626edf35a0cd3429aae34b070a832352b94R39

dklimpel:
I can reproduce this with GH action. Also with alpine:3.12.

Apache2 does not listen on IPv4. Either port 80 nor 8080. I have no problems on my local machine.

I tried it on port 8080:

```
Proto Recv-Q Send-Q Local Address           Foreign Address         State       PID/Program name    
tcp        0      0 127.0.0.11:37453        0.0.0.0:*               LISTEN      -
tcp        0      0 0.0.0.0:8888            0.0.0.0:*               LISTEN      -
tcp        0      0 :::8888                 :::*                    LISTEN      -
tcp        0      0 :::8080                 :::*                    LISTEN      168/httpd
```

aelsabbahy:
fixed by #937

PR #935, opened 2024-06-19
Try CI with cached images

aelsabbahy:
##### Checklist

- [ ] `make test-all` (UNIX) passes. CI will also test this
- [ ] unit and/or integration tests are included (if applicable)
- [ ] documentation is changed or added (if applicable)

### Description of change

Seeing intermittent (increasingly more persistent) issues with apache on new alpine linux, updates images on dockerhub and md5 cache files to see if leveraging cached copies helps.

PR #937, opened 2024-06-21
fix integration tests for alpine3

dklimpel:
##### Checklist

### Description of change

Apache2 has only one listener by design. Either IPv4 or IPv6. 

It seems that the order of the ip addresses has been changed (in CI/docker runtime). Previously IPv4 first to now IPv6 first. If the test with Apache2 should explicitly test IPv4, the service should also explicitly run on IPv4.

Related:
- #935 
- #936 
- #907

### DIFF
--- a/integration-tests/Dockerfile_alpine3
+++ b/integration-tests/Dockerfile_alpine3
@@ -4,6 +4,7 @@ MAINTAINER Ahmed
 # install apache2 and remove un-needed services
 RUN apk update && \
   apk add --no-cache openrc apache2=2.4.59-r0 bash ca-certificates tinyproxy && \
+  sed -i 's/Listen 80/Listen 0.0.0.0:80/g' /etc/apache2/httpd.conf && \
   rc-update add apache2 && \
   rc-update add tinyproxy && \
   rm -rf /etc/init.d/networking /etc/init.d/hwdrivers /var/cache/apk/* /tmp/*


aelsabbahy:
Yeah, I think this is related to https://github.com/goss-org/goss/issues/149

My guess is travis-ci changed that setting in their CI runners. I'm not able to reproduce that behavior on my local machine and the CI used to work just fine.. so that's my theory.

I added some debugging to my branch yesterday which showed that it was indeed listening to tcp6 here:

https://app.travis-ci.com/github/goss-org/goss/jobs/623192771#L5295

```
Active Internet connections (only servers)
Proto Recv-Q Send-Q Local Address           Foreign Address         State       PID/Program name    
tcp        0      0 127.0.0.11:42279        0.0.0.0:*               LISTEN      -
tcp        0      0 0.0.0.0:8888            0.0.0.0:*               LISTEN      -
tcp        0      0 :::8888                 :::*                    LISTEN      -
tcp        0      0 :::80                   :::*                    LISTEN      164/httpd
```

This PR is the cleanest fix, so I'll merge this and close out the other ones.

Great find and thanks for the elegant fix!

dklimpel:
I had a discussion with my colleague. One alternative is to switch from Apache2 to Nginx. Nginx runs on both ports, IPv4 and IPv6.

PR #938, opened 2024-06-21
Pushed new integration test images and update image md5

aelsabbahy:
##### Checklist

### Description of change

This might fail for a day or so until the travis image cache expires.

TODO: proper image tagging instead of relying on `:latest`


dklimpel:
With #901 you can create images by pushing to master or by start job manually.

It creates images with `latest` or `<md5_checksum_from_imagefile>` tag. It is possible to add images when a tag in repo is set.

Then it would be possible, for example, to create a `latest` image (or `<branch_name>` or `<md5_checksum_from_imagefile>` in the pipeline if the checksum is not correct and otherwise use the image from a defined or checksum tag.

aelsabbahy:
Oops, clicked merge before reading your message. I'll check out that PR.

PR #954, opened 2024-07-18
fix(ci): rework get latest tag for build Docker image

dklimpel:
##### Checklist

### Description of change

I have seen that building the Docker image had failed: https://github.com/goss-org/goss/actions/runs/9982852891/job/27589357250

Reason is a not fixed upstream bug. https://github.com/actions-ecosystem/action-get-latest-tag/pull/22
I was not able to reproduce this in my fork.

I have reworked this part of CI:
- only run to get the tag if required
- remove the dependency. this does not seem to be well maintained
- apply the fix from the PR

### DIFF
--- a/.github/workflows/docker-goss.yaml
+++ b/.github/workflows/docker-goss.yaml
@@ -45,10 +45,20 @@ jobs:
             ghcr.io/${{ github.repository_owner }}/goss
 
       - name: Get latest git tag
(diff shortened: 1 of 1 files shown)


aelsabbahy:
I'll merge this now. I can either cut a release to test.. or if we're close to switching over to GHA, I can cut a release using GHA and test all of it at once

aelsabbahy (approved):

PR #955, opened 2024-07-18
Fixed: the user-agent test on release builds

aelsabbahy:
The user-agent test assumed goss will always have v0.0.0, this is not the case in a release build.

##### Checklist

### Description of change

Fixes release builds.


dklimpel:
I am confused. I had expected that this failure also had happened with release v0.4.7. Presumably it was introduced with #892

Issue #915, opened 2024-05-17
Can we add ansible galaxy goss install to README?


marcinpraczko:
I would like to ask for adding ansible galaxy goss installation which I am maintaining. Long time ago I was working with `goss` on regular basic. Today I am using this on my local machine for different tasks. I've created Ansible Galaxy role for installation. 
If all maintainers agrees - this can be added to `goss` README file `Community Contributions` section.

Links for Repos and Ansible Galaxy:

- https://github.com/marcinpraczko/ansible-goss-install
- https://galaxy.ansible.com/ui/standalone/roles/marcinpraczko/goss-install/install/

You can also check Issues (closed) as there are my old suggestions and feature requests (few of them has been implemented in current version of goss).

Thank you in advance for reply for this question.

aelsabbahy:
@marcinpraczko Thanks for submitting this, looks great! Curious how this differs from https://github.com/naftulikay/ansible-role-degoss

@naftulikay I've used degoss in the past, great project! Wondering if it's still maintained?

marcinpraczko:
Hi @aelsabbahy.

I've done some quick check of this repository [https://github.com/naftulikay/ansible-role-degoss](https://github.com/naftulikay/ansible-role-degoss), and those are difference comparing with mine:
1. Above repository is installing always the latest version of `goss`,
2. Above repository is having also ansible module (not sure whether is still working) - to run `goss` and present result
3. Which means that above repo can install `goss` and run `goss tests`. 

My Ansible role is only installing `goss` (so far for amd64), however my intention is to add more options and support development for now. 

I am not asking replace above repository, but kindly asking add to `goss` readme as possible option of installing `goss` as user or as root (my ansible role allows to user choose whether this should be installed as root or not). 

Hope that I answered to your questions?

aelsabbahy:
That answers my question clearly, thank you.

Yes, feel free to submit a PR adding your repo to community contributions.

marcinpraczko:
Hi @aelsabbahy 

Thank you for confirmation - new PR created as requested.

marcinpraczko:
Hi @aelsabbahy  - I've done PR - however from some reasons CI is failing - my change was only with `README.md` - so not sure how this is failing. Is there anything which we can check what needs to be done to allow this PR pass CI?

aelsabbahy:
Closed by #995

PR #995, opened 2024-11-02
Add alternative version of installing goss from ansible

marcinpraczko:
Added alternative version of installing `goss` from `ansible role`.

Updated README.md.

- Related with: #915


aelsabbahy (approved):
Thanks for the addition.  Goss CI is going through a bit of a transition currently. Travis will be deprecated really soon.

PR #997, opened 2024-11-05
fix(dgoss): edgecase where no log is written

dklimpel:
##### Checklist

### Description of change

I think this is a bug or a edge case. I used `podman` and always got an empty log file.
Partly it is discribed: https://docs.podman.io/en/stable/markdown/podman-logs.1.html#follow-f

> **--follow, -f**
> Follow log output. Default is false.
>
> Note: When following a container which is removed by podman container rm or removed on exit (podman run --rm ...), there is a chance that the log file is removed before podman logs reads the final content.

I'm not sure if it's related to this or a problem with the flushing of the pipe in the operating system.

If the container is stopped before removal and a second pipe is created, the log is written. The content of the second pipe does not arrive and is therefore presumably swallowed.

If I am set timeout to `0` (`$CONTAINER_RUNTIME stop -t 0 "$id" > /dev/null`) then the last log entry appears in the log file in the tmp directory, but is apparently written to late and not copied to the target file.

### DIFF
--- a/extras/dgoss/dgoss
+++ b/extras/dgoss/dgoss
@@ -20,6 +20,8 @@ cleanup() {
     set +e
     { kill "$log_pid" && wait "$log_pid"; } 2> /dev/null
     if [ -n "$CONTAINER_LOG_OUTPUT" ]; then
+        info "Stopping container and copy log"
+        $CONTAINER_RUNTIME stop "$id" > /dev/null
         cp "$tmp_dir/docker_output.log" "$CONTAINER_LOG_OUTPUT"
     fi
     rm -rf "$tmp_dir"
@@ -66,6 +68,8 @@ run(){
     esac
 
     $CONTAINER_RUNTIME logs -f "$id" > "$tmp_dir/docker_output.log" 2>&1 &
+    # There is a chance that the log will not be written, an empty line is added for mitigation.
+    echo "" >> "$tmp_dir/docker_output.log" 2>&1 &
     log_pid=$!
     info "Container ID: ${id:0:8}"
 }
--- a/extras/dgoss/dgoss
+++ b/extras/dgoss/dgoss
@@ -18,14 +18,20 @@ error() {
 
 cleanup() {
     set +e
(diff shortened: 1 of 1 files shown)


extras/dgoss/dgoss
```diff
@@ -66,6 +68,8 @@ run(){
     esac
 
     $CONTAINER_RUNTIME logs -f "$id" > "$tmp_dir/docker_output.log" 2>&1 &
+    # There is a chance that the log will not be written, an empty line is added for mitigation.
+    echo "" >> "$tmp_dir/docker_output.log" 2>&1 &
```
aelsabbahy, line 72: Perhaps `touch` makes more sense here?

extras/dgoss/dgoss
```diff
@@ -20,6 +20,8 @@ cleanup() {
     set +e
     { kill "$log_pid" && wait "$log_pid"; } 2> /dev/null
     if [ -n "$CONTAINER_LOG_OUTPUT" ]; then
+        info "Stopping container and copy log"
+        $CONTAINER_RUNTIME stop -t 1 "$id" > /dev/null
```
aelsabbahy, line 24: Does the -t flag work consistently with podman or docker?

aelsabbahy (commented):
Small comments, otherwise looks good.

Interesting find!

extras/dgoss/dgoss
dklimpel: Yes.

- Podman: https://docs.podman.io/en/stable/markdown/podman-stop.1.html#time-t-seconds
- Docker: https://docs.docker.com/reference/cli/docker/container/stop/#time

extras/dgoss/dgoss
dklimpel: I will test and check this.

dklimpel: I have looked at it. `touch` does not help. I think I have found an better and more transparent solution.

aelsabbahy (approved):
Nice, this solution is elegant and clear.

PR #1013, opened 2025-01-21
feat: Add support for specifying URL Query params in HTTP resource

riton:
##### Checklist

- [X] `make test-all` (UNIX) passes. CI will also test this
- [X] documentation is changed or added (if applicable)

### Description of change

This P.R adds support for specifying _URL query params_ in the `http` resource.

This will properly handle _URL encoding_ as opposed to specifying _raw query params_ directly in the `url`.

----
📚 Documentation preview 📚: https://goss--1013.org.readthedocs.build/en/1013/

### DIFF
--- a/system/http.go
+++ b/system/http.go
@@ -30,22 +30,23 @@ type HTTP interface {
 }
 
 type DefHTTP struct {
-	http              string
-	allowInsecure     bool
-	noFollowRedirects bool
-	resp              *http.Response
-	RequestHeader     http.Header
-	RequestBody       string
-	Timeout           int
-	loaded            bool
-	err               error
-	Username          string
-	Password          string
-	CAFile            string
-	CertFile          string
-	KeyFile           string
-	Method            string
-	Proxy             string
+	http               string
+	allowInsecure      bool
+	noFollowRedirects  bool
+	resp               *http.Response
+	RequestHeader      http.Header
+	RequestBody        string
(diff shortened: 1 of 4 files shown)


system/http.go
```diff
@@ -30,22 +30,23 @@ type HTTP interface {
...
-	KeyFile           string
-	Method            string
-	Proxy             string
+	http               string
+	allowInsecure      bool
+	noFollowRedirects  bool
+	resp               *http.Response
+	RequestHeader      http.Header
+	RequestBody        string
+	RequestQueryParams map[string]string
```
aelsabbahy: Should this be: `map[string][]string` to support duplicate query params?

example: `?foo=1&foo=2`

aelsabbahy (commented):
This is a nice addition, I added one small comment regarding support for duplicate keys.

system/http.go
riton: Good idea, I'll make the modification

riton: This should have been addressed by https://github.com/ccin2p3/goss/commit/6d1142d0eb191ab79693c37c4882ae11a59a466a

kgaughan:
Could you rebase this against master or merge from master? I've fixed up the pipelines.

kgaughan:
Asking again, and it would also be helpful to have some integration tests and/or unit tests to make sure everything's copacetic.

PR #1017, opened 2025-01-28
Implement Service Checks For Windows

cthiel42:
##### Checklist

### Description of change

I noticed some of the functionality for Windows wasn't implemented when I was using Goss with Windows a few months ago and I'd like to implement some of it, starting with services.

I tried to keep the initial changes here simple to see what you were looking for. For instance, services and processes are set up a little different than the others, so there would need to be some small refactoring if you wanted the windows service file to only be compiled into the windows build. I have an idea about how it could be accomplished somewhat cleanly, but wanted to get your take on whether it was worth it or not first. Same goes for integration tests. There was one test already in there, let me know if you wanted more and what you want them to cover.


system/service_win.go
```diff
@@ -0,0 +1,54 @@
...
+		service: service,
+	}
+}
+
+func (s *ServiceWindows) Service() string {
+	return s.service
+}
+
+func (s *ServiceWindows) Exists() (bool, error) {
+	cmd := util.NewCommand("powershell", "-command", fmt.Sprintf("Get-Service -Name %s", s.service))
```
petemounce:
```suggestion
	cmd := util.NewCommand("powershell", "-NoProfile", "-command", fmt.Sprintf("Get-Service -Name %s", s.service))
```

... so that powershell doesn't spend time loading whose-ever's profile this is, nor is subject to whatever might be inside it.

(it would be good to refactor out a powershell helper method such that all powershell invocations had that flag)

petemounce: I forget; is the match greedy? For example, if there are two services `foo1` and `foo2`, will `Get-Service -Name foo` return both?

system/service_win.go
```diff
@@ -0,0 +1,54 @@
...
+	if strings.Contains(cmd.Stderr.String(), "Cannot find any service with service name") {
+		return false, nil
+	}
+	return true, cmd.Err
+}
+
+func (s *ServiceWindows) Enabled() (bool, error) {
+	cmd := util.NewCommand("powershell", "-command", fmt.Sprintf("$(Get-Service -Name %s).StartType", s.service), "-eq", "\"Automatic\"")
+	cmd.Run()
+	if strings.Contains(cmd.Stdout.String(), "True") {
```
petemounce: (suggestion only, to make the shell command easier to read by moving the conditional into go)

```suggestion
	cmd := util.NewCommand("powershell", "-command", fmt.Sprintf("$(Get-Service -Name %s).StartType", s.service))
	cmd.Run()
	if strings.Contains(cmd.Stdout.String(), "Automatic") {
```

system/service_win.go
```diff
@@ -0,0 +1,54 @@
...
+		return true, cmd.Err
+	}
+	return false, cmd.Err
+}
+
+func (s *ServiceWindows) Running() (bool, error) {
+	cmd := util.NewCommand("powershell", "-command", fmt.Sprintf("$(Get-Service -Name %s).Status", s.service), "-eq", "\"Running\"")
+	cmd.Run()
+	if strings.Contains(cmd.Stdout.String(), "True") {
```
petemounce:
```suggestion
	cmd := util.NewCommand("powershell", "-command", fmt.Sprintf("$(Get-Service -Name %s).Status", s.service))
	cmd.Run()
	if strings.Contains(cmd.Stdout.String(), "Running") {
```

petemounce (commented):
@cthiel42 thanks very much for your contribution, and apologies for latency on review. This looks broadly good to me. I think ideally it'd be conditionally compiled so it's only present within the Windows binary to follow the pattern the other OS-specific support established.

re: my suggestions - please consider them my own in-memory suggestions for changes as opposed to relying on them to be syntactically / semantically correct. I made them inline during the review, not via an editor.

@aelsabbahy any comments?

cthiel42:
@petemounce Thanks for the review. I'll take a look over the next few days and push some changes for your comments.

system/service_win.go
cthiel42: I changed and tested this in the latest commit. I like this approach better.

system/service_win.go
cthiel42: It has to be an exact match otherwise it'll return not found.


pwsh_example

cthiel42: I made a helper function for calling powershell and including this flag as a default

cthiel42:
@petemounce I pushed up some changes. I switched the powershell calls to be through a helper function. I removed the comparison operator from the powershell commands and put them in the code (thanks for that suggestion, it looks a lot cleaner to me).  Lastly, the windows service file is now not included in linux builds.

On that last item, I accomplished that by stubbing out the call to NewServiceWindows. That's a functioning way for the linux builds to run with that call in there. Another way would be to take `system.go` and split it into `system_linux.go` and `system_windows.go`. Let me know your thoughts on that.

system/service_windows.go
```diff
@@ -0,0 +1,57 @@
...
+	cmd := util.NewCommandForWindowsPowershell("Get-Service", "-Name", s.service)
+	cmd.Run()
+	if strings.Contains(cmd.Stderr.String(), "Cannot find any service with service name") {
+		return false, nil
+	}
+	return true, cmd.Err
+}
+
+func (s *ServiceWindows) Enabled() (bool, error) {
+	cmd := util.NewCommandForWindowsPowershell(fmt.Sprintf("$(Get-Service -Name %s).StartType", s.service))
```
petemounce:
```suggestion
	cmd := util.NewCommandForWindowsPowershell(fmt.Sprintf("$(Get-Service -Name %q).StartType", s.service))
```

... to quote the input, making resilient to service names with spaces breaking the invocation and also reduce injection risk (I think...?).

(Again, no local way to try verify that my suggestion is legit; apologies)

system/service_windows.go
```diff
@@ -0,0 +1,57 @@
...
+	cmd.Run()
+	if strings.Contains(cmd.Stdout.String(), "Automatic") {
+		return true, cmd.Err
+	}
+	return false, cmd.Err
+}
+
+func (s *ServiceWindows) Running() (bool, error) {
+	cmd := util.NewCommandForWindowsPowershell(fmt.Sprintf("$(Get-Service -Name %s).Status", s.service))
```
petemounce:
```suggestion
	cmd := util.NewCommandForWindowsPowershell(fmt.Sprintf("$(Get-Service -Name %q).Status", s.service))
```

util/command_windows.go
```diff
@@ -28,3 +27,25 @@ func NewCommandForWindowsCmd(name string, arg ...string) *Command {
 
 	return command
 }
+
+func NewCommandForWindowsPowershell(name string, arg ...string) *Command {
+	//fmt.Println(arg)
```
petemounce:
```suggestion
```

util/command_windows.go
```diff
@@ -28,3 +27,25 @@ func NewCommandForWindowsCmd(name string, arg ...string) *Command {
...
+	command := new(Command)
+	command.name = "powershell"
```
petemounce, line 33: Maybe not this PR, but I gather that powershell-core's binary is `pwsh`. 

Is there a way this choice might be later threaded through without a breaking change? I guess it would be a property on the `service` like `powershell-binary`, defaulting to `powershell`...?

petemounce (commented):
@cthiel42 thanks very much for your increment, and apologies again for latency. Couple more suggestions.

I see I now have an `Approve workflows to run` button, so I've blithely pressed that and can see GHA has done its thing. Not sure if that button is needed each time; will see. Still no Travis, though; I lack access to address that.

I'd be cool with this merging once CI is run and green. @ripienaar / @aelsabbahy, any thoughts?

system/service_win.go
cthiel42: This is a lot cleaner so I switched to this.

system/service_windows.go
cthiel42: I switched it to %q and it works fine. Service names cannot actually have spaces in them (only the display name can). I think technically there could be an injection risk if quoting is not used, so might as well use it.

util/command_windows.go
cthiel42: this is removed now

util/command_windows.go
cthiel42: Are you thinking of using pwsh if its available, and falling back to powershell if its not available?

petemounce: No, I was thinking of this being left hard-coded to `powershell` and then (later in a follow-up) providing the ability to a test-author to override that with (something like) `powershell_binary: pwsh` in the test YAML definition. If that test-attribute is omitted, the default no-value choice would be `powershell` as in this PR.

petemounce:
@cthiel42 would you mind addressing https://github.com/goss-org/goss/actions/runs/21811022610/job/63155554726?pr=1017#step:4:34, which ought to just be a `go fmt` to make it progress?

cthiel42:
> @cthiel42 would you mind addressing https://github.com/goss-org/goss/actions/runs/21811022610/job/63155554726?pr=1017#step:4:34, which ought to just be a `go fmt` to make it progress?

@petemounce This should hopefully be resolved now

util/command_windows.go
cthiel42: I haven't looked too much into this, but I think as long as we can get the config to be accessible within this method, we could do that.

kgaughan:
@cthiel42 It looks like that's expired, so could you try merging from master or rebasing against master?

kgaughan:
@cthiel42  Sorry to bother you again, but could you try merging from master again? I'm hoping I'll be able to trigger the integration tests if you do so.

cthiel42:
@kgaughan Hopefully there are integration tests waiting for your approval now

kgaughan:
Due to some unfortunate timing, and earlier PR (#1092) introduced a race conditions in a couple of the tests, but it didn't manifest until it was merged into master. I just approved and merged #1098, which fixes the race introduced to the test suite. Unfortunately, I can't merge that into your PR branch, so I'll need ask you to do it again. Hopefully, this'll be the last time. Apologies about this!

cthiel42:
@kgaughan Not a problem! Should be good for another go now

kgaughan (approved):

PR #1020, opened 2025-03-14
Feature request - Add file.capabilities (Linux)

ahuret:
##### Checklist

### Description of change

This is a new test feature included in `file` test called `capabilities`.
It aims to add a simple way to test for file Linux **effective** [capabilities](https://man7.org/linux/man-pages/man7/capabilities.7.html) .
The test uses this Golang package https://github.com/syndtr/gocapability .

Usage:

```yaml
file:
  my-file:
    exists: true
    capabilities:
        - 38 # CAP_PERFMON
        - 21 # CAP_SYS_ADMIN
```
It will test if the file has the capabilities `CAP_PERFMON` and `CAP_SYS_ADMIN`, if not the test fails.
One problem with that is to use the `enum` syntax to describe capabilities (see [caps enums](https://github.com/syndtr/gocapability/blob/42c35b437635/capability/enum.go#L47)). I think it would be easier for user to write the name of the capability instead of a number but it requires to add code to transform enum in string, so I was not sure about it in term of code simplicity.
 
This PR is related to this issue https://github.com/goss-org/goss/issues/971 .

----
📚 Documentation preview 📚: https://goss--1020.org.readthedocs.build/en/1020/

### DIFF
--- a/docs/goss.yaml
+++ b/docs/goss.yaml
@@ -40,6 +40,7 @@ file:
     sha512: cb71b1940dc879a3688bd502846bff6316dd537bbe917484964fe0f098e9245d80958258dc3bd6297bf42d5bd978cbe2c03d077d4ed45b2b1ed9cd831ceb1bd0
     linked-to: /usr/sbin/sendmail.sendmail
     skip: false
+    capabilities: []
 # Inline templates are not valid json/yaml that could be validated as-in by json schema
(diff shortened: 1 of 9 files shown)


kgaughan:
Are you looking to proceed with this at any point? There are some merge conflicts that need resolving.

PR #1022, opened 2025-04-23
fix: Use Prometheus standard content-type

### LINKED ISSUE #1010: goss /healthz endpoint response is rejected by default by prometheus 3.0: unsupported Content-Type "application/vnd.goss-prometheus"
**Describe the bug**

Prometheus 3.0 is now rejecting goss /healthz endpoint responses

**Workaround**

In the prometheus scrape configuration documented at https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config, specify the  `fallbackScrapeProtocol: PrometheusText0.0.4`

**How To Reproduce**

`goss --gossfile goss-integration.yaml serve --listen-addr :8080 --format prometheus`

**Expected Behavior**

Prometheus accepts goss http responses without error.

**Actual Behavior**

The UI displays 
> Error scraping target: received unsupported Content-Type "application/vnd.goss-prometheus" and no fallback_scrape_protocol specified for target

The prometheus logs include the following error

`time=2025-01-02T13:45:26.208Z level=ERROR source=scrape.go:1585 msg="Failed to determine correct type of scrape target." component="scrape manager" scrape_pool=podMonitor/intranet-proxy/goss-pod-monitor/0 target=http://10.42.13.15:8080/healthz content_type=application/vnd.goss-prometheus fallback_media_type="" err="received unsupported Content-Type \"application/vnd.goss-prometheus\" and no fallback_scrape_protocol specified for target"`

**Environment:**
 - Version of goss: `goss version v0.4.9`
 - OS/Distribution version (if applicable)

**References**

https://prometheus.io/docs/prometheus/latest/migration/#scrape-protocols
> Prometheus v3 is more strict concerning the Content-Type header received when scraping. Prometheus v2 would default to the standard Prometheus text protocol if the target being scraped did not specify a Content-Type header or if the header was unparsable or unrecognised. This could lead to incorrect data being parsed in the scrape. Prometheus v3 will now fail the scrape in such cases.
> 
> If a scrape target is not providing the correct Content-Type header the fallback protocol can be specified using the fallback_scrape_protocol parameter.
> This is a breaking change as scrapes that may have succeeded with Prometheus v2 may now fail if this fallback protocol is not specified.

https://prometheus.io/docs/instrumenting/exposition_formats/#text-based-format
Basic info
> HTTP Content-Type 	text/plain; version=0.0.4 (A missing version value will lead to a fall-back to the most recent text format version.)

carlosjgp:
> Contributors guide: ./CONTRIBUTING.md

Looks like this is missing...?

##### Checklist

### Description of change

fixes #1010 
Use Prometheus standard content-type
https://prometheus.io/docs/instrumenting/exposition_formats/

### DIFF
--- a/serve.go
+++ b/serve.go
@@ -186,5 +186,9 @@ func (h healthHandler) responseContentType(outputName string) string {
 	if outputName == "json" {
 		return "application/json"
 	}
+	if outputName == "prometheus" {
+		return "text/plain; version=0.0.4"
+	}
+
 	return fmt.Sprintf("%s%s", mediaTypePrefix, outputName)
 }
--- a/serve_test.go
+++ b/serve_test.go
@@ -117,6 +117,15 @@ func TestServeNegotiatingContent(t *testing.T) {
 			expectedHTTPStatus:  http.StatusOK,
 			expectedContentType: "application/json",
 		},
+		"accept prometheus": {
+			acceptHeader: []string{
+				"text/plain; version=0.0.4",
+			},
+			outputFormat:        "prometheus",
+			specFile:            filepath.Join("testdata", "passing.goss.yaml"),
+			expectedHTTPStatus:  http.StatusOK,
+			expectedContentType: "text/plain; version=0.0.4",
+		},
 		"accept header contains vendor-specific output format different from process-level": {
 			acceptHeader: []string{
 				"application/vnd.goss-rspecish",


carlosjgp:
We are seeing the same errors as this [PR](https://github.com/goss-org/goss/pull/1013)... something seems to be broken on the integration tests :(

aelsabbahy:
I might have to merge this and fix the integration tests as a follow up PR.

The CI was migrated from Travis -> GitHub, but the image publishing doesn't seem to be working properly so the images (and expectations) drift.

carlosjgp:
> I might have to merge this and fix the integration tests as a follow up PR.
> 
> The CI was migrated from Travis -> GitHub, but the image publishing doesn't seem to be working properly so the images (and expectations) drift.


Hey @aelsabbahy do you need help with this?

PR #1024, opened 2025-05-09
Feat: Make cli argument `vars` to be slice of string

### LINKED ISSUE #1023: Support multiple vars files
**Describe the feature:**

Allow multiple variable files to be specified on the command line. Currently we can only define one variable file using `--vars var-file.yaml`. 

Motivation for this is scenario when `goss` is used in combination with `ansible`. In this scenario user cannot specify --vars files to be for example `./ansible/inventory/prod/group_vars/all.yaml`, `./ansible/inventory/prod/group_vars/v1.32.yaml` and `./tests/my-extra-goss-vars.yaml`.

Advantage of being able to specify this is in repositories that build multiple images with packer (using packer goss provisioner plugin) in which images are provisioned by ansible and then goss tested. In that case the test file has to redefine variables which are already defined in ansible (eg. package versions, which packages are installed, extra packages, etc).

This way there is no need to redefine expected outputs into goss-vars file(s). 

**Example**

Buildling multiple images of kubernetes versions with packer. Packer uses goss plugin. User passes multiple files into `goss_vars_files`. The pipeline will configure packer variable `var.kubernetes_series` to be equal to CI var `KUBERNETES_SERIES` (filled by using `matrix` feature - matrix of variables to build multiple images, each with different kubernetes series value).

Final `goss` command example (with packer it will be invoked over the plugin interface):
```bash
goss \
  --vars "./ansible/inventory/prod/group_vars/all.yaml" \
  --vars "./ansible/inventory/prod/group_vars/v${KUBERNETES_SERIES}.yaml" \
  --vars "./tests/common-goss-vars.yaml" \
  --vars-inline "{ kubernetes_series: \"${KUBERNETES_SERIES}\" }" \
  validate
```

Content of file `./ansible/inventory/prod/group_vars/all.yaml`:

```yaml
kubernetes_deb_repo: "https://pkgs.k8s.io/core:/stable:/{{ kubernetes_series }}/deb/"
```

Content of file `./ansible/inventory/prod/group_vars/v1.32.yaml`:

```yaml
---
kubernetes_cni_deb_version: "1.6.0-1.1"
```

Then goss file `goss.yaml` can contain:
```
package:
  kubelet:
    installed: true
    versions:
    - {{ kubernetes_cni_deb_version }}
```

Support entering `--vars` multiple times. Override empty keys by subsequently defined file.

I haven't found any. Let me know if you see one.

Lirt:
This change allows passing multiple variables files as input to CLI. Order of files is important as overlapped non-empty keys in yaml or json files will be overwritten by subsequent file.

This change is backwards compatible because CLI library implements slice definition as repetition of the `--vars` argument. Using goss with single file produces the same outputs with the code change as it was before.

Usage:
```
./goss --vars file1.yaml --vars file2.yaml validate
```

##### Checklist

`make test-all` doesn't exist anymore. Probably should be `make test-int-all`. `make test` passes correctly.

### Description of change

Closes https://github.com/goss-org/goss/issues/1023

----
📚 Documentation preview 📚: https://goss--1024.org.readthedocs.build/en/1024/


Lirt:
Hello @aelsabbahy,

any time expectation when you could look at this change?

Lirt:
Ping for visibility. If there is low amount of maintainers, I can offer my help.

kgaughan:
@Lirt Could you fix the merge conflicts on this PR? I suspect that the main cause is that urfave/cli has been updated to v3.

Lirt:
Hello @kgaughan,

I rebased the PR. There was only small diff in the config.

kgaughan:
@Lirt I've triggered the test suite and it looks like you've fallen afoul of the markdown linter. Could you fix that up? Everything else is looking good though.

Lirt:
Hey, I've updated the line length to match linter expectations.

kgaughan (approved):
Excellent! Thanks for that!

PR #1029, opened 2025-06-25
Set loglevel for the Stale cache logs

akarasik:
When the loglevel for a log is not set, the loglevel filter will ignore that and it will be printed regardless of the loglevel.
This will allow the loglevel filter to filter those out when using WARN or ERROR.

More information : https://github.com/goss-org/goss/issues/991

##### Checklist

- [x] `make test` (UNIX) passes. CI will also test this

### Description of change

When the loglevel for a log is not set, the loglevel filter will ignore that and it will be printed regardless of the loglevel.
This will allow the loglevel filter to filter those out when using WARN or ERROR.

More information : https://github.com/goss-org/goss/issues/991


kgaughan:
Resolved by #1092, though there are unfortunately some issues with it still due to data races.

Thanks though!

Issue #1034, opened 2025-08-08
Retry Mechanism for Package, DNS and Command Tests


j1shnu:
**Describe the feature:**

In environments like mine, where automated infrastructure changes (like package installation/removal) occur frequently or concurrently, Goss tests may incorrectly report failures if they happen to run during transient states (eg: while a package is being uninstalled or reinstalled).

To make Goss more resilient in such situations, I propose adding retry logic for packages testing. This would allow Goss to re-check a resource after a short delay, mitigating false negatives caused by temporary inconsistencies during automation overlaps.

**Describe the solution you'd like:**

The proposed solution introduces two new optional fields for each resource block:
```
package:
  httpd:
    retry: true             # Enables retry mechanism
    retry_delay: 180        # Delay (in seconds) before retrying the check
    installed: true
    name: httpd
    versions:
      - 2.2.15
    skip: false
```
When `retry: true` is set:

- Goss should perform the initial check.
- If the check fails, Goss should wait for `retry_delay` seconds and then re-run the check once again before reporting the failure.

This simple addition would be especially useful in setups where:

- Packages are reinstalled periodically.
- Tests may overlap with these operations, leading to race conditions and false alerts.

### Additional Notes:
I am willing to work on implementing this feature and contributing it to the project. Before I begin, I’d love to get feedback from the maintainers:

- Any preferred approach for implementing retries for package testing?
- I also would like to implement similar mechanism for DNS and Command tests too.

Looking forward to hearing your thoughts and thanks for maintaining such a useful tool!

ns-mjames:
@aelsabbahy I think this is a much needed feature, not just in packages but in all the test. We do have a global retry option, but If we need to selectively retry a specific test, this would be much useful. Can we get it done ?

ns-mjames:
@aelsabbahy could you help with this ?

j1shnu:
A slight change in the approach I mentioned. I would love to use this mechanism in this way:

```
package:
  httpd:
    retry_count: 1          # Enables retry mechanism when greater than 0 and retry number of time mentioned here
    retry_delay: 180        # Delay (in seconds) before retrying the check
    installed: true
    name: httpd
    versions:
    - 2.2.15
    skip: false
```

j1shnu:
@aelsabbahy I have implemented the same and tested in my environment. Everything is working as expected. Here is the PR:
https://github.com/goss-org/goss/pull/1037

kgaughan:
[...]

PR #1037, opened 2025-09-08
Retry Mechanism for Package, DNS and Command Tests

j1shnu:
##### Checklist

### Description of change

Added retry mechanism for package/DNS/command testing. Details are mentioned in issue [#1034](https://github.com/goss-org/goss/issues/1034). The goss file syntax will be the following:

````yaml
package:
  httpd:
    retry_count: 1      # Enables retry mechanism
    retry_delay: 10000  # Delay in milliseconds before retrying the check; duration strings like 10s also work
    installed: true
    name: httpd
    versions:
      - 2.2.15
    skip: false

dns:
  CNAME:c.dnstest.io:
    resolvable: true
    server: 208.67.222.222
    addrs:
      - "a.dnstest.io."
    retry_count: 2      # Enables retry mechanism
    retry_delay: 500ms  # Delay before retrying the check; bare numeric values are milliseconds

command:
  'go version':
    exit-status: 0
    exec: "go version"
    stdout:
      - go version go1.6 linux/amd64
    stderr: []
    timeout: 10000
    skip: false
    retry_count: 1      # Enables retry mechanism
    retry_delay: 5s     # Delay before each retry attempt; bare numeric values are milliseconds
````

----
📚 Documentation preview 📚: https://goss--1037.org.readthedocs.build/en/1037/

### CODE REVIEW

**Comment by ns-mjames** [comment]
@aelsabbahy Could you help merge this pr? This is imperative for our testing env

**Comment by petemounce** [comment]
I've hit the workflow-approve button for the github actions, but Travis is still a required check and I cannot address its not running.

Please would you add some test-coverage to exercise this code-path? Retries as a feature can become quite fiddly because the feature tends to grow as people decide they want also "exponential back-off" "... with jitter" "... conditional on x" and so on - is there a best-of-breed golang library that might be used instead of a bespoke implementation that this contribution begins? My thought is longer-tail support would be reduced if so, and the testing burden I'm asking for would probably be reduced were a library used instead of something bespoke.

Also, what do you think about making the delay interval units be milliseconds to support finer-grained intervals than seconds - or is this not a concern as the input will accept floats like `0.5`? My thought is that being limited to 1s as the finest grained interval would be limiting in cases where goss tests are used as a health-check expecting to evaluate several times a minute.

**Comment by j1shnu** [comment]
Thanks @petemounce, these are helpful suggestions.

I’ve added focused test coverage for the retry path. The new tests cover `retry_delay` parsing/serialization and retry behavior for `package`, `dns`, and `command`, including sub second retry intervals.

On the implementation approach: I kept it in-tree for now rather than adding a retry dependency. The current behavior is still fixed-attempt/fixed-delay only, so I consolidated the retry logic into a shared helper to reduce duplication. If the project would strongly prefer a library here, I’m happy to revisit that.

On delay units: I updated `retry_delay` so bare numeric values are interpreted as milliseconds, and it also accepts Go style duration strings such as `500ms`, `2s`, and `1m`. That gives finer-grained retry intervals for health check style usage. Bare floats are also treated as milliseconds, although explicit duration strings are probably the clearest form going forward.

I don’t think I can address the Travis required check issue from the fork side; it looks like repository side required check configuration rather than a failing code path in the PR.

**Comment by kgaughan** [comment]
The test suite has been fixed up and Travis is no longer in the mix. It could be worth doing a merge from master or rebase. If you do that, @petemounce should be able to poke Github Actions to kick off the integration test workflow.

**Comment by kgaughan** [comment]
@j1shnu There's a merge conflict: could you resolve it when you've a chance, please?

**Comment by j1shnu** [comment]
@kgaughan Done.

File: resource/retry_test.go [hunk 1]
```diff
@@ -0,0 +1,309 @@
...
+					return []string{"127.0.0.1"}, nil
+				},
+			}
+		},
+	}
+
+	dns := &DNS{
+		id:         "localhost",
+		Resolvable: true,
+		Addrs:      []interface{}{"127.0.0.1"},
```
kgaughan:
```suggestion
		Addrs:      []any{"127.0.0.1"},
```

resource/retry_test.go
```diff
@@ -0,0 +1,309 @@
...
+				},
+			}
+		},
+	}
+
+	cmd := &Command{
+		id:         "echo ok",
+		Exec:       "echo ok",
+		ExitStatus: 0,
+		Stdout:     []interface{}{"ok"},
```
kgaughan:
```suggestion
		Stdout:     []any{"ok"},
```

resource/retry_test.go
```diff
@@ -0,0 +1,309 @@
...
+					return []string{"1.0.0"}, nil
+				},
+			}
+		},
+	}
+
+	pkg := &Package{
+		id:         "nginx",
+		Installed:  true,
+		Versions:   []interface{}{"1.0.0"},
```
kgaughan:
```suggestion
		Versions:   []any{"1.0.0"},
```

kgaughan (commented):
This is just a few minor nits I spotted on a quick scan of the code. I'll post something a bit more substantial tomorrow.

kgaughan:
The code looks fine. I applied the fixes for the nits myself. I'll kick off the tests now to make sure everything's still good.

kgaughan (approved):

Issue #1015, opened 2025-01-24
Debug output is unable to print command names


aieri:
**Describe the bug**
When invoking debug logging for command checks I would expect the output to include the check name and resemble what was shown in https://github.com/goss-org/goss/issues/816#issuecomment-1730357116.
In place of the check name I see `%!s(<nil>)` instead.

**How To Reproduce**

Given a goss.yaml as follows:
```
command:
  foo:
    exec: echo "this is a test"
    exit-status: 0
```

**Expected Behavior**

```
$ goss -l debug v
2025-01-24T20:54:31Z [DEBUG] Setting log level to DEBUG
2025-01-24T20:54:31Z [DEBUG][Command][foo][stdout]  this is a test
2025-01-24T20:54:31Z [DEBUG][Command][foo][stdout]  
.

Total Duration: 0.002s
Count: 1, Failed: 0, Skipped: 0
2025-01-24T20:54:31Z [DEBUG] OK SUMMARY: Total Duration: 0.002s Count: 1, Failed: 0, Skipped: 0 
```

**Actual Behavior**

```
$ goss -l debug v
2025-01-24T20:56:04Z [DEBUG] Setting log level to DEBUG
2025-01-24T20:56:04Z [DEBUG][Command][%!s(<nil>)][stdout]  this is a test
2025-01-24T20:56:04Z [DEBUG][Command][%!s(<nil>)][stdout]  
.

Total Duration: 0.002s
Count: 1, Failed: 0, Skipped: 0
2025-01-24T20:56:04Z [DEBUG] OK SUMMARY: Total Duration: 0.002s Count: 1, Failed: 0, Skipped: 0 
```

**Environment:**
 - Version of goss: `goss version v0.4.9`
 - OS/Distribution version (if applicable): `Ubuntu 24.04.1 LTS`

acnickv:
+1

I am seeing the same behavior.

FixTestRepeat:
Does this bug occur in any of the other outputs , only under debug

FixTestRepeat:
hi @aieri  and @acnickv  , I think I've got a build that fixes this issue. 

Snapshot build for Linux and MacOS [is available here](https://github.com/FixTestRepeat/goss/actions/runs/20405216463/artifacts/4935361026) if you'd like to test it.

aieri:
[...]

PR #1046, opened 2025-12-21
Fix debug logging regression bug affecting Command tests

FixTestRepeat:
Fixes a bug introduced by commit 6a8fc96 (according to git bisect + testing ) that broke the logging detail that previously identified the child elements below the Command key.  Examples are visible in #1015

##### Checklist

### Description of change

My understanding is that when using strings to identify context, the lookup worked fine. The pull request #922  that included commit 6a8fc96 didn't mention making this change explicitly, only that it included fixes to resolve some linting warnings.  Since the code still compiled I think this bug went unnoticed. 

Open to suggestions/feedback on the method to fix this bug 'correctly' if this implementation isnt the best.

### DIFF
--- a/system/command.go
+++ b/system/command.go
@@ -8,8 +8,16 @@ import (
 	"os/exec"
 	"time"
 
+
 	"github.com/goss-org/goss/util"
+	
 )
+
+// CommandIDKeyType is the unique key for command IDs in context.
+type CommandIDKeyType struct{}
+
+// CommandIDKey is the only instance that must be used everywhere.
+var CommandIDKey = CommandIDKeyType{}
 
 type Command interface {
 	Command() string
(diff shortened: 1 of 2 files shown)


ns-mjames:
@aelsabbahy could you approve this ?

kgaughan (commented):
This looks fine to me in principle, but it'd be really useful if there was a regression test too.

Could you merge from master also?

kgaughan:
As the original creator of the PR hasn't responded, I think I'll create a separate PR that's a slight variation on this. I'll do it after I'm happy that #1053 can be merged, as there are a few spots where I'll want to replace `idKey{}` in that too.

system/command.go
```diff
@@ -8,9 +8,17 @@ import (
...
+
 	"github.com/goss-org/goss/util"
+	
 )
 
+// CommandIDKeyType is the unique key for command IDs in context.
+type CommandIDKeyType struct{}
+
+// CommandIDKey is the only instance that must be used everywhere.
+var CommandIDKey = CommandIDKeyType{}
```
kgaughan: This is part of the generalisation I want to make: it generalises what you're doing to make it usable in more places.

```suggestion
// ContextKey is  for minting unique keys in a context.
type ContextKey struct{}

// CommandIDKey is the only instance that must be used everywhere.
var CommandIDKey = ContextKey{}
```

system/command.go
```diff
@@ -8,9 +8,17 @@ import (
 	"os/exec"
 	"time"
 
+
 	"github.com/goss-org/goss/util"
+
```
kgaughan:
```suggestion
```

system/command.go
```diff
@@ -8,9 +8,16 @@ import (
 	"os/exec"
 	"time"
 
+
```
kgaughan:
```suggestion
```

kgaughan (approved):

PR #1052, opened 2026-04-12
GoReleaser config and multi-platform image builds

### LINKED ISSUE #1012: Support for linux/ppc64le architecture
**Describe the feature:**
Add linux ppc64le build

https://github.com/goss-org/goss/pull/1011

We will build it ourselves in the interim.

### LINKED ISSUE #1065: s390x not working with install.sh (https://goss.rocks/install)
**Describe the bug**

We are using goss with the architecture s390x using Ubuntu.

To detect the architecture, the install.sh script has the following logic:
```
arch=""
if [ "$(uname -m)" = "x86_64" ]; then
    arch="amd64"
elif [ "$(uname -m)" = "aarch32" ]; then
    arch="arm"
elif [ "$(uname -m)" = "aarch64" ] || [ "$(uname -m)" = "arm64" ]; then
    arch="arm64"
else
    arch="386"
fi
```

"$(uname -m)" in s390x is "s390x", same as in the s390x release of goss. As the default installs the `386` goss, it does not work. Either the default should be "$(uname -m)" or a new elif branch should be added for s390x (with both strings the same value).

**How To Reproduce**

Just run `curl -fsSL https://goss.rocks/install | sh` in a s390x machine.

**Expected Behavior**

The install script should install the correct binary for the architecture.

**Actual Behavior**

The s390x binary should be installed.

**Environment:**
 - Version of goss
 - Ubuntu 24.04 with arch s390x.

kgaughan:
##### Checklist

### Description of change

Using GoReleaser simplifies the build process substantially, as it handles almost all the logic that needed to be in the makefile and the pipelines themselves.

As my repo is a fork and I've specified `ghcr.io/goss-org/goss` as the image name, GoReleaser is (naturally) unable to push the final image, leading to the following expected behaviour:

```
ERROR: failed to push ghcr.io/goss-org/goss:latest: failed to authorize: failed to fetch anonymous token:
unexpected status from GET request to https://ghcr.io/token?scope=repository%3Agoss-org%2Fgoss%3Apull%2Cpush&service=ghcr.io: 403 Forbidden
```

However, the 'GITHUB_TOKEN' secret should allow this to work on the main repo. This does at least show that it's able get through the multiplatform image build without issue.

Owing to how GoReleaser works, I added something so that generating the .sha256 files for the {d,dc,k}goss scripts wouldn't cause GoReleaser to treat the repo as dirty. It also encourages building archives rather than attaching uncompressed binaries to releases.

Incidentally, this means that the goss images are now built with SBOMs. It's straightforward to extend the configuration to also build packages and generate SBOMs for the release artifacts too.

I believe this should mean the Travis CI pipelines can go away.

I'd also note that the pipeline described by docker-goss.yaml is broken owing to aquasecurity/trivy-action's 0.24.0 tag no longer exists (likely due to the issues with and attacks on Trivy over the past few months). This new pipeline can do everything as far as release builds go, so I think that could be trimmed down to just a simple build of the master image on pushes to the master branch.

#### Incidental changes:

* Alpine base image bumped to 3.23.
* Fixes #1065.
* Fixes #1012.


kgaughan:
That Ubuntu integration test wasn't failing previously. I'm going to dig into it, but I'll have to wait until the weekend.

kgaughan:
The issues look to be mostly down to changes in test services and new OS releases that bumped some version numbers. There are a few dummy services (http in, for example) that look to either not be starting to not accessible.

kgaughan:
I investigated this while on the train yesterday. The issue appear to be because dnstest.io, which the integration tests rely on, is now parked.

A sensible alternative would be to spin up something like dnsmasq internally and have it act as the resolver for the tests. dnsmasq can serve authoritative zones, and goss's `dns` test module supports the specification of a server to query, so specifying `server: 127.0.0.1:8053` should work. GH Actions supports service containers, though if necessary it wouldn't be the end of the world to spin up dnsmasq in the background if necessary. I'll create an issue for this.

kgaughan:
My integration tests fixes work, and I've opened a pull request for them: #1061.

kgaughan:
Looks like the previous merge commit only failed because of unfortunate timing.

aelsabbahy:
Can you add a small note in the readme or in a new markdown file that explains how to run/test this without forking the repo and using GitHub?

I assume it can be run with goreleaser locally for testing, right?

kgaughan:
Sure!

Yeah, it can be run locally. All the configuration is in `.goreleaser.yaml` to allow a full build of all targets.

I'll update the documentation to make sure it covers this.

kgaughan:
It's probably best to merge #1068 before this, as there'll be a tweak needed to `install.sh` to decompress the binaries.

kgaughan:
@aelsabbahy I've pushed up an example of how to built it with goreleaser.

aelsabbahy (approved):

PR #1053, opened 2026-04-15
feat: add registry resource type for Windows

### LINKED ISSUE #616: Goss should be able to assert against Windows Registry state

Blankf:
> **Depends on:** #1055 (fix: bytes.Reader issue in HavePatternsMatcher)

## Summary

Adds a new `registry` resource type that validates Windows registry keys natively using the `golang.org/x/sys/windows/registry` API. This replaces the need to shell out to PowerShell for registry checks.

Closes #616

## Motivation

The ansible-lockdown Windows CIS audit repos (Windows 2016/2019/2025) currently rely entirely on the `command` resource shelling out to PowerShell for ~300+ registry checks per benchmark. This is slow (~180ms per check due to PowerShell process spawn) and verbose.

With the native `registry` resource, the same checks run in milliseconds via direct Windows API calls.

**Benchmark results (250 identical checks):**
| Method | Time |
|--------|------|
| Native `registry` resource | **0.020s** |
| `command` + PowerShell | **43.784s** |

## Gossfile syntax

```yaml
registry:
  HKLM\SOFTWARE\Microsoft\Windows NT\CurrentVersion\ProductName:
    exists: true
    value: "Windows Server 2025 Datacenter"
    type: REG_SZ
```

For value names containing backslashes (e.g. HardenedPaths UNC entries), use `::` as an explicit separator:

```yaml
registry:
  HKLM\SOFTWARE\Policies\Microsoft\Windows\NetworkProvider\HardenedPaths::\\*\NETLOGON:
    exists: true
    value: "RequireMutualAuthentication=1, RequireIntegrity=1"
```

## Details

- Supported hives: HKLM, HKCU, HKCR, HKU, HKCC
- Supported types: REG_SZ, REG_EXPAND_SZ, REG_DWORD, REG_QWORD, REG_BINARY, REG_MULTI_SZ
- Opens keys with `QUERY_VALUE` access (read-only, least privilege)
- On non-Windows platforms, returns an error ("registry resource is only supported on Windows"), following the `NullPackage` pattern
- Cross-compilation from Linux continues to work

## Files changed

**New files:**
- `system/registry.go` -- Registry interface + key path parser
- `system/registry_test.go` -- 17 unit tests for the parser
- `system/registry_notwindows.go` -- Non-Windows stub
- `system/registry_windows.go` -- Windows implementation
- `resource/registry.go` -- Resource type definition
- `integration-tests/goss/windows/tests/registry.goss.yaml` -- Integration tests
- `docs/windows-parity-progress.md` -- Progress tracker

**Modified files:**
- `system/system.go` -- Added `NewRegistry` factory
- `resource/resource_list_genny.go` + `resource/resource_list.go` -- Added RegistryMap
- `goss_config.go` -- Wired into config
- `add.go` -- Added to AddResource switch
- `cmd/goss/goss.go` -- Added CLI subcommand
- `docs/schema.yaml` -- Added schema definition
- `docs/platforms.md` -- Updated platform matrix

## Test plan

- [x] `go vet ./...` passes
- [x] `go test ./...` passes (17 new parser tests + all existing)
- [x] `GOOS=windows GOARCH=amd64 go build` cross-compiles successfully
- [x] `gosec ./system/ ./resource/` -- zero findings in new code
- [x] Tested on Windows Server 2025 -- registry checks work correctly
- [x] Tested on Windows Server 2025 -- existing `command` resource still works
- [x] Benchmarked: 250 registry checks in 0.020s vs 43.784s via PowerShell


cmd/goss/goss.go
```diff
@@ -401,6 +401,14 @@ func main() {
 						return goss.AddResources(c.GlobalString("gossfile"), resource.InterfaceResourceName, c.Args(), newRuntimeConfigFromCLI(c))
 					},
 				},
+				{
```
kgaughan: Just an FYI: #1060 will break this if/when it's merged. The fix is easy though: it's just an indentation change and `Action: func(c *cli.Context) error {` needs to become `Action: func(_ context.Context, c *cli.Context) error {`.

kgaughan:
It was merged just a few hours ago, so you'll need to rebase or merge from master to resolve the issues. It should be a relatively straightforward merge conflict to resolve.

Blankf:
Rebased on top of current `master` (`906c395`). Scope of this update is limited to what's needed to resolve the conflicts from #1060 and get the branch buildable again:

- **`cmd/goss/goss.go`** — the new `registry` `add` subcommand now uses the v3 API:
  ```go
  Action: func(ctx context.Context, c *cli.Command) error {
      fatalAlphaIfNeeded(c)
      return goss.AddResources(c.String("gossfile"), resource.RegistryResourceName, c.Args().Slice(), newRuntimeConfigFromCLI(c))
  },
  ```
  Note: @kgaughan's hint suggested `*cli.Context`, but the actual v3 rename in #1060 is to `*cli.Command`, so that's what I used (matches every other subcommand in the file).

  Also fixed a placement bug the auto-merge introduced: the block ended up outside the `add` command's `Commands` slice, which wouldn't compile. It's now back inside the slice next to `interface`.

- **`go.mod`** — kept upstream's Go 1.25.11 and bumped deps; the only PR-owned change is promoting `golang.org/x/sys` from indirect to direct require. `go mod tidy` clean, no `go.sum` changes needed.

- **`goss_config.go`** — one gofmt-driven change: the `Registries` struct tag alignment (single-space fixed to match the column).

- **`system/registry.go`** — one gofmt-driven change: doc-comment indentation reflow (Go 1.25 stricter comment formatter).

Verified locally in `golang:1.25`:
- `go vet ./...` clean
- `go test ./...` all packages pass
- `go build ./...` (linux/amd64) clean
- `GOOS=windows GOARCH=amd64 go build ./...` clean
- `gofmt -l .` clean

Diff shape unchanged: 15 files, +711/-2 (was +726/-3; the delta is entirely from the tighter `go.mod` hunk).

The dependency note on #1055 stays as-is — this PR doesn't touch any file that #1055 touches, so the two rebase independently, but #1055 is still recommended to land first per the original description.

resource/resource_list.go
```diff
@@ -1529,3 +1529,101 @@ func (ret *HTTPMap) UnmarshalYAML(unmarshal func(v interface{}) error) error {
 	*ret = tmp
 	return nil
 }
+
+type RegistryMap map[string]*Registry
+
+func (r RegistryMap) AppendSysResource(sr string, sys *system.System, config util.Config) (*Registry, error) {
+	ctx := context.WithValue(context.Background(), idKey{}, sr)
```
kgaughan, line 1536: This has put something on my to-do list: we should really be passing the context down from `main()`. I'll look into doing that later.

system/registry.go
```diff
@@ -0,0 +1,95 @@
...
+//   - "HKLM\SOFTWARE\Microsoft\Windows\CurrentVersion\Run"
+//     -> Hive="HKLM", SubKey="SOFTWARE\Microsoft\Windows\CurrentVersion", ValueName="Run"
+//   - "HKLM\SOFTWARE\Microsoft\Windows\"
+//     -> Hive="HKLM", SubKey="SOFTWARE\Microsoft\Windows", ValueName=""
+//   - "HKLM\SOFTWARE\Policies\Microsoft\Windows\NetworkProvider\HardenedPaths::\\*\NETLOGON"
+//     -> Hive="HKLM", SubKey="SOFTWARE\Policies\Microsoft\Windows\NetworkProvider\HardenedPaths",
+//     ValueName="\\*\NETLOGON"
+func parseRegistryKey(key string) (registryPathParts, error) {
+	if key == "" {
+		return registryPathParts{}, errors.New("empty registry key")
```
kgaughan: Could you lift this error out so it's defined once? Something like this:
```go
var errInvalidRegistryKey = errors.New("invalid registry key")

// ...

	if key == "" {
		return registryPathParts{}, fmt.Errorf("%w: empty key", errInvalidRegistryKey)
```

system/registry.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+
+	parts := strings.SplitN(key, `\`, 2)
+	if len(parts) < 2 {
+		return registryPathParts{}, errors.New("invalid registry key: missing subkey path")
```
kgaughan: Ditto: could you lift this error out so it's defined once? Going off the above, this would become:

```suggestion
		return registryPathParts{}, fmt.Errorf("%w: missing subkey path", errInvalidRegistryKey)
```

system/registry.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+
+	hive := strings.ToUpper(parts[0])
+	switch hive {
+	case "HKLM", "HKCU", "HKCR", "HKU", "HKCC":
+	default:
+		return registryPathParts{}, errors.New("invalid registry hive: " + parts[0])
```
kgaughan: Same with this, though I'd suggest something like this:
```go
var errInvalidRegistryHive = errors.New("invalid registry hive")

// ...
	default:
		return registryPathParts{}, fmt.Errorf("%w: %s", errInvalidRegistryHive, parts[0])

File: system/registry.go [hunk 6]
```diff
@@ -0,0 +1,95 @@
...
+	}
+
+	rest := parts[1]
+	if rest == "" {
+		return registryPathParts{}, errors.New("invalid registry key: empty subkey path")
```
kgaughan: Same again:
```suggestion
		return registryPathParts{}, fmt.Errorf("%w: empty subkey path", errInvalidRegistryKey)
```

resource/resource_list.go
kgaughan: #1090 is there as a reminder for me to fix up context handling.

kgaughan (commented):
Looks good! I've some suggestions regarding some tweaks to the error handling: I want to be able to turn on the err113 linter at some point.

kgaughan:
I've merged #1055, so you should be good.

system/registry.go
```diff
@@ -0,0 +1,95 @@
...
+)
+
+type Registry interface {
+	Key() string
+	Exists() (bool, error)
+	Value() (string, error)
+	Type() (string, error)
+}
+
+var ErrRegistryUnsupported = errors.New("registry resource is only supported on Windows")
```
kgaughan:
```suggestion
var (
	ErrRegistryUnsupported = errors.New("registry resource is only supported on Windows")
	errInvalidRegistryKey  = errors.New("invalid registry key")
	errInvalidRegistryHive = errors.New("invalid registry hive")
)
```

system/registry.go
kgaughan:
```suggestion
		return registryPathParts{}, fmt.Errorf("%w: %s", errInvalidRegistryHive, parts[0])
```

system/registry.go
kgaughan:
```suggestion
		return registryPathParts{}, fmt.Errorf("%w: empty key", errInvalidRegistryKey)
```

system/registry.go
```diff
@@ -0,0 +1,99 @@
+package system
+
+import (
+	"errors"
```
kgaughan, line 4: My bad...
```suggestion
	"errors"
	"fmt"
```

kgaughan (approved):
I fixed up those few bits around error handling, and everything looks good. Unfortunately #1103 isn't merged yet, so I expect the test suite might fail after the merge due to a bit of integration test flakiness, but that's obviously nothing to do with your change.

PR #1055, opened 2026-04-20
Fix bytes.Reader issues 

ckbaker10:
##### Checklist

- [x] `make test-all` (UNIX) passes. CI will also test this
- [x] unit and/or integration tests are included (if applicable)
- [ ] documentation is changed or added (if applicable)

### Description of change

`file` resource `contents` checks always reported `object: *bytes.Reader` on failure instead of the actual file content, making it completely useless for debugging.

Root cause: in `ValidateGomegaValue`, the `func() (io.Reader, error)` switch arm called `f()` and passed the raw `io.Reader` as `foundValue` to `HavePatterns`. The matcher scanned it fine, but once consumed, `FailureResult` had nothing left to read and fell through to `fmt.Sprintf("object: %T", actual)`.

Fix: materialize the reader into a string immediately in the switch arm before handing it to the matcher. `HavePatterns` already handles string input, and `FailureResult` already formats strings correctly — so no changes needed there.

Updated three golden test fixtures that were asserting the broken `"object: *strings.Reader"` output.

PR #1057, opened 2026-04-22
Fixed pattern issue

ckbaker10:
##### Checklist

### Description of change

`have-patterns` matcher silently ignored the `/i` (case-insensitive) flag — and any other trailing flags — on `/pattern/flags` style patterns, causing those patterns to always fail when the output contained mixed-case text.

**Root cause:** `sliceToPatterns` classified a string as a regex only when it ended with a bare `/` (`strings.HasSuffix(s, "/")`). A pattern like `/loglevel (verbose|info)/i` ends with `/i`, so it fell through to the `stringPattern` branch and was evaluated with a `strings.Contains` check. The regex and the `/i` flag were both completely ignored. Additionally, `newRegexPattern` stripped only the literal trailing `/` character, so even patterns that did reach the regex path would have had any flags silently included in the compiled expression (causing a compile error or incorrect behaviour depending on the flag string).

**Fix:**

- Replaced the `strings.HasSuffix(s, "/")` check in `sliceToPatterns` with a new `isRegexPattern` helper that correctly recognises `/pattern/`, `/pattern/i`, `/pattern/ms`, etc. (allowing only valid Go RE2 inline flag characters `i`, `m`, `s` after the closing delimiter).
- Updated `newRegexPattern` to split at the last `/`, extract any trailing flags, and prepend them as a Go inline flag group (`(?flags)`) before compiling — e.g. `/loglevel (verbose|info)/i` is compiled as `(?i)loglevel (verbose|info)`.
- All existing behaviour for plain `/pattern/` patterns and literal string patterns is preserved.

**Tests added** (`matchers/have_patterns_test.go`):

- `TestSliceToPatterns_FlaggedRegex` — verifies that patterns with trailing flags are classified as regex patterns, not string-contains patterns.
- `TestNewRegexPattern_CaseInsensitiveFlag` — unit-tests `newRegexPattern` with the exact `sshd -T` output lines from the failing CIS audit controls (5.1.6, 5.1.12, 5.1.14, 5.1.15).
- `TestHavePatternsMatcher_CaseInsensitiveFlag` — end-to-end tests through the full `HavePatternsMatcher`, covering all four SSH-related CIS control failures where `/i` patterns failed against mixed-case output such as `loglevel INFO`.


kgaughan:
@ckbaker10 Could you merge from master, please?

kgaughan:
With #1055 merged, there are now merge conflicts. Would you mind resolving them when you get a chance?

kgaughan:
I did a best-guess merge of the test suite... :crossed_fingers:

kgaughan:
Excellent! One of the few times I've been happy to get a merge conflict!

kgaughan (approved):

PR #1058, opened 2026-05-07
fix: close HTTP response bodies to prevent OOM with many checks

dukelion:
When running 100+ HTTP checks concurrently, response bodies were not always closed after validation. This caused unclosed TCP connections to accumulate, holding TLS state and transport buffers until the client timeout expired, leading to OOM kills under memory limits.

##### Checklist

### Description of change
Fixes OOM when goss is configured with many HTTP checks and limited RAM.

- Add `Close()` to the `system.HTTP` interface and `DefHTTP`
- Defer `sysHTTP.Close()` from HTTP resource validation for guaranteed cleanup
- Always close `resp.Body` when `Close()` is called, including after a body matcher reads it
- Add a unit test covering close-after-body-read behavior

### DIFF
--- a/resource/http.go
+++ b/resource/http.go
@@ -78,6 +78,7 @@ func (u *HTTP) Validate(sys *system.System) []TestResult {
 		RequestHeader: u.RequestHeader, RequestBody: u.RequestBody, Method: u.Method})
 	sysHTTP.SetAllowInsecure(u.AllowInsecure)
 	sysHTTP.SetNoFollowRedirects(u.NoFollowRedirects)
+	defer sysHTTP.Close()
 
 	var results []TestResult
 	results = append(results, ValidateValue(u, "status", u.Status, sysHTTP.Status, skip))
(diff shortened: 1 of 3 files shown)


kgaughan:
Hi! I fixed the integration test pipelines recently. Could you rebase or merge from master? I checked you integration test failures, and most if not all of them would be fixed by that.

dukelion:
Rebased this branch onto current `goss-org/master` and force-pushed with lease.

I also tightened the cleanup path so `DefHTTP.Close()` closes `resp.Body` even after a body matcher reads it, and added a focused unit test for that lifecycle.

Local validation run on Windows:
- `go test ./system -run TestDefHTTPCloseClosesBodyAfterBodyRead`
- `go test ./resource`
- `git diff --check HEAD~1..HEAD`

aelsabbahy (approved):

kgaughan:
The problem is the lack is the .md5 file. That's used to determine if an image should be built from scratch. It's an easy fix, but annoying that I missed it when fixing the integration tests.

kgaughan:
I know what the issue is. @dukelion or @aelsabbahy, could one of you merge from master? The issue is that this PR is still using an older version of `integration-tests/test.sh`.

aelsabbahy:
Merging based on discussion here: https://github.com/goss-org/goss/issues/1049#issuecomment-4887171568

PR #1059, opened 2026-05-11
Remove Cent OS 7 and update the Rocky Linux Apache version

kgaughan:
##### Checklist

(none are applicable)

### Description of change

The Cent OS 7 config isn't referenced in the integration tests, and the check on Rocky Linux for the httpd package is failing due to a newer version being available.

There are further issues, as noted in #1052, with the fact that dnstest.io is now parked, so the DNS integration tests are now breaking. My plan, as noted in #1052, is to update the GH Actions 'Golang ci' pipeline to spin up Dnsmasq to serve the dnstest.io zone so that it can serve the integration test traffic.


kgaughan:
Probably best to merge #1061 before this one. A good chunk of this is already in that PR.

PR #1061, opened 2026-05-18
Replace dnstest.io with a zone served by dnsmasq locally

kgaughan:
##### Checklist

### Description of change

dnstest.io is now parked, so it's no longer suitable for running goss integration tests against.

I've instead replaced it with a zone served by dnsmasq running within the integration test containers. dnsmasq can serve as an authoritative DNS server for at least a single zone, so it's perfect for running integration tests like this. Serving the zone from within the integration test container means they'll be more resilient against issues and the zone can be maintained alongside the integration tests themselves.

Out of necessity, I've also worked on getting Alpine, Ubuntu, and Debian updated to maintained versions. I've had to remove 32-bit integration tests for them due to a lack of 32-bit images of jammy. ~I'm still working on getting Arch working, but haven't had much luck, unfortunately, which may be a blocker against getting this merged, but I think the others cover everything sufficiently as is. I'm hoping to get Arch working with the full integration test suite too.~ Fixed! It turns out I hadn't noticed the bootup wait in `test.sh` wasn't being done for Arch too.

I've added an affordance for people like me who use `podman` rather than `docker` in the form of the `DOCKER_BIN` environment variable: this way, it's a bit easier to run the integration test suite without constantly having to fiddle with `test.sh`.


ripienaar:
I've approved test runs here and there are some failures to investigate

kgaughan:
Hmmm... That was working before. I'll check it out.

kgaughan:
There's an updated version of the kernel for Rocky Linux, which is what broke the test. The ultimate fix for that, I think, is for somebody to run `development/push_images.sh` and get the MD5 files committed.

Ultimately, it might be an idea to get a proper pipeline in place for building and and pushing these images, maybe to the GH container registry rather than Docker Hub.

kgaughan:
It's possible there might be package updates for the other images, but I only have my Chromebook with me currently, and the Linux terminal for that doesn't play nice with docker container builds. Hopefully not though!

kgaughan:
Looks good!

kgaughan:
I think the Travis CI check should probably be disabled and the 'Golang ci' integration tests should be made required instead. The Travis pipeline is just running the integration tests, which is what the 'Golang ci' workflow is doing.

ripienaar (dismissed):

ripienaar:
I dont have access to that for the repo, so will merge this and we can delete the travis yaml and then ask for that to be turned off

ripienaar:
@aelsabbahy looks like we need to add me as a code owner to let me merge things

kgaughan:
This is superseded by #1063.

PR #1063, opened 2026-06-08
Fix the main CI pipelines

kgaughan:
##### Checklist

### Description of change

 * Relevant actions upgraded
 * The integration test pipeline now pulls from ghcr.io where it can.
 * Linux integration tests now run in parallel with the Windows and MacOS tests.
 * dnstest.io tests now use a zone hosted locally dnsmasq, unbreaking all DNS-related integration tests. All images now include dnsmasq and have been updated to reasonably recent versions of each target OS
 * CentOS 7 is removed.
 * Improved ArchLinux support: still not on par with everything else just yet, but it's not as much of a special case.
 * Support for overriding the use of `docker` in `test.sh`
 * Image builds use the `--file` flag for specifying the dockerfile rather than piping it in. This gets rid of some weird behaviours I encountered when building images.
 * `release-build.sh` now leans on `git describe` to get the release tag name.
 * Use `-trimpath` to avoid incorporating build context into binaries: this makes the build more stable.

I'm temporarily leaving in the MD5 check in `integration-tests/test.sh`, but I think it should ultimately be removed once new integration test runner images are built.

I've kept any dependency/toolchain updates out of this as they introduce changes to some of the diagnostic output that breaks the tests. That can be handled separately.

Closes #1059, #1061.


.github/workflows/docker-integration-tests.yaml
```diff
@@ -78,10 +75,10 @@ jobs:
       - name: Build and push tag
         uses: docker/build-push-action@v6
         with:
-          context: .
+          context: integration-tests
           file: ${{ matrix.dockerfile }}
           push: true
           tags: |
-            ghcr.io/${{ github.repository_owner }}/${{ matrix.image }}:latest
+            ghcr.io/${{ github.repository_owner }}/goss_integration_${{ matrix.image }}:latest
```
kgaughan, line 82: If you want me to get rid of the `goss_integration_` prefix, I'm fine with that. I added it to make it obvious that it's not a containerised version of goss with a certain base OS image.

aelsabbahy: I like the prefix, very clear.

aelsabbahy:
Once this is merged. What's the process for publishing/updating the images?

aelsabbahy:
Oh, and many thanks for doing this. Much appreciated!

aelsabbahy (approved):

aelsabbahy:
https://github.com/goss-org/goss/actions/runs/27158027185

kgaughan:
I'll rebase my remaining branches, starting with the dependency updates.

PR #1064, opened 2026-06-08
Bump dependencies

### LINKED ISSUE #1051: Compromised aquasecurity/trivy-action detected in GitHub Actions workflows
### Compromised `aquasecurity/trivy-action` detected in workflow run(s)

Our automated platform at [StepSecurity](https://www.stepsecurity.io) has detected that this repository used a **compromised version of `aquasecurity/trivy-action`** in its GitHub Actions workflows during the recent Trivy incident. I have also manually confirmed that the affected workflow run(s) indeed used the compromised action.

#### What happened?

The `aquasecurity/trivy-action` GitHub Action was compromised, and a malicious version (`v0.69.4`) was published. Workflow runs in this repository executed a compromised SHA of this action, which may have exposed sensitive information such as secrets, environment variables, or build artifacts.

For more details on the incident, see [StepSecurity Blog: Trivy Compromised a Second Time](https://www.stepsecurity.io/blog/trivy-compromised-a-second-time---malicious-v0-69-4-release).

#### Compromised SHAs detected

- `aquasecurity/setup-trivy@e6c2c5e321ed9123bda567646e2f96565e34abe1` (e6c2c5e321ed9123bda567646e2f96565e34abe1)
- `aquasecurity/trivy-action@e0198fd2b6e1679e36d32933941182d9afa82f6f` (0.24.0)

#### Affected workflow runs

| # | Workflow Run | Build Log (compromised step) |
|---|-------------|------------------------------|
| 1 | [23328257040](https://github.com/goss-org/goss/actions/runs/23328257040) | [View compromised action step](https://github.com/goss-org/goss/actions/runs/23328257040/job/67854038921#step:1:30) |

#### Current workflow status

The workflow in this repository is still referencing the action using a **mutable tag** rather than a pinned SHA: [View current workflow](https://github.com/goss-org/goss/blob/master/.github/workflows/docker-goss.yaml#L101).

#### Recommended actions

1. Review the compromised workflow job run(s) and identify if the job had access to any secrets. If yes, consider them exfiltrated and rotate them immediately
2. [Pin GitHub Actions to full-length commit SHAs](https://docs.github.com/en/actions/reference/security/secure-use#using-third-party-actions) to prevent future tag-based supply chain attacks. You can use [StepSecurity Secure Repo](https://app.stepsecurity.io/securerepo) to automatically pin GitHub Actions

#### References

- [StepSecurity Blog: Trivy Compromised a Second Time](https://www.stepsecurity.io/blog/trivy-compromised-a-second-time---malicious-v0-69-4-release)

kgaughan:
##### Checklist

### Description of change

As the Go toolchain is being bumped too, some additional changes were needed in the testdata and also to the code to deal with issues flagged during the build. golangci-lint also needs a bump to the most recent v1.* to avoid the following breakage:

```
$ make
INFO: Starting build fmt
./ci/go-fmt.sh
.../bin/gofmt
valid gofmt
INFO: Starting build lint
go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.59
../../go/pkg/mod/golang.org/x/tools@v0.22.0/internal/tokeninternal/tokeninternal.go:64:9: invalid array length -delta * delta (constant -256 of type int64)
make: *** [Makefile:40: lint] Error 1
```

~A bump to v2 can come later.~ golangci-lint v1.64.8 and go v1.25 aren't playing nice together, forcing a bump to v2.12.2. I've included the required changes to make the linter happy too.

Finally, I've bumped the actions for Trivy in `trivy-schedule.yaml` to fix that pipeline too. Closes #1051.


kgaughan:
I guess I should note why I bumped the toolchain to 1.25.11: when I got the Trivy scans working, it detected a vulnerability in the Go standard library.

testdata/out_matching_basic_failing.1.documentation
```diff
@@ -153,7 +153,7 @@ not to be numerically eq
     42
 Matching: negated_basic_reader: matches:
 Error
-    ContainElements matcher expects an array/slice/map.  Got:
+    ContainElements matcher expects an array/slice/map/iter.Seq/iter.Seq2.  Got:
```
aelsabbahy, line 156: Hmm.. this output seems confusing. What's causing that change?

We may have to wrap this error.

aelsabbahy (commented):
Looks good, small question/feedback since it impacts the user experience.

testdata/out_matching_basic_failing.1.documentation
kgaughan: That's caused by the update of the Go runtime to 1.23+. [iter](https://pkg.go.dev/iter) was introduced in Go 1.23.

kgaughan: I took a second look at the matcher itself, and I think we'd need to override `gomega/matchers.ContainsElementsMatcher` if we wanted to maintain the old error message: https://github.com/onsi/gomega/blob/d72697b90777dcb0813622b58827d27e4a9d3f22/matchers/contain_elements_matcher.go#L18

(Arguably, I'd say mentioning slices would be confusing to a user too, unless they also knew Go or a language with a similar concept.)

I'm good with looking to see if there's a reasonable way to get the fluff out of the error message after the dependencies are updated. I suspect this may require wrapping `ContainsElementsMatcher.Match` rather than relying on embedding to expose the method.

aelsabbahy: Hah, was just getting on to mention that I think it's comming from ContainElementsMatcher.

Yeah, I'm fine with it being a follow up.

aelsabbahy (approved):

scadu:
Do y'all plan to cut a release with these? 👀

PR #1067, opened 2026-06-15
Restore old error messages for the ContainElements matcher

kgaughan:
##### Checklist

### Description of change

Before delegating to the gomega matcher, it checks for whether its argument is a map, array, or slice, which is a subset of the checks done by gomega. As the configuration file won't have an iterator, this precheck means we can deal with the error a bit more appropriately.


aelsabbahy (approved):

aelsabbahy:
Awesome, thanks!

PR #1068, opened 2026-06-15
install.sh: include s390x; ease maintenance with a case statement

### LINKED ISSUE #1065: s390x not working with install.sh (https://goss.rocks/install)

kgaughan:
##### Checklist

### Description of change

Fixes #1065.

The case statement means adding additional machine architectures is easier and also means `uname -m` is invoked only once. It's more lines, but less code.

As this is meant to cover all supported architectures, it's probably best to cleanly error out rather than installing an unsupported binary.

~Given this is meant to be run as root, it's useful to error out if it isn't.~

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -3,11 +3,6 @@
 {
 set -e
 
-if [ "$(id -u)" != "0" ]; then
-  echo "warning: you probably need to be running this as root" >&2
-  exit 1
-fi
-
 LATEST_URL="https://github.com/goss-org/goss/releases/latest"
 LATEST_EFFECTIVE=$(curl -s -L -o /dev/null ${LATEST_URL} -w '%{url_effective}')
 LATEST=${LATEST_EFFECTIVE##*/}
--- a/install.sh
+++ b/install.sh
@@ -3,6 +3,11 @@
 {
 set -e
 
+if [ "$(id -u)" != "0" ]; then
+  echo "warning: you probably need to be running this as root" >&2
+  exit 1
+fi
+
 LATEST_URL="https://github.com/goss-org/goss/releases/latest"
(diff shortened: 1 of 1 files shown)


aelsabbahy:
> Given this is meant to be run as root, it's useful to error out if it isn't.

I would have to test this out, but I thought by overriding the install location, it was possible to run this as non root?

kgaughan:
I could add something to allow a configurable install location.

kgaughan:
Ah, I'd missed the bit that checks if the installation location is writable! :facepalm: Fixed!

kgaughan:
Actually, I'll close this out and handle it in #1052.

PR #1083, opened 2026-07-19
Parallel builds for Linux integration tests

kgaughan:
##### Checklist

### Description of change

While this doesn't speed up integration test runs dramatically (Windows runs impose a baseline of 1min, so now dominate the execution time), it means we can add additional distros if need be at little cost it's easier to see which distro has failing tests.

### DIFF
--- a/.github/workflows/golangci.yaml
+++ b/.github/workflows/golangci.yaml
@@ -19,7 +19,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v7
-      - uses: actions/setup-go@v6
+      - uses: actions/setup-go@v7
         with:
           go-version-file: go.mod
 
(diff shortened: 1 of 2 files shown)


kgaughan:
Well, this run went better than the ones from when I was testing it out! Windows took ~55s when I was testing things. Assuming Windows runs at about the same speed as everything else, that means this change shaves around 50s off the build times.

aelsabbahy (dismissed):

aelsabbahy:
Let me know if you want to use this PR to test your permissions/access.

If there are other github changes that need to be made, happy to help.

aelsabbahy (approved):

aelsabbahy:
Re-approved, curious once I approve it.. You're able to merge it without me, correct?

kgaughan:
It's because of the additional changesets I'd pushed up before I'd realised you'd approved the change.

PR #1054, opened 2026-04-17
Add marks parameter for subset test selection

dukelion:
### Summary
This PR introduces a `marks` feature that allows users to tag resources and run only specific subsets of tests, similar to pytest's `-m` flag. This enables more flexible test workflows in CI/CD and local development.

### Changes

**Core Feature:**
- Added `Marks []string` field to all 16 resource types (addr, command, dns, file, gossfile, group, http, interface, kernel_param, matching, mount, package, port, process, service, user)
- Added `--marks` and `--exclude-marks` CLI flags for `validate` and `serve` commands
- Added `?marks=` and `?exclude_marks=` query parameters to HTTP endpoints
- Added `--marks` flag to `goss add` and `goss autoadd` commands for tagging new resources

**Infrastructure:**
- Thread-safe skip-state handling for concurrent HTTP requests
- Cache isolation by marks (resources filtered by marks don't pollute cache)
- Debug logging for mark filter summary (visible with `-L DEBUG`)

### Usage

```yaml
# goss.yaml
file:
  /etc/nginx/nginx.conf:
    exists: true
    marks:
      - nginx
      - config

port:
  tcp:80:
    listening: true
    marks:
      - nginx

command:
  echo test:
    exit-status: 0
```

```bash
# Run only nginx-related tests
goss validate --marks nginx

# Run all tests EXCEPT config tests
goss validate --exclude-marks config

# HTTP endpoint with marks
curl "http://localhost:8080/healthz?marks=nginx"
```

### Testing
- Unit tests added for mark filtering, config parsing, and HTTP endpoint behavior
- All existing tests pass: `go test -race -count=3 ./...`
- Tested serve-mode with concurrent requests using different marks

### Related
- This PR includes the logger interface refactor (prerequisite for safe concurrent access)
- Closes the gap for users wanting selective test execution without maintaining multiple goss files

----
📚 Documentation preview 📚: https://goss--1054.org.readthedocs.build/en/1054/

PR #1092, opened 2026-08-03
Fix data races in goss serve and mute the cache-miss log by default

### LINKED ISSUE #991: Unable to mute "Stale cache[res]: running tests" with loglevel
**Describe the bug**
Unable to mute "Stale cache[res]: running tests" log output irrespective of log level set.  This is annoying because a container behind an ALB will flood logs unnecessarily.

p.s. the documentation in https://goss.readthedocs.io/en/stable/cli/ suggests that `--loglevel` is per command, but in actual fact it is at the global-level.

**How To Reproduce**

https://gist.github.com/halberom/2081e2652617a973af4db049a66f4d1f

**Expected Behavior**

Changing log level to WARN or at least ERROR should stop "Stale cache[res]: running tests" from being output.

**Actual Behavior**

I have to change the cache duration to 0, which is fine in my situation, but seems odd.

**Environment:**
 - Version of goss - tested with 0.4.8 and 0.4.9
 - OS/Distribution version (if applicable) - Python Docker container, and others.

dukelion:
First piece split out of #1054. Self-contained and depends on nothing else.

## Data races in the serve request path

`goss serve` handles each request in its own goroutine, but three pieces of
process-wide state were mutated from the request path with no synchronisation:

- `color.NoColor`, written by `newHealthHandler`, `getOutputer` and the JSON and
  JUnit outputers, read by the color helpers in `outputs`
- `outStoreFormat` and `currentTemplateFilter` in `store.go`, written while
  loading a gossfile and read while rendering one
- `debug` in `store.go`

Two concurrent probes are enough to trip this. `go test -race ./...` fails on a
clean checkout of `master` today:

```
WARNING: DATA RACE
  github.com/goss-org/goss.newHealthHandler()
WARNING: DATA RACE
  github.com/goss-org/goss.getGossConfig()
  github.com/goss-org/goss.newHealthHandler()
WARNING: DATA RACE
  github.com/goss-org/goss.getOutputer()
  github.com/goss-org/goss.newHealthHandler()
FAIL	github.com/goss-org/goss	9.695s
```

The color state moves behind an accessor pair in the new `outputs/color.go`, so
the `fatih/color` global is touched in one place, and the `store.go` globals get
a mutex. Both keep their existing last-writer-wins behaviour. This makes the
hand-off safe; giving each request its own configuration would be a much larger
change and is not attempted here.

## The cache-miss log line could not be muted (#991)

Every cache miss logged `Stale cache[...], running tests` with no level prefix,
so `logutils` let it through at any `--loglevel`. Anything polling the health
endpoint more often than the cache TTL, a load balancer or a Kubernetes probe
for instance, got a line per request with no way to turn it off.

It now carries `[DEBUG]`, so it shows at `DEBUG` and `TRACE` and stays quiet at
the `INFO` default. The cache-hit line beside it has always been `[TRACE]`.

`--loglevel` was also documented twice, under `serve` and under `validate`, with
two different and partly wrong descriptions of the levels, including a `FATAL`
that goss actually rejects. It is a global option and works before or after the
command name, so it is now documented once under "Global options".

Closes #991

## The -race commit

`Run the unit tests with -race` is last so it can be dropped on its own. These
races survived this long because nothing in CI would have caught them: neither
`make cov`, which the coverage job runs, nor `ci/go-test.sh` passed `-race`.

The suite is clean under `-race -count=2` and `make cov` with `-race` takes
about 24s, but that is only verified on darwin. If it goes red elsewhere for
reasons unrelated to this branch, dropping that commit unblocks the rest.

## For anyone looking at log/slog later

`logutils` is confined to `logs.go`, and there are 14 level-prefixed
`log.Printf` calls across 6 files. That is the whole surface.

A slog migration would replace `newLogFilter`'s body and drop the `[DEBUG]`
prefix constant, since slog carries the level as a value rather than as text
inside the message. The signature maps over cleanly though: `(level string, w
io.Writer) (slog.Handler, error)`, and the `error` return stays earned because
`slog.Level.UnmarshalText` does not know `TRACE`.

The tests here are meant to help rather than get in the way. Their tables say
which levels emit what, which is a statement about behaviour and not about
logutils, so they port as-is with a few lines of plumbing changed per test.
Nothing pinned that contract before, so a migration that quietly broke level
gating had nothing to fail against.

Worth knowing before starting: slog has no `TRACE` level, so goss would need a
custom `slog.Level(-8)`, and the level list in `docs/cli.md` wants another pass
at the same time.

## Testing

Each new test was run against the pre-fix code and fails there:

- `outputs/color_test.go` exercises every outputer concurrently
- `serve_concurrency_test.go` drives the health handler from 16 goroutines
- `logs_test.go` pins which levels mute the cache-miss line, and checks the
  documented levels behave the way `docs/cli.md` says they do

Also run: `gofmt`, `go vet`, `golangci-lint v2.12.2` (0 issues),
`markdownlint-cli2`, `go test -race -count=2 ./...`.

----
📚 Documentation preview 📚: https://goss--1092.org.readthedocs.build/en/1092/

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,7 @@ test:
 	./ci/go-test.sh
 
 cov:
-	go test -coverpkg=./... -coverprofile=c.out ./...
+	go test -race -coverpkg=./... -coverprofile=c.out ./...
 	# go tool cover -func ./c.out
 
 funcov:
--- a/add.go
+++ b/add.go
@@ -13,15 +13,14 @@ import (
 
 // AddResources is a simple wrapper to add multiple resources
 func AddResources(fileName, resourceName string, keys []string, c *util.Config) error {
-	var err error
-	err = setLogLevel(c)
+	if err := setLogLevel(c); err != nil {
+		return err
+	}
+	format, err := getStoreFormatFromFileName(fileName)
 	if err != nil {
 		return err
 	}
-	outStoreFormat, err = getStoreFormatFromFileName(fileName)
-	if err != nil {
-		return err
-	}
+	setStoreFormat(format)
 
 	var gossConfig GossConfig
 	if _, err := os.Stat(fileName); err == nil {
@@ -96,11 +95,11 @@ func AddResource(fileName string, gossConfig GossConfig, resourceName, key strin
 
 // AutoAddResources is a simple wrapper to add multiple resources
 func AutoAddResources(fileName string, keys []string, c *util.Config) error {
-	var err error
-	outStoreFormat, err = getStoreFormatFromFileName(fileName)
+	format, err := getStoreFormatFromFileName(fileName)
 	if err != nil {
 		return err
 	}
+	setStoreFormat(format)
 
 	var gossConfig GossConfig
 	if _, err = os.Stat(fileName); err == nil {
--- a/ci/go-test.sh
+++ b/ci/go-test.sh
@@ -3,7 +3,7 @@ set -euo pipefail
 
 command -v go
 
-go test -coverpkg=./... ./... -coverprofile="c.out"
+go test -race -coverpkg=./... ./... -coverprofile="c.out"
 
 sed 's|github.com/goss-org/goss/||' <"c.out" >"c.out.tmp"
 
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -62,6 +62,23 @@ GLOBAL OPTIONS:
     * `pacman`
     * `rpm`
 
+`--loglevel <level>`, `-L <level>`
+:   Goss logging verbosity level (default: `INFO`).
+    This is a global option: it applies to every command, and may be given either
+    before or after the command name.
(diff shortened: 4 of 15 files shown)


kgaughan (approved):
Looks good!

kgaughan:
Post-merge, the coverage tests failed with detected race conditions. There's no obvious reason for why this would've started happening and not when the original run happened, as there were no intervening commits to the master branch.

dukelion:
it enables `go test -race` globally, so could be new latent bugs, let me have a look.

kgaughan:
Thanks!

dukelion:
I was able to reproduce it locally and it can be fixed by removing t.Parallel() from serve_test.go
Cost is +1.4s (14.4s → 15.8s)

this is a latent bug with racing on access to process-wide logger.

Issue #1095, opened 2026-08-05
Vacuous `stderr: []` matcher


dukelion:
**Describe the bug**
Setting stderr: [] (or stdout: []) on a command resource results in a vacuous assertion. Because the matcher list contains zero items, Goss iterates over zero conditions and immediately marks the test as passed, even if standard error contains output. This leads to false positives where users believe they are asserting that "stderr is empty", when in reality no check is being executed.

**How To Reproduce**
Create a goss.yaml file with a command that outputs text to stderr alongside stderr: []:command:
  vacuous_stderr_test:
    exec: "sh -c 'echo \"something went wrong\" >&2'"
    exit-status: 0
    stderr: []
Run goss validate:goss validate

**Expected Behavior**
Either:`stderr: []` should fail if any output is received on standard error. Goss should emit a syntax error/validation error indicating that an empty matcher array [] performs no assertions (directing users to use stderr: "" instead).

**Actual Behavior**
The check passes successfully with zero failures:.
Total Duration: 0.002s
Count: 1, Failed: 0, Skipped: 0

***Workaround & Context***
In Goss v0.4.2+, checking for empty output is performed using a string matcher:stderr: ""
Or using a regex negation matcher:stderr:
  - "!/./"
However, stderr: [] currently passes silently without validating anything, creating a potential pitfall for users writing YAML tests.

**Environment:**
Version: v0.4.4 (or latest master)
OS: Linux / macOS

vsolano9:
Confirmed the behaviour on `master` at `056389c`, and the cause is `isSet` in `resource/resource.go`:

```go
case []interface{}:
    return len(v) > 0
```

A zero-length list makes the property look absent, so `ValidateValue` is never called and no check is created at all — which is why the run reports `Count: 1` rather than a passing `Count: 2`.

One thing worth flagging before anyone implements the suggested syntax error: **an empty matcher list is generated by goss itself.**

```console
$ goss add file /etc/hosts
file:
    /etc/hosts:
        exists: true
        ...
        contents: []
```

[...]

PR #1096, opened 2026-08-05
docs: explain that an empty matcher list asserts nothing

vsolano9:
Refs #1095

## Why docs rather than a behaviour change

#1095 is a real trap and the report is accurate: `stderr: []` looks like "assert stderr is empty" and asserts nothing. `isSet` in `resource/resource.go` returns false for a zero-length list, so the property is treated as absent and no check is ever created — the run reports `Count: 1` where you would expect 2.

The issue proposes making it an error instead. I started down that path and stopped, because an empty matcher list is **generated by goss itself** and is load-bearing in this repo:

```console
$ goss add file /etc/hosts
file:
    /etc/hosts:
        exists: true
        ...
        contents: []
```

`resource/file.go:125` sets `Contents: []string{}` for exactly this reason, and the generated form is checked in as expected output at `integration-tests/goss/rockylinux9/goss-expected-q.yaml`. `stderr: []` and friends also appear in `testdata/passing.goss.yaml`, `testdata/failing.goss.yaml`, `integration-tests/goss/goss-shared.yaml`, `goss-dummy.yaml`, and `goss-serve.yaml`.

So rejecting `[]` would make `goss add file` emit an immediately-invalid gossfile, break goss's own fixtures and integration tests, and break every user config produced by `goss add`. `[]` is not a mistake in this design — it is how "no expectation recorded" is written.

Worth noting that `goss add command` already writes `stderr: ""` when a command produced no error output, which is the assert-empty form. The two generators disagree, and that inconsistency is most of why the distinction is easy to miss.

That leaves the actual gap: the docs never say any of this. The `command` example uses `stderr: []` with no explanation, and neither `stdout`/`stderr` nor `contents` mentions that `""` and `[]` mean different things.

## What this adds

Two admonitions in `docs/gossfile.md`, one under `command` and one under `file`, stating that an empty list is zero conditions and always passes, and pointing at `""` for asserting emptiness. Both use the existing `!!! warning` style already used elsewhere in that file.

## Verified

Everything asserted above was run locally against `master` at `056389c`:

| Config | Result |
| --- | --- |
| `stderr: []` against a command writing to stderr | `Count: 1, Failed: 0` — passes, no check created |
| `stderr: ""` against the same command | `Count: 2, Failed: 1` — fails correctly |
| `contents: []` against a non-empty file | `Count: 1, Failed: 0` — passes |
| `contents: ""` against the same file | `Count: 2, Failed: 1` — fails correctly |

`npx markdownlint-cli2 "docs/**/*.md"` reports 0 issues in 14 files, matching the `DavidAnson/markdownlint-cli2-action` step in `.github/workflows/docs.yaml`.

## If you would rather change the behaviour

Happy to follow up if you want code as well. The least disruptive option I can see is a warning rather than an error, but it would fire on every `goss add`-generated file until the generator changed too, so it needs a decision about `goss add file` first. Your call — I did not want to guess at that in a first PR.

## AI assistance

This contribution was prepared with AI assistance. Every command and result quoted above was run locally.

----
📚 Documentation preview 📚: https://goss--1096.org.readthedocs.build/en/1096/

### DIFF
--- a/docs/gossfile.md
+++ b/docs/gossfile.md
@@ -173,6 +173,20 @@ command:
 
 `stdout` and `stderr` can be a string or [pattern](#patterns)
 
+!!! warning "An empty list asserts nothing"
+
+    A list of patterns is a list of conditions that must all be satisfied. An
+    empty list is therefore zero conditions, and always passes:
+
+    ```yaml
+    stderr: []   # asserts nothing; passes even when stderr has output
+    stderr: ""   # asserts stderr is empty
+    ```
+
+    This is why `goss add` writes `stderr: ""` rather than `stderr: []` when a
+    command produced no error output. Use `[]` only where you deliberately want
+    no assertion at all.
+
 The `exec` attribute is the command to run; this defaults to the name of
 the hash for backwards compatibility
 
@@ -277,6 +291,14 @@ file:
 ```
 
 `contents` can be a string or a [pattern](#patterns)
+
+!!! warning "An empty list asserts nothing"
+
+    As with [`stdout` and `stderr`](#command), `contents: []` is a list of zero
+    conditions and always passes. `goss add file` writes it to record that it
+    captured no content expectation, so it is normal in generated gossfiles —
+    but it does not check that the file is empty. To assert emptiness, use
+    `contents: ""`.
 
 ### gossfile
 


kgaughan (approved):
I'm fine with the documentation update rather than changing the behaviour. If somebody wants to assert that the file is actually empty, there's a mechanism there for them. Thanks!

PR #1099, opened 2026-08-07
fix: stop AndMatcher panicking when no child matcher ran (#982)

### LINKED ISSUE #982: Nil pointer dereference when gjson nested key itself doesn't match
**Describe the bug**
Using gjson matching, if the key does not exist in content, goss panics.

**How To Reproduce**
```yaml
# repro.yaml
matching:
  example:
    content: '{"this": {"is": {"just": {"a": "test"}}}}'
    matches:
      gjson:
        this.is.typo:
          and:
            - {have-key: "a"}
```
```
$ goss -g repro.yaml validate

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x84e53b]

goroutine 23 [running]:
github.com/goss-org/goss/matchers.(*AndMatcher).FailureResult(0x4119db?, {0x0?, 0x0?})
	/home/travis/gopath/src/github.com/goss-org/goss/matchers/and.go:32 +0x5b
github.com/goss-org/goss/matchers.(*WithSafeTransformMatcher).FailureResult(0xb567f8?, {0xa676c0, 0xc000330c30})
	/home/travis/gopath/src/github.com/goss-org/goss/matchers/with_safe_transform.go:45 +0x1bc
github.com/goss-org/goss/matchers.(*AndMatcher).FailureResult(0xc0000a4630?, {0xa676c0?, 0xc000330c30?})
	/home/travis/gopath/src/github.com/goss-org/goss/matchers/and.go:32 +0x62
github.com/goss-org/goss/resource.ValidateGomegaValue({_, _}, {_, _}, {_, _}, {_, _}, _)
	/home/travis/gopath/src/github.com/goss-org/goss/resource/validate.go:199 +0x788
github.com/goss-org/goss/resource.ValidateValue({_, _}, {_, _}, {_, _}, {_, _}, _)
	/home/travis/gopath/src/github.com/goss-org/goss/resource/validate.go:126 +0x174
github.com/goss-org/goss/resource.(*Matching).Validate(0xc000320c00, 0x0?)
	/home/travis/gopath/src/github.com/goss-org/goss/resource/matching.go:62 +0x15a
github.com/goss-org/goss.validate.func2()
	/home/travis/gopath/src/github.com/goss-org/goss/validate.go:184 +0x86
created by github.com/goss-org/goss.validate in goroutine 1
	/home/travis/gopath/src/github.com/goss-org/goss/validate.go:181 +0x187
```

**Expected Behavior**
To get a regular goss `F` failure with details on the mismatched key.

**Actual Behavior**
```
panic: runtime error: invalid memory address or nil pointer dereference
```

**Environment:**
 - goss v0.4.8
 - Ubuntu 22.04.5

vsolano9:
Closes #982.

## Reproduced on current master

The gossfile from the issue, unchanged, against `master` @ `4a91642`:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x20]

matchers.(*AndMatcher).FailureResult          matchers/and.go:32
matchers.(*WithSafeTransformMatcher).FailureResult  matchers/with_safe_transform.go:45
matchers.(*AndMatcher).FailureResult          matchers/and.go:32
resource.ValidateGomegaValue                  resource/validate.go:199
```

Same frames as the 2024 report, 23 months on.

## Cause

`WithSafeTransformMatcher.Match` returns as soon as its transform errors:

```go
m.transformedValue, err = m.Transform.Transform(actual)
...
if err != nil {
    return false, fmt.Errorf("%#v: %w", m.Transform, err)
}
return m.Matcher.Match(m.transformedValue)   // never reached
```

`Gjson.Transform` errors with `Path not found: …` when the path is absent, so the wrapped
`And` never evaluates a child and `firstFailedMatcher` stays `nil`.
`resource/validate.go:199` then branches only on `success`, not on `err`, and calls
`FailureResult` anyway — into the nil.

Your read in the thread was right: it only panics with `and`. I checked — `or`, `not` and a
bare matcher were each run against the same gossfile and none of them panics. `OrMatcher`
survives because its `FailureResult` doesn't touch `firstSuccessfulMatcher` at all.

## Fix

A nil guard in `AndMatcher.FailureResult`, reporting the `And` itself when no child ran.

**I chose this over the other obvious fix, and the output is why.** The alternative is to make
`validate.go` skip `FailureResult` when `Match` returned an error — arguably more correct,
since a matcher that errored has undefined state, and it would mirror the existing early
return a few lines above. I implemented both and compared what the user actually sees:

*validate.go variant:*
```
Matching: example: matches:
Error
    matchers.Gjson{Path:"this.is.typo"}: Path not found: this.is.typo
```

*this PR:*
```
Matching: example: matches:
Error
    matchers.Gjson{Path:"this.is.typo"}: Path not found: this.is.typo
the transform chain was
    [{"gjson":{"Path":"this.is.typo"}}]
the raw value was
    "{\"this\": {\"is\": {\"just\": {\"a\": \"test\"}}}}"
```

Keeping `FailureResult` in the chain preserves the transform chain and the raw value — which
is exactly what makes a typo'd gjson path obvious, since you can see the document you were
matching against. Happy to switch to the `validate.go` version if you'd rather have the
stricter semantics; it's a two-line change and I have it ready.

## Tests

Two, both in `matchers/and_test.go`:

- the panic case — an erroring transform wrapping an `And`, asserting `FailureResult` doesn't
  panic and reports the And;
- the delegation case — an `And` whose child genuinely failed must still report *that child*,
  not the And's own message. Without this, a guard that always fired would look correct.

**Mutation-tested:** removing the guard fails the first test with the original panic; making
the guard always fire fails the second. Each test fails for its own reason and nothing else
moves.

`go test -race ./...` passes, `gofmt -l` clean, `go vet` clean, and `golangci-lint run` (v2.12.2,
the full `$(pkgs)` list) reports **0 issues** on this branch.

*Edited: this originally said I didn't have `golangci-lint` locally. @kgaughan pointed out that
`make lint` installs it — he was right, and the real reason I didn't have a result is in
[the comment below](https://github.com/goss-org/goss/pull/1099#issuecomment-5225391505).*

### DIFF
--- a/matchers/and.go
+++ b/matchers/and.go
@@ -28,7 +28,22 @@ func (m *AndMatcher) Match(actual interface{}) (success bool, err error) {
 	return true, nil
 }
 
+// FailureResult reports the first matcher that failed. firstFailedMatcher is
+// only set once Match has actually run a child matcher, and that does not
+// always happen: WithSafeTransformMatcher.Match returns early when its
+// transform errors — a gjson path that does not exist, for example — so the
+// wrapped And never evaluates anything and its state stays nil. Dereferencing
+// it there panicked (#982). Report the And itself in that case, which keeps the
+// surrounding transform chain and raw value in the output so the user can see
+// which path they actually matched against.
 func (m *AndMatcher) FailureResult(actual interface{}) MatcherResult {
+	if m.firstFailedMatcher == nil {
+		return MatcherResult{
+			Actual:   actual,
+			Message:  "to satisfy all of these matchers",
+			Expected: m.Matchers,
+		}
+	}
 	return m.firstFailedMatcher.FailureResult(actual)
 }
 
--- a/matchers/and.go
+++ b/matchers/and.go
@@ -36,7 +36,7 @@ func (m *AndMatcher) Match(actual interface{}) (success bool, err error) {
 // it there panicked (#982). Report the And itself in that case, which keeps the
 // surrounding transform chain and raw value in the output so the user can see
 // which path they actually matched against.
-func (m *AndMatcher) FailureResult(actual interface{}) MatcherResult {
+func (m *AndMatcher) FailureResult(actual any) MatcherResult {
 	if m.firstFailedMatcher == nil {
 		return MatcherResult{
 			Actual:   actual,
--- a/matchers/and_test.go
+++ b/matchers/and_test.go
@@ -1,0 +1,49 @@
+package matchers
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+// erroringTransform fails the way Gjson does when a path is absent.
+type erroringTransform struct{}
+
+func (erroringTransform) Transform(in interface{}) (interface{}, error) {
+	return in, errors.New("Path not found: this.is.typo")
+}
+
+func (erroringTransform) TransformerType() string { return "erroringTransform" }
+
+// A transform that errors makes WithSafeTransformMatcher.Match return before it
+// ever calls the wrapped matcher, so the And never runs a child and its
+// firstFailedMatcher stays nil. Asking it for a FailureResult used to panic
+// (#982).
+func TestAndMatcherFailureResultWhenNoChildMatcherRan(t *testing.T) {
+	inner := And(HaveKey("a"))
+	matcher := WithSafeTransform(erroringTransform{}, inner)
+
+	success, err := matcher.Match(`{"this": {"is": {"just": {"a": "test"}}}}`)
+	assert.False(t, success)
+	assert.Error(t, err)
+
+	assert.NotPanics(t, func() {
+		result := matcher.FailureResult("actual")
+		assert.Equal(t, "to satisfy all of these matchers", result.Message)
+	})
+}
+
+// The nil result must not be returned once a child has genuinely failed —
+// otherwise the guard would swallow the specific matcher that failed.
+func TestAndMatcherFailureResultStillReportsTheFailedChild(t *testing.T) {
+	m := And(HaveKey("missing"))
+
+	success, err := m.Match(map[string]interface{}{"present": 1})
+	assert.False(t, success)
+	assert.NoError(t, err)
+
+	result := m.FailureResult(map[string]interface{}{"present": 1})
+	assert.Equal(t, "to have key matching", result.Message,
+		"must delegate to the child that failed, not the And's own message")
+}
--- a/matchers/and_test.go
+++ b/matchers/and_test.go
@@ -10,7 +10,7 @@ import (
 // erroringTransform fails the way Gjson does when a path is absent.
 type erroringTransform struct{}
 
-func (erroringTransform) Transform(in interface{}) (interface{}, error) {
+func (erroringTransform) Transform(in any) (any, error) {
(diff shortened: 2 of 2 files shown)


matchers/and_test.go
```diff
@@ -0,0 +1,49 @@
...
+	"errors"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+// erroringTransform fails the way Gjson does when a path is absent.
+type erroringTransform struct{}
+
+func (erroringTransform) Transform(in interface{}) (interface{}, error) {
```
kgaughan:
```suggestion
func (erroringTransform) Transform(in any) (any, error) {
```

matchers/and_test.go
```diff
@@ -0,0 +1,49 @@
...
+		assert.Equal(t, "to satisfy all of these matchers", result.Message)
+	})
+}
+
+// The nil result must not be returned once a child has genuinely failed —
+// otherwise the guard would swallow the specific matcher that failed.
+func TestAndMatcherFailureResultStillReportsTheFailedChild(t *testing.T) {
+	m := And(HaveKey("missing"))
+
+	success, err := m.Match(map[string]interface{}{"present": 1})
```
kgaughan:
```suggestion
	success, err := m.Match(map[string]any{"present": 1})
```

matchers/and_test.go
```diff
@@ -0,0 +1,49 @@
...
+	assert.False(t, success)
+	assert.NoError(t, err)
+
+	result := m.FailureResult(map[string]interface{}{"present": 1})
```
kgaughan:
```suggestion
	result := m.FailureResult(map[string]any{"present": 1})
```

kgaughan (commented):
A new nits: prefer `any` to `interface{}`.

matchers/and_test.go
vsolano9: Updated to `any` in 55e3890.

matchers/and_test.go
vsolano9: Updated to `map[string]any` in 55e3890.

matchers/and_test.go
vsolano9: Updated to `map[string]any` in 55e3890.

matchers/and.go
```diff
@@ -28,7 +28,22 @@ func (m *AndMatcher) Match(actual interface{}) (success bool, err error) {
...
 
+// FailureResult reports the first matcher that failed. firstFailedMatcher is
+// only set once Match has actually run a child matcher, and that does not
+// always happen: WithSafeTransformMatcher.Match returns early when its
+// transform errors — a gjson path that does not exist, for example — so the
+// wrapped And never evaluates anything and its state stays nil. Dereferencing
+// it there panicked (#982). Report the And itself in that case, which keeps the
+// surrounding transform chain and raw value in the output so the user can see
+// which path they actually matched against.
 func (m *AndMatcher) FailureResult(actual interface{}) MatcherResult {
```
kgaughan:
```suggestion
func (m *AndMatcher) FailureResult(actual any) MatcherResult {
```

kgaughan (approved):

PR #1106, opened 2026-08-09
feat(kgoss): add GOSS_KUBECTL_TIMEOUT to allow slow pod startups

venc0r:
##### Checklist

### Description of change
When running many builds in parallel on a small machine, pod starts can be too slow. Currently the timeout had a fixed value of 60s. I moved the setting to a new env var GOSS_KUBECTL_TIMEOUT with the same default, to make it configurable. 

----
📚 Documentation preview 📚: https://goss--1106.org.readthedocs.build/en/1106/


kgaughan (approved):
Looks reasonable! I'll merge as soon as the test suite has run successfully.

PR #1107, opened 2026-08-11
test: cover SSHFP in the integration suite

### LINKED ISSUE #1104: Add SSHFP to the integration tests
PR #1033 added SSHFP record support, but there is a lack of checks in the integration tests.

Add integration tests based on the documentation provided in that PR.

vsolano9:
Issue #1104 records that SSHFP support merged without integration coverage. This adds the documented SSHFP lookup to the shared Linux integration configuration and updates the expected assertion totals for the five-distro matrix.

The test exercises the MCP-independent, end-to-end path against all six published SSHFP records: YAML parsing, typed DNS dispatch, UDP lookup through the explicit resolver, fingerprint normalization, and address matching.

##### Checklist

- [x] unit and/or integration tests are included

### Validation

- `go test ./...`
- `go vet ./...`
- `bash -n integration-tests/test.sh`
- `git diff --check`
- focused live validation: 2 assertions passed against the documented SSHFP records through `8.8.8.8`
- negative control: replacing one expected fingerprint produced 1 failed assertion

Docker is not available in the local environment, so the five-distro container matrix remains for CI.

Fixes #1104.


integration-tests/goss/goss-shared.yaml
```diff
@@ -144,6 +144,17 @@ dns:
     - 10 10 443 b.dnstest.io.
     timeout: 2000
     server: 127.0.0.1:8053
+  SSHFP:mars.yellowjacket.io:
```
kgaughan: While this does technically work, it reintroduces some of the brittleness that #1063 fixed by using dnsmasq running on 127.0.0.1:8053 to serve the records for `dnstest.io` locally using a zone described by `integration-tests/dnsmasq.conf`. If these records change or if the domain `mars.yellowjacket.io` goes away, it'll break the integration tests.

While dnsmasq doesn't have direct support for SSHFP records, it does have indirect support via the `dns-rr` setting, allowing records to be given in the following fashion based on [RFC 3597](https://www.rfc-editor.org/info/rfc3597/):
```
dns-rr==<name>,<RR-number>,[<hex data>]
```
Thus one such record could be specified as follows (with the zone switched from yellowjacket.io to dnstest.io for consistency):
```
dns-rr=mars.dnstest.io,44,1:2:422F22EFB548FEAC403A633A86F74553599B85AC93E7EC3BB0A46B6CD6DDABF8
```

vsolano9: Implemented in b086738: `integration-tests/dnsmasq.conf` now serves the SSHFP RDATA through the existing local `dnstest.io` zone, and the assertion queries `mars.dnstest.io` at `127.0.0.1:8053` with one deterministic record. `go test ./...`, `bash -n integration-tests/test.sh`, and the diff checks pass locally; the container integration matrix is running on the pushed commit.

vsolano9: The first push exposed one more integration-harness detail: Linux jobs pull prebuilt images, so the repository config was not reaching dnsmasq. Follow-up 5f97b73 bind-mounts the checked-out `dnsmasq.conf` read-only at `/etc/dnsmasq.conf`, which makes the local-zone fixture effective and keeps future DNS fixture edits testable. Fresh upstream postflight is 12/12 checks green, including all five Linux container images.

kgaughan: That's definitely an improvement. I really ought to have done something similar when I added dnsmasq to to fix up the DNS tests. Thanks for this, and sorry for not getting to it sooner. I've been a bit busy lately.

kgaughan (approved):

PR #1109, opened 2026-08-14
fix(prometheus): return a failure exit code, and seed the run outcome once

### LINKED ISSUE #992: Serve with Prometheus Output does not 503 on failed check
Unlike the other output modes (ex: documentation/json), the prometheus output does not report 503 at `/healthz` when a test fails.

Is this by design? If so could something be added to the docs to say this is the exception.

Thanks in advance!

For example using this `goss.yaml`
```
file:
  /etc/nothinghere:
    exists: true
```

Run this command `goss-darwin-amd64 -g goss.yaml serve -f json --format-options verbose`

Then curl localhost:8080/healthz and examine the status code. Other format options return 503, while prometheus option returns 200.

VXNCXNX:
`Prometheus.Output` always returned 0, so `serve -f prometheus` never answered 503 on `/healthz` and `validate -f prometheus` never failed a pipeline:

```
documentation  exit=1
json           exit=1
prometheus     exit=0    <- failing spec
```

This is what @petemounce proposed in #992: L87 should return 1 when there are failure outcomes.

Closes #992

## The second half

Fixing only the exit code would have made the 503 fire at random.

The seed check was `i == 0 || canChangeOverallOutcome(...)`, where `i` is the index **within a result group**. goss delivers results as several groups in a nondeterministic order, so any group whose first entry passed reset a previously recorded failure. Same spec, six consecutive runs, before this change:

```
goss_tests_run_outcomes_total{outcome="fail"} 1
goss_tests_run_outcomes_total{outcome="pass"} 1
goss_tests_run_outcomes_total{outcome="pass"} 1
goss_tests_run_outcomes_total{outcome="fail"} 1
goss_tests_run_outcomes_total{outcome="fail"} 1
goss_tests_run_outcomes_total{outcome="fail"} 1
```

So the reported run outcome was already wrong roughly a third of the time, independently of the exit code. Seeding once across all groups rather than once per group fixes it — six runs after the change are all `fail`, and an all-passing spec still reports `pass` with exit 0.

I kept `canChangeOverallOutcome` unchanged. Dropping the seed entirely and relying on its `unknown` branch looks tempting but is wrong: `unknown` is also a real test result, and `unknown-success` must stay `unknown` rather than being promoted to `pass`.

## Behaviour change worth noting in release notes

`validate -f prometheus` now exits 1 on failure where it exited 0 before. That is the point of the issue, but it is user-visible for anyone using that format in CI.

`/metrics` is unaffected — it is served by a separate `promhttp.Handler()` on the default registry and shares nothing with this outputer's private registry. The 503 also still writes the full metrics body, since `serve.go` writes the body unconditionally.

## Tests

`go test ./...` passes across all six packages.

`TestPrometheusOutput` hardcoded `assert.Equal(t, 0, exitCode)` for every case, which is exactly what caught this change. The struct now carries `expectedExitCode`, set to 1 on the 10 cases whose expected run outcome is `fail`, and left 0 elsewhere.

### DIFF
--- a/outputs/prometheus.go
+++ b/outputs/prometheus.go
@@ -46,9 +46,10 @@ func (r Prometheus) Output(w io.Writer, results <-chan []resource.TestResult,
 	}
 
 	overallOutcome := resource.OutcomeUnknown
+	first := true
 	var startTime time.Time
 	for resultGroup := range results {
-		for i, tr := range resultGroup {
+		for _, tr := range resultGroup {
 			if startTime.IsZero() || tr.StartTime.Before(startTime) {
 				startTime = tr.StartTime
 			}
@@ -62,8 +63,11 @@ func (r Prometheus) Output(w io.Writer, results <-chan []resource.TestResult,
 				testOutcomes.WithLabelValues(resType, outcome).Inc()
 				testDurations.WithLabelValues(resType, outcome).Add(float64(tr.Duration.Milliseconds()))
 			}
-			if i == 0 || canChangeOverallOutcome(overallOutcome, outcome) {
+			// seed once across all groups, not once per group: results arrive
+			// in several groups in a nondeterministic order
+			if first || canChangeOverallOutcome(overallOutcome, outcome) {
 				overallOutcome = outcome
+				first = false
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


kgaughan (approved):
Nice! Everything looks good.

PR #1110, opened 2026-08-15
fix(serve): dedupe concurrent validation runs on a cold cache

### LINKED ISSUE #1094: goss serve runs one full validation per concurrent request when the cache expires
**Describe the bug**

When the `goss serve` result cache is cold, every request that arrives runs its
own complete validation. There is no deduplication, so N simultaneous requests
do N times the work.

This hits the case `goss serve` exists for. Anything scraping `/healthz` on an
interval, several Prometheus replicas, a load balancer health check, a
Kubernetes probe, will converge on the cache expiry and arrive in the cold
window together. The default `--cache` is 5s, so that window comes round often.

`healthHandler` has a `gossMu *sync.Mutex` field (`serve.go:57` and
`serve.go:73`) that looks like it was meant for this, but it is never locked
anywhere in the package. So there is not even serialisation, let alone
deduplication. The runs happen genuinely concurrently, each building its own
`system.New` and running every check in the gossfile.

**How To Reproduce**

Save as `stampede_test.go` in the root package and run
`go test -run TestStampede -v ./`. It counts the cache-miss line, which is
logged once per validation run.

```go
package goss

import (
	"bytes"
	"log"
	"net/http"
	"net/http/httptest"
	"path/filepath"
	"strings"
	"sync"
	"testing"

	"github.com/goss-org/goss/util"
	"github.com/stretchr/testify/require"
)

type syncBuf struct {
	mu  sync.Mutex
	buf bytes.Buffer
}

func (s *syncBuf) Write(p []byte) (int, error) {
	s.mu.Lock()
	defer s.mu.Unlock()
	return s.buf.Write(p)
}

func (s *syncBuf) String() string {
	s.mu.Lock()
	defer s.mu.Unlock()
	return s.buf.String()
}

func TestStampede(t *testing.T) {
	config, err := util.NewConfig(
		util.WithSpecFile(filepath.Join("testdata", "matching_basic.yaml")),
		util.WithOutputFormat("json"),
	)
	require.NoError(t, err)

	handler, err := newHealthHandler(config)
	require.NoError(t, err)

	var out syncBuf
	log.SetOutput(&out)
	log.SetFlags(0)

	const n = 16
	var wg sync.WaitGroup
	wg.Add(n)
	for range n {
		go func() {
			defer wg.Done()
			rr := httptest.NewRecorder()
			handler.ServeHTTP(rr, httptest.NewRequest(http.MethodGet, "/healthz", nil))
		}()
	}
	wg.Wait()

	t.Logf("%d concurrent requests produced %d validation runs",
		n, strings.Count(out.String(), "running tests"))
}
```

Without Go: start `goss serve`, fire 16 parallel curls at `/healthz`, and count
the `Stale cache[res], running tests` lines. They appear at the default log
level.

**Expected Behavior**

One validation run per cold-cache window. Requests that arrive while a
validation is already in flight should wait for it and share the result.

**Actual Behavior**

```
16 concurrent requests produced 16 validation runs
```

One run per request. With a real gossfile that multiplier applies to every
check in it, including any `command` resources, which get executed 16 times
against the machine instead of once.

**Environment:**

* Version of goss: master at 5fc15e4, and v0.4.10 has the identical code path
* OS/Distribution version: reproduced on darwin/arm64, though the code path is
  platform independent

---

**Possible fix**

`golang.org/x/sync/singleflight` around the cache-miss branch of
`processAndEnsureCached` covers this in a few lines. The first caller runs the
validation, the rest block on it and receive the same result. A second cache
check inside the shared function handles the case where a run has just
populated it.

We have been running this on an internal fork for a couple of months and it
behaves as expected. Happy to open a PR if the approach sounds reasonable. The
unused `gossMu` field could go at the same time.

VXNCXNX:
##### Checklist

- [x] `make test-all` (UNIX) passes. CI will also test this (the Makefile has no `test-all` target; ran `make test-short-all`, which is `fmt lint vet test`, all green, plus `golangci-lint run` explicitly since the make step could not find the binary on PATH: `0 issues.`)
- [ ] documentation is changed or added (if applicable) (no docs change needed, this fixes internal serialisation and does not alter documented behaviour)

### Description of change

Fixes #1094.

## What's broken

When the result cache is cold, every concurrent request runs its own full validation. With the reporter's test on master:

```
16 concurrent requests produced 10 validation runs
```

For a gossfile with `command` resources, that means every command executes ten times against the machine for one cold window. The reporter saw 16 of 16 on darwin, so the multiplier is timing dependent, but the shape is the same.

## The cause

The cache-miss branch of `processAndEnsureCached` in `serve.go` has no serialisation. N requests arriving before the first result is stored each build their own `system.New` and run every check.

The telling detail: `gossMu` is already on the handler, constructed at `serve.go:63` and declared at `:79`. `grep` finds zero `.Lock()` calls on it anywhere in the repo. The intent was there, the lock was never taken.

## The fix

Lock around the miss branch and re-check the cache inside the lock. The second `Get` is what turns serialisation into deduplication: without it the waiters would queue up and each still run.

No new dependency, and it uses the field that was already there.

## The trade-off, stated plainly

Waiters now block on the leader instead of each starting their own run, so during a cold window p99 latency becomes one validation duration rather than N concurrent ones. That is the intended exchange and it is strictly less load on the machine, but it is a behaviour change worth naming.

## Verification

```
before:  16 concurrent requests caused 3 validation runs
after:   16 concurrent requests caused 1 validation run
```

The new test lives beside `TestServeHandlesConcurrentRequests` and counts the cache-miss log line, reusing the `syncBuffer` helper the serve tests already use for this. Reverting the source with the test kept makes it fail with the line above.

`go build ./...`, `go vet ./...`, `gofmt -l .` and `go test -race -count=1 ./...` are all clean. The existing `TestServeHandlesConcurrentRequests` still passes unchanged; it was passing with the stampede present, so it needed checking rather than amending.

## One note on ownership

@dukelion offered in the issue to open a PR for this and said "Happy to open a PR if the approach sounds reasonable". That was twelve days ago with no reply. If he would rather take it, I am happy to close this in favour of his.

*Disclosure: written with AI assistance (Claude Code). I reproduced the issue, ran the change and the verification myself.*

### DIFF
--- a/serve.go
+++ b/serve.go
@@ -104,23 +104,20 @@ func (h healthHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 func (h healthHandler) processAndEnsureCached(negotiatedContentType string, outputer outputs.Outputer) res {
 	var tra [][]resource.TestResult
 	cacheKey := "res"
-	tmp, found := h.cache.Get(cacheKey)
-	if found {
+	// Held across the lookup so a miss does not let a second request start its
+	// own validate() before the first one has filled the cache.
+	h.gossMu.Lock()
+	if tmp, found := h.cache.Get(cacheKey); found {
 		log.Printf("[TRACE] Returning cached[%s].", cacheKey)
 		tra = tmp.([][]resource.TestResult)
 	} else {
-		h.gossMu.Lock()
-		// another request may have filled the cache while we waited for the lock
-		if tmp, found := h.cache.Get(cacheKey); found {
-			tra = tmp.([][]resource.TestResult)
-		} else {
-			log.Printf(cacheMissLogFormat, cacheKey)
-			h.sys = system.New(h.c.PackageManager)
-			tra = h.validate()
-			h.cache.SetDefault(cacheKey, tra)
-		}
-		h.gossMu.Unlock()
+		log.Printf(cacheMissLogFormat, cacheKey)
+		h.sys = system.New(h.c.PackageManager)
+		tra = h.validate()
+		h.cache.SetDefault(cacheKey, tra)
 	}
+	h.gossMu.Unlock()
+
 	trc := testResultArrayToChan(tra)
 	return h.output(trc, outputer)
 }
--- a/serve.go
+++ b/serve.go
@@ -109,10 +109,17 @@ func (h healthHandler) processAndEnsureCached(negotiatedContentType string, outp
 		log.Printf("[TRACE] Returning cached[%s].", cacheKey)
 		tra = tmp.([][]resource.TestResult)
 	} else {
-		log.Printf(cacheMissLogFormat, cacheKey)
-		h.sys = system.New(h.c.PackageManager)
-		tra = h.validate()
-		h.cache.SetDefault(cacheKey, tra)
+		h.gossMu.Lock()
+		// another request may have filled the cache while we waited for the lock
+		if tmp, found := h.cache.Get(cacheKey); found {
+			tra = tmp.([][]resource.TestResult)
+		} else {
+			log.Printf(cacheMissLogFormat, cacheKey)
+			h.sys = system.New(h.c.PackageManager)
+			tra = h.validate()
+			h.cache.SetDefault(cacheKey, tra)
+		}
+		h.gossMu.Unlock()
 	}
 	trc := testResultArrayToChan(tra)
 	return h.output(trc, outputer)
--- a/serve_concurrency_test.go
+++ b/serve_concurrency_test.go
@@ -1,9 +1,12 @@
 package goss
 
 import (
+	"log"
 	"net/http"
 	"net/http/httptest"
+	"os"
 	"path/filepath"
+	"strings"
 	"sync"
 	"testing"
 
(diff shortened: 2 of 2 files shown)


serve.go
```diff
@@ -109,10 +109,17 @@ func (h healthHandler) processAndEnsureCached(negotiatedContentType string, outp
 		log.Printf("[TRACE] Returning cached[%s].", cacheKey)
 		tra = tmp.([][]resource.TestResult)
 	} else {
-		log.Printf(cacheMissLogFormat, cacheKey)
-		h.sys = system.New(h.c.PackageManager)
-		tra = h.validate()
-		h.cache.SetDefault(cacheKey, tra)
+		h.gossMu.Lock()
```
kgaughan: I get what you're trying to do here, but why not instead lock the mutex before line 107 and release it before calling `testResultArrayToChan` on line 117 instead? I know the speculative `Get` on the cache is technically safe as it's internally protected by a mutex itself, but I can't see a good reason to try it twice beyond maybe a smidgen of performance.

That said, this is a worthy fix. I've run into this cache miss issue previously at $WORK as it triggered some redundant load on a service during health checks that sometimes led the health check to fail (a weird locking issue in OpenLDAP when checking its replication status, if you're interested, and I can think of a few other situations too that I can't mention).

VXNCXNX: Agreed, that is simpler and the speculative Get was not buying anything real. One lock around the whole lookup now, pushed.

The dedup test still catches a revert: without the lock the 16 concurrent requests produce 16 validation runs instead of 1.

kgaughan (approved):

PR #1115, opened 2026-08-15
fix(http): stop a request-header without a space after the colon panicking

VXNCXNX:
##### Checklist

- [x] `make test-all` (UNIX) passes. CI will also test this (the Makefile has no `test-all` target; ran `make test-short-all`, which is `fmt lint vet test`, all green, plus `golangci-lint run` explicitly since the make step could not find the binary on PATH: `0 issues.`)
- [ ] documentation is changed or added (if applicable) (the docs only show the well-formed `Key: value` spelling, which still behaves exactly as before, so nothing documented changes)

### Description of change

No matching issue is filed for this one; I found it while writing a request-header in the HTTP wire form.

## What's broken

A `request-headers` entry written in the HTTP wire form, with no space after the colon, panics the whole run.

```
http:
  http://127.0.0.1:9/:
    status: 200
    request-headers:
      - "X-Foo:bar"
```

```
$ goss validate
panic: runtime error: index out of range [1] with length 1
	github.com/goss-org/goss/system.NewDefHTTP(...) system/http.go:61
	github.com/goss-org/goss/resource.(*HTTP).Validate(...) resource/http.go:71
```

An entry with no colon at all panics identically. `"X-Foo:  bar"`, with two spaces, happens to work.

## The fix

```go
str := strings.SplitN(r, ": ", 2)
headers.Add(str[0], str[1])
```

The separator is the two-character string `": "`, so anything not matching that exact shape returns a one-element slice and `str[1]` is out of range. Nothing validates the entry beforehand, and the docs only show the well-formed spelling, so writing a header the way HTTP itself writes it gets you a stack trace.

It now cuts on the first `:` and trims both halves, which accepts `X-Foo:bar` and leaves `X-Foo: bar` behaving exactly as before. That matches `hasUserAgentHeader` a few lines up, which already prefix-matches on `"user-agent:"` without requiring the space.

An entry with no colon cannot become a header pair at all, so it is skipped with a warning on stderr rather than indexed. `NewDefHTTP` returns no error and its callers consume the value directly, so returning one would mean changing the `System.NewHTTP` func type and every call site. Warning and continuing is what the codebase already does for other unusable input, in `resource/resource.go`.

## Verification

Table-driven cases in `system/http_test.go` for the colon-space form, the no-space form and the no-colon form, asserting the resulting header values. With the change reverted the no-space case panics with the message above. `go test ./...` passes. `integration-tests/goss/goss-shared.yaml` uses `"Foo: bar"`, which is unaffected.

*Disclosure: written with AI assistance (Claude Code). I reproduced the issue, ran the change and the verification myself.*

### DIFF
--- a/system/http.go
+++ b/system/http.go
@@ -6,6 +6,7 @@ import (
 	"crypto/x509"
 	"fmt"
 	"io"
+	"log"
 	"net/http"
 	"net/url"
 	"os"
@@ -59,7 +60,7 @@ func NewDefHTTP(_ context.Context, httpStr string, system *System, config util.C
 	for _, r := range config.RequestHeader {
 		name, value, found := strings.Cut(r, ":")
 		if !found {
-			fmt.Fprintf(os.Stderr, "WARNING: ignoring malformed request header %q, expected \"Name: value\"\n", r)
+			log.Printf("[WARNING] ignoring malformed request header %q, expected \"Name: value\"", r)
 			continue
 		}
 		headers.Add(strings.TrimSpace(name), strings.TrimSpace(value))
--- a/system/http.go
+++ b/system/http.go
@@ -57,8 +57,12 @@ func NewDefHTTP(_ context.Context, httpStr string, system *System, config util.C
 	}
 
 	for _, r := range config.RequestHeader {
-		str := strings.SplitN(r, ": ", 2)
-		headers.Add(str[0], str[1])
+		name, value, found := strings.Cut(r, ":")
+		if !found {
+			fmt.Fprintf(os.Stderr, "WARNING: ignoring malformed request header %q, expected \"Name: value\"\n", r)
+			continue
+		}
+		headers.Add(strings.TrimSpace(name), strings.TrimSpace(value))
 	}
 	return &DefHTTP{
 		http:              httpStr,
--- a/system/http_test.go
+++ b/system/http_test.go
@@ -1,7 +1,6 @@
 package system
 
 import (
-	"context"
 	"io"
 	"net/http"
 	"strings"
@@ -60,7 +59,7 @@ func TestNewDefHTTPRequestHeaderParsing(t *testing.T) {
 
 	for _, tc := range tests {
 		t.Run(tc.name, func(t *testing.T) {
-			h := NewDefHTTP(context.Background(), "http://example.com", nil, util.Config{
+			h := NewDefHTTP(t.Context(), "http://example.com", nil, util.Config{
 				RequestHeader: []string{tc.header},
 			})
 			def, ok := h.(*DefHTTP)
--- a/system/http_test.go
+++ b/system/http_test.go
@@ -1,10 +1,13 @@
 package system
 
 import (
+	"context"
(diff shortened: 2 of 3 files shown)


system/http_test.go
```diff
@@ -40,3 +43,33 @@ func TestDefHTTPCloseClosesBodyAfterBodyRead(t *testing.T) {
...
+		want   string
+	}{
+		{"colon and space", "X-Foo: bar", "bar"},
+		{"colon no space", "X-Foo:bar", "bar"},
+		{"no colon", "X-Foo", ""},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			h := NewDefHTTP(context.Background(), "http://example.com", nil, util.Config{
```
kgaughan:
```suggestion
			h := NewDefHTTP(c.Context(), "http://example.com", nil, util.Config{
```

system/http.go
```diff
@@ -57,8 +57,12 @@ func NewDefHTTP(_ context.Context, httpStr string, system *System, config util.C
 	}
 
 	for _, r := range config.RequestHeader {
-		str := strings.SplitN(r, ": ", 2)
-		headers.Add(str[0], str[1])
+		name, value, found := strings.Cut(r, ":")
+		if !found {
+			fmt.Fprintf(os.Stderr, "WARNING: ignoring malformed request header %q, expected \"Name: value\"\n", r)
```
kgaughan: Use the logging library for this:

```suggestion
			log.Printf("[WARNING] ignoring malformed request header %q, expected \"Name: value\"", r)
```

system/http_test.go
VXNCXNX: Applied, thanks. I took `t.Context()` on the subtest's `t`, which also let the `context` import go.

Worth flagging: the same `context.Background()` appears in `resource/file_test.go:17` from another of my PRs (#1113). That one is not in a subtest closure, so it is a straight swap too. Happy to change it there if you would rather the repo be consistent, or leave it if you are only steering new tests.

system/http.go
VXNCXNX: Done, with the `log` import added.

One thing to sanity check on your side: `log.Printf` writes to stderr by default, same as before, but goss configures the standard logger in `main.go` rather than here, so the warning now picks up whatever prefix and flags are in effect. That reads like the right behaviour to me, since it makes this line consistent with the `[INFO]`/`[DEBUG]` output in `serve.go`, but it does mean the text is no longer bare.

kgaughan: Yup, that's what's expected.

system/http_test.go
kgaughan: While it's a bit out of scope for this particular PR, yeah, that'd be good! I'd like to get this stuff handled consistently. Since the move to urfave/cli v3, I'd like to tighten up the handling of contexts in general (there are 50 instances of `context.Background()` in the code that probably shouldn't exist). Fixing this extra instance in the tests would definitely be a good idea.

kgaughan (approved):

Issue #947, opened 2024-07-10
New --exact-match option to "goss add command" to perform exact multi-line matches


gberche-orange:
**Describe the feature:**

Enable `goss add command` to capture stdout/stderr as a multiline script

This enables to capture & then validate various multiline outputs that are sensitive to space and linefeed (typically yaml output). 

This is useful in the test approach sometimes called golden master testing / approval testing/ Characterization tests
see https://en.wikipedia.org/wiki/Characterization_test

Here is a sample library that that supports such practice https://approvaltests.com/

Currently, the `goss add command` does a great job at capturing by default an array of trimmed strings to match. But this can not be used for approval testing commands that produce yaml.

**Describe the solution you'd like**

A new --exact-match option to "goss add command" to perform exact multi-line matches

**Describe alternatives you've considered**

wrap `goss add` with a script which would replace using yq the `command.cmd.stdout`

aelsabbahy:
Makes sense, marked as approved. I'll accept a PR implementing this. Can also be useful for the http check.

PR #1122, opened 2026-08-24
feat(add): --exact-match flag for the command generator

ChrisJr404:
Implements the approved feature request in #947.

`goss add command` records stdout/stderr as a list of trimmed lines, which validate then treats as patterns. That's great for most checks, but it drops whitespace, blank lines and ordering, so you can't use it for approval / golden master style tests where the whole output matters (the issue's example is yaml).

This adds a `--exact-match` flag to `goss add command`. When set, the generator keeps the raw output as a single string, which goss already validates as an exact, whitespace sensitive match via the string form of `stdout`/`stderr`. Nothing changes without the flag.

Before:

```console
$ goss add command "printf 'line one\n  indented two\n'"
```
```yaml
command:
  printf 'line one\n  indented two\n':
    exit-status: 0
    stdout:
      - line one
      - indented two
    stderr: ""
```

After, with `--exact-match`:

```yaml
command:
  printf 'line one\n  indented two\n':
    exit-status: 0
    stdout: |
      line one
        indented two
    stderr: ""
```

The second one fails validation if the indentation or a newline drifts, which is the point.

##### Checklist

### Description of change

CLI flag on the `command` add subcommand, threaded through `util.Config.ExactMatch` into `NewCommand`, which picks the raw string instead of the trimmed line slice. Added `resource/command_test.go` covering the default, exact-match, empty-output and `--exclude-attr` cases, and documented the flag in `docs/cli.md` and the command section of `docs/gossfile.md`.

----
📚 Documentation preview 📚: https://goss--1122.org.readthedocs.build/en/1122/

### DIFF
--- a/resource/command.go
+++ b/resource/command.go
@@ -103,20 +103,38 @@ func NewCommand(sysCommand system.Command, config util.Config) (*Command, error)
 
 	if !contains(config.IgnoreList, "stdout") {
 		stdout, _ := sysCommand.Stdout()
-		outSlice := readerToSlice(stdout)
-		if len(outSlice) != 0 {
-			c.Stdout = outSlice
+		if out := readerToMatcher(stdout, config.ExactMatch); out != nil {
+			c.Stdout = out
 		}
 	}
 	if !contains(config.IgnoreList, "stderr") {
 		stderr, _ := sysCommand.Stderr()
-		errSlice := readerToSlice(stderr)
-		if len(errSlice) != 0 {
-			c.Stderr = errSlice
+		if out := readerToMatcher(stderr, config.ExactMatch); out != nil {
+			c.Stderr = out
 		}
 	}
 
 	return c, err
+}
+
+// readerToMatcher turns captured command output into the matcher goss writes
+// for it. By default that's a list of trimmed lines, matched as patterns. With
+// exactMatch it's the raw output as a single string, which goss validates for
+// an exact (whitespace and newline sensitive) match. It returns nil when there
+// is nothing to assert so the caller can leave the default empty value in place.
+func readerToMatcher(reader io.Reader, exactMatch bool) matcher {
+	if exactMatch {
+		b, _ := io.ReadAll(reader)
+		if len(b) == 0 {
+			return nil
+		}
+		return string(b)
+	}
+	slice := readerToSlice(reader)
+	if len(slice) == 0 {
+		return nil
+	}
+	return slice
 }
 
 func escapePattern(s string) string {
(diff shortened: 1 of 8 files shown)


resource/command.go
```diff
@@ -103,22 +103,40 @@ func NewCommand(sysCommand system.Command, config util.Config) (*Command, error)
...
 }
 
+// readerToMatcher turns captured command output into the matcher goss writes
+// for it. By default that's a list of trimmed lines, matched as patterns. With
+// exactMatch it's the raw output as a single string, which goss validates for
+// an exact (whitespace and newline sensitive) match. It returns nil when there
+// is nothing to assert so the caller can leave the default empty value in place.
+func readerToMatcher(reader io.Reader, exactMatch bool) matcher {
+	if exactMatch {
+		b, _ := io.ReadAll(reader)
```
kgaughan: Is there are reason you're swallowing the error here?

kgaughan (commented):
The change looks good in principle, though I did spot a place where there's a potential error being swallowed.

It looks like the changes introduced in #871 conflict with yours.  I resolved them as well as I could, but I'd like you to double-check me.

resource/command_test.go
```diff
@@ -1,12 +1,84 @@
...
+	"strings"
 	"testing"
 
+	"github.com/goss-org/goss/util"
 	"go.yaml.in/yaml/v3"
 )
 
+func newFakeCommand(stdout, stderr string) *fakeCommand {
+	return &fakeCommand{
+		command:      "some-command",
```
kgaughan:
```suggestion
		command:      util.ExecCommand{CmdStr: "some-command"},
```

kgaughan (commented):
Correction due to changes to the `fakeCommand` type.

resource/command_test.go
```diff
@@ -1,12 +1,84 @@
...
+	if err != nil {
+		t.Fatalf("NewCommand returned error: %v", err)
+	}
+
+	if got, ok := c.Stdout.(string); !ok || got != "" {
+		t.Fatalf("Stdout = %#v, want empty string when ignored", c.Stdout)
+	}
+	if got, ok := c.Stderr.(string); !ok || got != "some error\n" {
+		t.Fatalf("Stderr = %#v, want captured raw string", c.Stderr)
+  }
```
kgaughan:
```suggestion
	}
```

kgaughan (commented):
Also, could you add something to the integration tests to demonstrate this in practice?

kgaughan (approved):
Ah! I thought I'd merged this already! Approved!

PR #1114, opened 2026-08-15
fix(validate): warn on the remaining empty list matchers

VXNCXNX:
##### Checklist

- [x] `make test-all` (UNIX) passes. CI will also test this (the Makefile has no `test-all` target; ran `make test-short-all`, which is `fmt lint vet test`, all green, plus `golangci-lint run` explicitly since the make step could not find the binary on PATH: `0 issues.`. The Docker-based `integration-tests/test.sh` that diffs the updated fixtures was not run here)
- [x] unit and/or integration tests are included (if applicable)
- [ ] documentation is changed or added (if applicable) (no docs change needed, this only adds a warning; behaviour is unchanged)

### Description of change

Stacked on #1113, which has to land first. The diff below is what this branch adds on top of it. No separate issue is filed for this one.

## What's broken

`isSetWarnEmpty` warns when a matcher property is written out as an empty list, since a list of patterns is a list of conditions that must all hold, so zero conditions always passes. It is wired into `file.contains`, `file.contents`, `http.headers`, `http.body`, `command.stdout` and `command.stderr`, but the other list matchers still use a plain `!= nil`, so they pass silently.

```
mount:
  /:
    exists: true
    opts: []
```

```
$ goss validate
..

Total Duration: 0.000s
Count: 2, Failed: 0, Skipped: 0
```

Nothing was asserted about the mount options, and nothing said so. Contrast `opts: ["bogus"]`, which correctly fails.

## The fix

Six one-line conversions to `isSetWarnEmpty`, following the shape already used in `file.go` and `http.go`:

```
mount.opts, mount.vfs-opts, user.groups, interface.addrs, service.runlevels, port.ip, dns.addrs
```

Only fields whose backing getter returns `[]string` are converted. Scalars such as `mount.source`, `user.uid`, `interface.mtu`, `service.running` and `port.listening` are left on `!= nil`, since `isSetWarnEmpty` would fall through to `isSet`'s default branch and only add an unreachable type assertion.

This is a warning, not an error. Behaviour is unchanged: the property is still skipped and the check still passes, exactly as before.

## Why it is stacked

`port.ip` and `dns.addrs` are two of the fields whose generators emit an empty list today, so on master alone this would make goss warn about a file `goss add` had just written. #1113 fixes the generators. With both applied, `goss add` output validates clean.

[...]

PR #1127, opened 2026-09-06
fix(matchers): report a syntax error for matcher groups with no sub-matchers

VXNCXNX:
##### Checklist

  - the Makefile has no `test-all` target, so I left this unticked rather than tick it for something else. `make test-short-all` (fmt, lint, vet, test) passes in full. The `test-int-*` targets need a Docker daemon; no fixture in the repo contains any of the affected shapes, grepped for `and: []`, `contain-elements: []` and `gjson: {}` across every yaml, json and golden file.
  - `docs/gossfile.md` already documents these as `and: [matcher, ..]` and `contain-elements: [matcher, ...]`, so the change makes the behaviour match the documented contract rather than altering it.

### Description of change

Gomega's `And()` with zero matchers is vacuously true, so a matcher group that reduces to no sub-matchers is reported as a pass without the value ever being looked at.

```yaml
user:
  orca:
    groups:
      and: []
```

```
ok 3 - User: orca: groups: matches expectation: {"and":[]}
```

`orca` is in exactly one group, and any assertion about it would have been checked. `and: []` checks nothing and still prints `ok`. The same holds for `contain-elements: []` and `gjson: {}`.

Two siblings in the same function already behave correctly and are deliberately left alone:

| matcher | before | after |
| --- | --- | --- |
| `and: []` | `ok`, asserts nothing | syntax error |
| `contain-elements: []` | `ok`, asserts nothing | syntax error |
| `gjson: {}` | `ok`, asserts nothing | syntax error |
| `consist-of: []` | fails | unchanged, it asserts the value is empty |
| `or: []` | fails | unchanged, it can never be satisfied |

Neither of the last two can produce a false pass, so neither is a syntax error.

This is the same treatment an empty matcher map already gets from `errEmptyMatcher` in #1118.

Worth flagging: a gossfile that contains one of these three today goes from passing to failing. That is the point of the change, but it will surface in existing suites, so it is a behaviour change and not purely a bug fix from the user's side.

Not included, because it overlaps with my open #1114: a bare empty list at the top level of an attribute (`groups: []`) is also vacuously true. #1114 proposes warning on those at the resource layer, and turning the same input into a hard error here would work against it. Happy to fold it in either way once #1114 is resolved.

### Verification

Ran against the built binary on a real gossfile, before and after, for all five shapes in the table. `and: []`, `contain-elements: []` and `gjson: {}` flip from `ok` to a syntax error; `consist-of: []` and `or: []` are byte-identical; `contain-elements: ["orca"]` still passes.

Four test cases added, plus one asserting the two exceptions still build a matcher. Dropping any one of the three guards fails exactly its own subtest and no other:

| guard removed | subtest that fails |
| --- | --- |
| `and` | `EmptyGroup/and` and `EmptyGroup/nested_in_not` |
| `contain-elements` | `EmptyGroup/contain-elements` |
| `gjson` | `EmptyGroup/gjson` |

`make test-short-all` passes, and `golangci-lint run ./resource/...` reports 0 issues.

AI disclosure: written with Claude Code. I ran the binary before and after and ran the mutations myself.

### DIFF
--- a/resource/gomega.go
+++ b/resource/gomega.go
@@ -11,7 +11,15 @@ import (
 var (
 	errMissingRequiredAttribute = errors.New("Syntax Error: Missing required attribute")
 	errEmptyMatcher             = errors.New("Syntax Error: Invalid matcher configuration. An empty map asserts nothing, exactly one matcher is required")
+	errEmptyMatcherGroup        = errors.New("asserts nothing, at least one matcher is required")
 )
+
+// emptyMatcherGroupError reports a matcher that reduces to no sub-matchers at
+// all. Gomega treats that as vacuously true, so the test would be reported as
+// passing without ever looking at the value.
+func emptyMatcherGroupError(name string) error {
+	return fmt.Errorf("Syntax Error: Invalid '%s' argument. An empty value %w", name, errEmptyMatcherGroup)
+}
 
 func matcherToGomegaMatcher(matcher any) (matchers.GossMatcher, error) {
 	// Default matchers
@@ -119,6 +127,9 @@ func matcherToGomegaMatcher(matcher any) (matchers.GossMatcher, error) {
 		if err != nil {
 			return nil, err
 		}
+		if len(subMatchers) == 0 {
+			return nil, emptyMatcherGroupError(matchType)
+		}
 		var interfaceSlice []any
 		for _, d := range subMatchers {
 			interfaceSlice = append(interfaceSlice, d)
@@ -144,6 +155,9 @@ func matcherToGomegaMatcher(matcher any) (matchers.GossMatcher, error) {
 		subMatchers, err := sliceToGomega(value, "and")
 		if err != nil {
 			return nil, err
+		}
+		if len(subMatchers) == 0 {
+			return nil, emptyMatcherGroupError(matchType)
 		}
 		return matchers.And(subMatchers...), nil
 	case "or":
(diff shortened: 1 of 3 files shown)


resource/gomega.go
```diff
@@ -11,8 +11,16 @@ import (
 var (
 	errMissingRequiredAttribute = errors.New("Syntax Error: Missing required attribute")
 	errEmptyMatcher             = errors.New("Syntax Error: Invalid matcher configuration. An empty map asserts nothing, exactly one matcher is required")
+	errEmptyMatcherGroup        = errors.New("asserts nothing, at least one matcher is required")
```
kgaughan, line 14: Can you please update the documentation to make it clear that empty lists/maps will no longer work? Same goes for the changes in other PRs. This is, after all, a change in behaviour. Even if the matches were previously vacuous, raising an error now is a change in behaviour.

VXNCXNX: Done, in both.

Here it is a warning block under Advanced Matchers naming the three that are now syntax errors, the two that look similar and still work, and saying plainly that it is a change in behaviour.

In #1114 the same treatment for the six attributes it touches, following the blocks already there for `stdout`/`stderr` and `contents`.

I checked each statement against a build of the branch rather than writing it off the diff, which caught one I had wrong: I claimed `goss add port` emits `ip: []`, and it does not, because #1114 also changes the generator to leave the attribute out when the list is empty.

kgaughan (approved):

